PR #50, opened 2013-07-31
fix ffmpeg input plugin

stonekyx:
I'm sorry but this is the same pull request as the one I closed just now. I'm new to Github and didn't know that one should create a feature branch for pull requests.

So the following is copied from my last pull request.

---

I don't actually know if the plugin is already working in other branches, but when I used the latest version in the master branch, cmus just made noises when playing .ape or .wma files, which obviously uses ffmpeg plugin.

First I fixed the use of deprecated APIs, by adding more #if branches. When I changed to use avcodec_decode_audio4, the noise partly disappeared, but some audio played too fast this time. It was about sample formats. There were different problems here, but only one solution.

It seemed that the previous version of ffmpeg plugin just gives up when sample_fmt is AV_SAMPLE_FMT_FLT or AV_SAMPLE_FMT_DBL, but I do have some files with these sample formats. The solution I adopted is to use swresample, as this is the simplest one. In ffmpeg_open, the parameters of SwrContext is set, and in ffmpeg_fill_buffer, only a call to swr_convert solves everything. Now my file with AV_SAMPLE_FMT_FLTP plays perfect.

The exact version of ffmpeg I have tested with is 1.2.1. The output plugin was alsa.

My system is ArchLinux. `uname -a` prints:
Linux stone-arch 3.9.3-1-ARCH #1 SMP PREEMPT Mon May 20 08:00:01 CEST 2013 i686 GNU/Linux

### DIFF
--- a/ffmpeg.c
+++ b/ffmpeg.c
@@ -30,9 +30,9 @@
 #include <ffmpeg/avcodec.h>
 #include <ffmpeg/avformat.h>
 #include <ffmpeg/avio.h>
-#include <libswresample/swresample.h>
-#include <libavutil/opt.h>
-#include <libavutil/audioconvert.h>
+#include <ffmpeg/swresample.h>
+#include <ffmpeg/opt.h>
+#include <ffmpeg/audioconvert.h>
 #else
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
@@ -253,14 +253,6 @@ static int ffmpeg_open(struct input_plugin_data *ip_data)
 			break;
 		}
 
-/*#if (LIBAVCODEC_VERSION_INT > ((51<<16)+(64<<8)+0))
-		if (cc->sample_fmt == AV_SAMPLE_FMT_FLT || cc->sample_fmt == AV_SAMPLE_FMT_DBL) {
-#else
-		if (cc->sample_fmt == AV_SAMPLE_FMT_FLT) {
-#endif
-			err = -IP_ERROR_SAMPLE_FORMAT;
-			break;
-		}*/
 		/* We assume below that no more errors follow. */
 	} while (0);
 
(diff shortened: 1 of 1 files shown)


marcelltisdol:
so there is still no clear and concise way to get ALAC/M4A working?

Installed via brew with the --ffmpeg flag

stonekyx:
Not sure how to make brew install cmus with FFmpeg..
But it works pretty well if compiled from source. I have FFmpeg 3.3.2 installed from brew, and compiled cmus with
`./configure prefix=~/fun/opt CONFIG_FFMPEG=y USE_FALLBACK_IP=y && make && make install`
It plays this file without problem: http://www.hyperion-records.co.uk/audiotest/18%20MacCunn%20The%20Lay%20of%20the%20Last%20Minstrel%20-%20Part%202%20Final%20chorus%20O%20Caledonia!%20stern%20and%20wild.M4A

Good to see that FFmpeg is still working here ;) It used to be broken pretty badly back in 2013.

stonekyx:
Just found a solution with brew. You can try this: `brew install cmus --with-ffmpeg --HEAD`

marcelltisdol:
@stonekyx you're awesome my friend, thank you - I tried to install from src last night but was running into problems - I did not have enough time to dedicate to it

one more thing please, I got https://github.com/nogizhopaboroda/cmus-cover-art working and used to have https://github.com/PhilipTrauner/cmus-osx working but not at the same time

The "example" given to setup a wrapper script is poor and I don't know what I'm doing

I'm talking about a wrapper script set for ':set status_display_program' 

If you could help me get this working, I will send in a pull request with better documentation for OSX users

cmus is Foobar for OSX man, if coupled with an external library and if I could get iTunes Sync drivers to work, we're on to something

marcelltisdol:
@stonekyx additional notes

- I also reinstalled ffmpeg with most libraries , via brew
- is there a way to queue up the files that will play next a previous for gapless playback? I'm streaming my entire 1TB music library from my NAS over LAN/VPN - the latency between the devices and also because I'm mounting an ssfs drive (encryption) leave a 2s gap when changing songs. cmus is quick - in iTunes, its around 8 seconds

PR #90, opened 2013-12-31
D-Bus support

mahkoh:
This is a work in progress

---

This patch adds D-Bus support to cmus. D-Bus is an IPC protocol and the de-facto standard on the linux desktop. The reference implementation also runs on OS X and Windows. D-Bus is supported by many languages and allows cmus to be tightly integrated into other applications and the desktop via notifications, panels, etc. 

Here is an example of how a python script can communicate with cmus:

```
#!/usr/bin/env python2
import dbus

bus = dbus.SessionBus()
obj = bus.get_object("net.sourceforge.cmus", "/net/sourceforge/cmus", introspect=False)
cmus = dbus.Interface(obj, "net.sourceforge.cmus")

print "currently playing: ", cmus.artist(), "-", cmus.title()
```

Similarly we can pause the player, go to the next track, or seek:

```
cmus.pause()
cmus.next()
cmus.seek("+10")
```

D-Bus supports two (for us relevant) ways of communication: Methods and Signals. We've seen Methods above. Signals allow cmus to tell other programs that an important event happened, e.g., that the track changed. An example of how this works in for python can be found in `dbus/examples/dbus2.py`. 

I've currently implemented many of the methods from `cmus-remote` and three signals: `track_change`, `vol_change`, and `status_change`.  The full documentation of the D-Bus interface can be found in `dbus/dbus.md`. 

D-Bus support can be disabled at compile time via `CONFIG_DBUS=n`.


mahkoh:
I'm still maintaining this branch in my fork if anyone else is interested in working on is.

PR #230, opened 2015-01-04
Added pwd command

dweymouth:
This small patch adds a `:pwd` command to cmus that allows the user to query the current working directory (which can be set with `:cd`)

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -484,6 +484,9 @@ push <text>
 	This command can only be bound to a key but not used in the command
 	line directly.
 
+pwd
+	Prints the current working directory.
+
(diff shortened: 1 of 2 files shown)


flyingmutant:
Thanks! Merged.

nefthy:
I am bit late since this is already merge, sorry... I don't think that `:pwd` is a good choice for this command. In bash and probably other shells `pwd` prints the name of the working directory. `:qwd` or `:queue-working-directory` seem more apropriate.

flyingmutant:
Well, this command does exactly this: it *prints the working directory*. You've probably misread "query" as "queue".

nefthy:
indeed.

PR #236, opened 2015-01-10
Fix search wrapping

archtimmy:
The wrap_search setting is ignored when the search_next function skips the current match. This prevents search from wrapping if the current match is the last (or first, when searching in reverse) item in view. For instance, searching for "search" in settings will not wrap, since "wrap_search" matches and is the last item in that view.

This is fixed by skipping the current match using the same code as in the search loop, which correctly handles the wrap_search setting.

### DIFF
--- a/search.c
+++ b/search.c
@@ -39,46 +39,58 @@ static void search_unlock(void)
 	editable_unlock();
 }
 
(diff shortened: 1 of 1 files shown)


flyingmutant:
Very nice PR, thank you a lot! Merged.

PR #237, opened 2015-01-11
OpenBSD sndio support

n0la:
This request includes most of the patches of the OpenBSD port. It includes a new output driver called sndio which is the default sound multiplexer in OpenBSD. Could someone test this on other platforms as well, so that these patches do not interfere?

### DIFF
--- a/configure
+++ b/configure
@@ -5,7 +5,7 @@
 check_cflags()
 {
 	check_cc_flag -std=gnu99 -pipe -Wall -Wshadow -Wcast-align -Wpointer-arith \
-		-Wwrite-strings -Wundef -Wmissing-prototypes -Wredundant-decls \
+		-Wwrite-strings -Wundef -Wmissing-prototypes \
 		-Wextra -Wno-sign-compare -Wformat-security
 
 	for i in -Wold-style-definition \
@@ -16,6 +16,16 @@ check_cflags()
 		check_cc_flag $i
 	done
 	return 0
+}
+
+check_sndio()
+{
+	case `uname -s` in
+	OpenBSD)
+		check_library SNDIO "" "-lsndio"
+		return $?
+	esac
+	return 1
 }
 
 check_compat()
@@ -466,6 +476,7 @@ Optional Features: y/n
   CONFIG_AO       	Libao cross-platform audio library              [auto]
   CONFIG_ARTS     	ARTS                                            [auto]
   CONFIG_OSS      	Open Sound System                               [auto]
+  CONFIG_SNDIO    	Sndio                                           [auto]
   CONFIG_SUN      	Sun Audio                                       [auto]
   CONFIG_WAVEOUT  	Windows Wave Out                                [auto]
   USE_FALLBACK_IP	Use a specific IP for every unrecognized	[n]
@@ -531,6 +542,7 @@ check check_samplerate CONFIG_SAMPLERATE
 check check_ao      CONFIG_AO
 check check_arts    CONFIG_ARTS
 check check_oss     CONFIG_OSS
+check check_sndio   CONFIG_SNDIO
 check check_sun     CONFIG_SUN
 check check_waveout CONFIG_WAVEOUT
 check check_roar    CONFIG_ROAR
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -5,7 +5,7 @@
 check_cflags()
 {
 	check_cc_flag -std=gnu99 -pipe -Wall -Wshadow -Wcast-align -Wpointer-arith \
-		-Wwrite-strings -Wundef -Wmissing-prototypes -Wredundant-decls \
+		-Wwrite-strings -Wundef -Wmissing-prototypes \
```
flyingmutant: Hm, why?

sndio.c
```diff
@@ -0,0 +1,261 @@
...
+#include <sndio.h>
+
+#include "op.h"
+#include "mixer.h"
+#include "sf.h"
+#include "xmalloc.h"
+
+static sample_format_t sndio_sf;
+static struct sio_par par;
+static struct sio_hdl *hdl = NULL;
```
flyingmutant: Nitpick: no need for `= NULL`.

sndio.c
```diff
@@ -0,0 +1,261 @@
...
+		break;
+	default:
+		return -1;
+	}
+
+	par.appbufsz = par.rate * 300 / 1000;
+	apar = par;
+
+	if (!sio_setpar(hdl, &par))
+		return -1;
```
flyingmutant: Here and elsewhere: can you please return `-OP_ERROR_SOMETHING` instead of `-1`? It can be a nice idea to log some errors (with `d_print()`) too.

sndio.c
```diff
@@ -0,0 +1,261 @@
...
+static struct sio_hdl *hdl = NULL;
+static int sndio_volume = 100;
+static int sndio_paused;
+
+static int sndio_mixer_set_volume(int l, int r)
+{
+	sndio_volume = l > r ? l : r;
+
+	if (hdl != NULL)
+		sio_setvol(hdl, sndio_volume * SIO_MAXVOL / 100);
```
flyingmutant: Here and elsewhere: can you please check the return values for errors?

sndio.c
```diff
@@ -0,0 +1,261 @@
...
+}
+
+static int sndio_mixer_exit(void)
+{
+	return OP_ERROR_SUCCESS;
+}
+
+static int sndio_mixer_open(int *volume_max)
+{
+	*volume_max = 100;
```
flyingmutant: Why `100` instead of `SIO_MAXVOL`?

flyingmutant:
By the way, `discid` "fix" can stay — if you'll e.g. refactor default device name into function containing `ifdef` for OpenBSD.

flyingmutant:
> Could someone test this on other platforms as well, so that these patches do not interfere?

Travic CI is building on Ubuntu, which supports majority of the plugins. It does not complain about the PR, so I think it is all good in this regard.

n0la:
The discid fix will be removed, and patched during OpenBSD build cycle. I will get the other issues fixed tomorrow and resubmit it.

configure
n0la: This causes a lot (and I mean a lot) of warnings on the OpenBSD version of the toolchain. I will move it back to an OpenBSD port patch.

sndio.c
n0la: The sndio volume is translated to a scale from 0 to 100. I have no idea what the intention behind this, but I assume to "normalize" volume levels across other mixer plugins that also go from zero to 100. I trust the intent of the original author here, and leave this untouched. It would also break current users settings as their values would be lower in volume after upgrading.

n0la:
I updated and cleaned the code according to your suggestions.

sndio.c
```diff
@@ -0,0 +1,268 @@
...
```
flyingmutant: Can you please explain why `100` here instead of `SIO_MAXVOL` used earlier in the file? That way you can avoid `sndio_volume * SIO_MAXVOL / 100` conversion.

n0la: Let me quote myself:

> The sndio volume is translated to a scale from 0 to 100. I have no idea what the intention behind this, but I assume to "normalize" volume levels across other mixer plugins that also go from zero to 100. I trust the intent of the original author here, and leave this untouched. It would also break current users settings as their values would be lower in volume after upgrading.

flyingmutant: Sorry, Github hid the comments because diffs were «outdated» — so I missed this note.

`…_mixer_open(int *volume_max)` has `volume_max` precisely so that mixer plugins do not have to do any scaling internally. I don't think this is a breaking change for users, though. I'll change this myself before merging for code to be a bit cleaner.

n0la: > …_mixer_open(int *volume_max) has volume_max precisely so that mixer plugins do not have to do any scaling internally. I don't think this is a breaking change for users, though. I'll change this myself before merging for code to be a bit cleaner.

I did not know that. That renders that internal conversion pointless then. I have not written that code myself, I just made my task to make it ready for "upstream". I will fix this and update this merge request.

flyingmutant: Thanks, but I've already done that — so no need to update the PR.

flyingmutant:
Merged as 02ca3da6c857fcb5545c7a9e8bb74a4814af74ea (`SIO_SOFTVOL` fix as a separate patch f980054d464f69e733d9a513e9b2677ba0858fa6).

Thanks a lot, once again!

flyingmutant:
@n0la,

This is super-delayed — half a year! — but here is the new cmus release: https://github.com/cmus/cmus/releases/tag/v2.7.0
Hope you can incorporate this into the OpenBSD port.

P.S. I've tried to email you, but that have failed for some reason.

n0la:
Thanks for the new release! I will work on the port ASAP.

Yes my spamd is rather strict :)

PR #239, opened 2015-01-18
Continue to 'cd ..' if parent directory not found.

jolbrich:
I just locked my cmus while adding/categorizing new music. I basically made the following steps:

```
mkdir /tmp/one/two
browse to /tmp/one/two in the Browser view
rm -r /tmp/one
try to browse to '..' in the Browser view: no such file or directory.
```

While this is perfectly correct, I think that I have locked the Browser view in this nonexistent directory.
I further think that it would be reasonable to try to browser_up() to the parent directory until an existing directory is found.
This pull request "works for me", but I'm not sure if I created a memory leak or something.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -334,6 +334,7 @@ void browser_up(void)
 
 	if (browser_load(new)) {
 		if (errno == ENOENT) {
+			free(browser_dir);
 			browser_dir = new;
 			free(pos);
 			browser_up();
--- a/browser.c
+++ b/browser.c
@@ -333,6 +333,12 @@ void browser_up(void)
 	pos = xstrdup(ptr);
 
 	if (browser_load(new)) {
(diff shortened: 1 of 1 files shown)


flyingmutant:
I don't think you've "locked" cmus: `:cd` command should still work, does it?

The patch looks OK, but I think you *did* create a memory leak: you need to `free(browser_dir)` before `browser_dir = new;`. Can you please fix that?

jolbrich:
yes, `:cd` works, didn't know about that command :+1:
free(browser_dir); included.

flyingmutant:
I've added the missing `errno = 0` (didn't spot it at first) and merged this as b8aacb3b2139e59af72792d65a8acd8d066d71ff. Thanks!

PR #240, opened 2015-01-18
Enable -O2

tuncer:
Some compiler flags require at least -O. To fix that, enable -O2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ all: main plugins man
 include config.mk
 include scripts/lib.mk
 
+CFLAGS += -O2
 CFLAGS += -D_FILE_OFFSET_BITS=64
 
 FFMPEG_CFLAGS += $(shell pkg-config --cflags libswresample)


flyingmutant:
*What* compiler flags require that? Your description is a bit mysterious.

And, isn't `-O2` the default [already](https://github.com/cmus/cmus/blob/master/scripts/checks.sh#L170)?

tuncer:
> What compiler flags require that? Your description is a bit mysterious.

`-D_FORTIFY_SOURCE`

In fact, some linux distros have `-D_FORTIFY_SOURCE` and `-fstack-protector-strong` enabled by default, 

Also see https://bugzilla.gnome.org/show_bug.cgi?id=677373.

> And, isn't -O2 the default already?

Seems like it, and I agree that it's not necessarily the right place to set `-O2`, but does this work as expected when a distro sets CFLAGS etc. in the env before running `./configure`? 

Here's the warning:
```
/usr/include/features.h:328:4: warning: #warning _FORTIFY_SOURCE
requires compiling with optimization (-O) [-Wcpp]
 #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
```

flyingmutant:
>> And, isn't -O2 the default already?

> Seems like it, and I agree that it's not necessarily the right place to set -O2, but does this work as expected when a distro sets CFLAGS etc. in the env before running `./configure`?

Well, I'd say it works exactly *as expected* — if distro provides its own CFLAGS, cmus uses those. It seems to me that the proper solution is to add `-O2` to distro-supplied CFLAGS, no? Otherwise the distro has no idea if `-D_FORTIFY_SOURCE` it wants has any effect at all.

tuncer:
Yeah, although some projects extend CFLAGS, it's perfectly fine for cmus not to. Thanks for the clarification. Closing.

flyingmutant:
Just to be perfectly clear: cmus does extend CFLAGS [here](https://github.com/cmus/cmus/blob/master/configure#L7-L18) – with `std=gnu99` and a bunch of warnings only, though.

tuncer:
Sure, those are project-specific flags you require in either case and can be considered exceptions.

PR #242, opened 2015-01-19
Fix -Wunused-result warnings

tuncer:
```
ui_curses.c: In function 'ui_curses_notify':
ui_curses.c:2094:2: warning: ignoring return value of 'write',
declared with attribute warn_unused_result [-Wunused-result]
  write(notify_in, &c, 1);
  ^
ui_curses.c: In function 'notify_init':
ui_curses.c:2332:2: warning: ignoring return value of 'pipe',
declared with attribute warn_unused_result [-Wunused-result]
  pipe(fildes);
  ^
ui_curses.c: In function 'main_loop':
ui_curses.c:2210:4: warning: ignoring return value of 'read',
declared with attribute warn_unused_result [-Wunused-result]
    read(notify_out, buf, sizeof(buf));
    ^
```

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -2091,7 +2091,9 @@ static void u_getch(void)
 void ui_curses_notify(void)
 {
 	char c;
-	write(notify_in, &c, 1);
+	if (write(notify_in, &c, 1) == -1) {
+		d_print("write failed: %s\n", strerror(errno));
+	}
 }
 
 static void main_loop(void)
@@ -2207,7 +2209,9 @@ static void main_loop(void)
 
 		if (FD_ISSET(notify_out, &set)) {
 			char buf[128];
-			read(notify_out, buf, sizeof(buf));
+			if (read(notify_out, buf, sizeof(buf)) == -1) {
+				d_print("read failed: %s\n", strerror(errno));
+			}
 		}
 	}
 }
@@ -2329,7 +2333,7 @@ static void init_curses(void)
 static void notify_init(void)
 {
 	int fildes[2];
-	pipe(fildes);
+	BUG_ON(pipe(fildes) == -1);
 	notify_out = fildes[0];
 	notify_in = fildes[1];
 	int flags = fcntl(notify_in, F_GETFL, 0);


ui_curses.c
```diff
@@ -2329,7 +2333,9 @@ static void init_curses(void)
 static void notify_init(void)
 {
 	int fildes[2];
-	pipe(fildes);
+	if (pipe(fildes) == -1) {
+		d_print("pipe failed: %s\n", strerror(errno));
```
nefthy: fildes is not valid since pipe failed, so there should be a return here.

nefthy: if `notify_init()` fails, calls to `ui_curses_notify()` will also fail. This means that there will be an inconsistent state. I think its better to exit cmus with an error message at this point.

tuncer: Sure, what's the right logging function to print an error message?

Do you think the other d_print() calls are fine, or should they also lead to cmus exiting?

tuncer: Should I call exit_all()?

flyingmutant:
To exit cmus, return an error code from `notify_init`, and just `return` if it is non-zero in `main()`. There is no need to exit cmus in other situations.

P.S. cmus does not actually use `-Wunused-result`.

tuncer:
Done, please review.

flyingmutant:
Why did you drop `d_print`? `error_msg` is wrong here.

tuncer:
Changed the non-terminating errors back to `d_print`, and left `notify_init` as is. 

Shouldn't we print an error message before terminating?

Also, fixup commits will be squashed once it's all finished.

ui_curses.c
```diff
@@ -2326,14 +2330,18 @@ static void init_curses(void)
 	update_mouse();
 }
 
-static void notify_init(void)
+static int notify_init(void)
 {
 	int fildes[2];
-	pipe(fildes);
+	if (pipe(fildes) == -1) {
+		error_msg("pipe failed: %s\n", strerror(errno));
```
flyingmutant: `error_msg` here makes no sense — no one will see it, because cmus will exit immediately. Please use `d_print` here, too.

flyingmutant:
These errors will never happen in practice — `d_print` is more than enough for it. Even `BUG` would be OK here, frankly.

tuncer:
Okay, please review.

flyingmutant:
Looks good now!

tuncer:
Thinking about it, it would probably be nice to know why pipe() failed. So a perror() could still be useful.

flyingmutant:
We have spent *way* too much time discussing changes of more or less purely theoretical importance. So let's finally merge this and be done with it :-)

PR #244, opened 2015-01-22
fix pulseaudio alsa-sink 24s bug

### LINKED ISSUE #224: Track time off by 24 seconds ALSA
As mentioned here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=695069
When I start a track, it displays that it's at 00:24 instead of 00:00. No effect on audio. Shows next track is playing during the last 24 seconds of current track.

I tried both with ffmpeg and libmad as plugin in use, so unlikely input plugin related. It's mentioned this is related to ALSA and **not** PulseAudio. Strange that it doesn't seem to affect Banshee/Rhythmbox etc.

Compiled on Ubuntu 14.04

Idk if below is useful:
```
> cat /proc/asound/version
Advanced Linux Sound Architecture Driver Version k3.13.0-40-generic.
```

```
alsa.h from `libasound2-dev (1.0.27.2-3ubuntu7)
```

It may also be a config related issue. I checked the official cmus package from ubuntu repositories and getting same results. Let me know if any more info would be useful.

mahkoh:
If you have `pulseaudio` and `pulseaudio-alsa` installed and when you start a new song, cmus will show that the song is already at the 24 seconds mark even though the song plays normally. When you seek, cmus will skip forward 24 seconds (+- what you're seeking.) 

If you start pulseaudio with verbose output, you can see the following message: 
```
D: [pulseaudio] alsa-util.c: Maximum hw buffer size is ___23777 ms___
D: [pulseaudio] alsa-util.c: Set buffer size first (to 88200 samples), period size second (to 88200 samples).
```

So it seems that when we're using the pulseaudio alsa sink, cmus will send alsa 24 seconds of audio before alsa tells it that the buffer is full. When you seek, cmus tells alsa to discard the buffer, which explains why we skip forward 24 seconds.

This patch tells alsa to buffer at most 100ms.

Closes #224 

PS: This bug was first reported in 2010: http://cmus-devel.narkive.com/gpx2XhqP/play-time-off-by-24s

### DIFF
--- a/alsa.c
+++ b/alsa.c
@@ -125,6 +125,7 @@ static int op_alsa_exit(void)
 static int alsa_set_hw_params(void)
 {
 	snd_pcm_hw_params_t *hwparams = NULL;
+	unsigned int buffer_time_max = 300 * 1000; /* us */
 	const char *cmd;
 	unsigned int rate;
 	int rc, dir;
@@ -133,6 +134,12 @@ static int alsa_set_hw_params(void)
 
 	cmd = "snd_pcm_hw_params_any";
 	rc = snd_pcm_hw_params_any(alsa_handle, hwparams);
+	if (rc < 0)
+		goto error;
+
+	cmd = "snd_pcm_hw_params_set_buffer_time_max";
+	rc = snd_pcm_hw_params_set_buffer_time_max(alsa_handle, hwparams,
+	                                           &buffer_time_max, &dir);
 	if (rc < 0)
 		goto error;
 


flyingmutant:
Thanks a lot — fantastic work!

But it looks to me that you set `buffer_time_max` to `1000000` (us), which is 1s, not 100ms. Maybe 200-300ms would be a bit better (less wakeups, underruns less likely)? Also, do you know what was ALSA's own default?

mahkoh:
You're right, the number is a typo. I tested it first with 50ms (I believe, it might have been 500ms if I had the same typo there) and that worked fine. I don't know what alsa uses because `...get_buffer_time_{max/min}` return `UINT_MAX` and `1` respectively.

I'm not using PA myself (and I've already uninstalled it again after debugging this.) Maybe someone else can report if 100ms works? (@eat-multi)

At the end of the track cmus will jump to the next song after it's sent alsa the last buffer but before alsa has finished playing it. To make sure that this is not noticeable, the buffer has to be quite small.

flyingmutant:
> At the end of the track cmus will jump to the next song after it's sent alsa the last buffer but before alsa has finished playing it

AFAICS this is purely UI thing, ALSA will still play the last buffer before playing anything new — correct?

mahkoh:
Yes, the same happens right now but with 24s instead of <1s.

flyingmutant:
Regarding 100ms — I am sure that will work, the question is, how likely it makes sound to drop in case machine becomes overloaded or unresponsive. I'd make the buffer «as big as possible, but not bigger».

mahkoh:
I've updated the PR with a 300ms buffer.

flyingmutant:
Thanks — I hope that'll be good enough. Merged.

PR #257, opened 2015-02-26
 Add CMUS_SOCKET environment variable.

mist:
This has basically the same effect as #68 .

Sometimes I am using several cmus instances at the same time and this is no longer possible since XDG_RUNTIME_DIR/cmus-socket is being used, so now the socket is located in CMUS_SOCKET if this env is set.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -19,7 +19,8 @@ can be controlled from the outside via *cmus-remote*(1).
 @h1 OPTIONS
 
 --listen ADDR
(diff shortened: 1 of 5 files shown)


flyingmutant:
Out of curiosity — what do you use multiple cmus instances for?

mist:
Basically I am having different config setups for different NAS availability. Sometimes I want to compare libraries with each other to see if I want to copy anything eg. from the NAS to my laptop and the most convenient way is by running two different cmus instances next to each other.

flyingmutant:
Thanks, merged!

PR #260, opened 2015-02-27
Split auto_expand_albums into three options. Fixes #258

### LINKED ISSUE #258: Don't expand artists when following
I'd like the feature of cmus following the currently played track but because I'm playing random all my artists are opened. Instead I would love cmus to show the currently played track in the artist overview except when the albums of the artist are already showing.

mist:
You can now differ between what to do when searching, following and selecting the currently played track, each of them has their own config options:
auto_expand_albums_search, auto_expand_albums_follow, and auto_expand_albums_selcur.

This fixes #258


tree.c
```diff
@@ -1249,7 +1249,12 @@ void tree_sort_artists(void)
 
 void tree_sel_current(void)
 {
-	tree_sel_track(lib_cur_track);
+	tree_sel_track(lib_cur_track, auto_expand_albums_selcur);
+}
+
+void tree_sel_current_follow(void)
```
flyingmutant: Hm, why is this still a separate function, with others like it gone?

mist:
In the first version there were three `tree_sel_track_*` functions which now are gone. The `tree_sel_current` function was called when changing tracks and when pressing i, now it ist only called when pressing i.

flyingmutant:
Now there is also `void tree_sel_current_follow(void)` — I was talking about it.

flyingmutant:
Any updates here?

mist:
I will remove it and change the function to `void tree_sel_current( int auto_expand_albums )`. I've been busy, will do it later this week I think

mist:
Sorry for not hearing from me for so long.

Now it should be as you suggested.

flyingmutant:
Thanks! Merged.

PR #264, opened 2015-03-10
Fix mp4 garbage output

mahkoh:
Turns cue garbage disposal into two more general functions


flyingmutant:
Thanks, looks good! Merged.

PR #283, opened 2015-05-20
New feature: Play queue with a circular mode when enabled.

fossette:
On branch circular-play-queue
Changes to be committed:
	modified:   Doc/cmus.txt
	modified:   cmus.c
	modified:   options.c
	modified:   options.h
	modified:   ui_curses.c

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -786,6 +786,9 @@ altformat_trackwin [`Format String`]
 buffer_seconds (10) [1-300]
 	Size of player buffer in seconds.
 
+circular (false)
+	Play queue in circular mode when true.
+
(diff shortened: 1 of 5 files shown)


fossette:
This new 'circular' configuration option, along with the 'resume' configuration option lets cmus continuously play the same tracks in the play queue, even after several power off - power on cycles.

:add -q <path>   to load the play queue
:set resume=true
:set circular=true

That's it!

Dominique.

flyingmutant:
Sorry — the "circular" queue looks exactly like a playlist (with repeat turned on). I don't think there is a point in creating duplicate  functionality here.

Regarding restoring the playlist position on resume — I agree that it is a welcome feature, and I'll gladly accept a separate PR for that.

Issue #297, opened 2015-06-29
Code to display SRT - Someone might evaluate?


brunomacabeusbr:
I am writing this code to display the music lyrics. But I am not sure whether it is good and organized.

One problem I'm having is in hundredths of seconds of the song position. I do not know where it is, then the synchronization is not good because it compared to the second.

Another problem is that I do not know where should I put the code to read SRT files, since they do not know very well the project organization.

**Diff:** http://pastebin.com/teF24j0m

flyingmutant:
Sub-second player position is calculated inside `player.c` as `(double)consumer_pos / (double)buffer_second_size()`, however, it is not exposed anywhere currently. 

Do you intend to propose this code for inclusion into cmus, or is it just for your own use?

I'd suggest developing on GitHub, since online pull request / diff functionality is very handy. Also, please use cmus' coding standard — your code should look exactly like the rest of the code (naming, indentation etc.). Code to read SRT files can probably go inside `srt.h` / `srt.c`.

brunomacabeusbr:
The lyrics is still not perfect. You can still see some delays in the exchange. The delay is probably caused during the update of the interface.

I created the files `srt.h / srt.c` and `read_srt_for_song` function, but I get the following error message during compilation when call it in the function `track_info_new`: `undefined reference to read_srt_for_song`
Even though I have made the header and include it in `track_info.c`
I do not know for what reason the `srt` not compiled.

It's getting here the code so far: http://pastebin.com/YHqcyg3A
Please, did you could create `srt.h / srt.c` files? I really could not make it work using them.


I want to send to the repository that improvement.

flyingmutant:
I am sorry to say that, but the patch you've posted (btw, [here](https://gist.github.com/flyingmutant/828f4afe015fb71ae529) is a prettier version) is of a pretty low quality right now, and it sounds like you are not very familiar with C and Makefiles. I would be happy to help — but I simply have not enough time to properly guide you through C, cmus [build system](https://github.com/cmus/cmus/blob/master/Makefile), source code organization and [coding style](http://www.kernel.org/doc/Documentation/CodingStyle).

Once again, sorry for not helping here (and probably sounding a bit condescending)— I am afraid I have not enough time on my hands to do that.

I am closing this issue. If you'll be able to produce a high quality patch (obviously correct, written in clean and idiomatic C, organized and looking like the rest of cmus' code) — please open a separate pull request for that.

PR #311, opened 2015-07-19
Load and view SRT

brunomacabeusbr:
I made several modifications to make it more readable compared to what <a href="https://github.com/cmus/cmus/issues/297">was here</a>


flyingmutant:
While it is slightly better than the previous version, most of the comments from #297 still stand — this code is not of quality that I consider mergeable.

PR #323, opened 2015-08-23
Call status display program on exit

DanielRS:
I added a new status to player, "exiting". After exiting all sub-systems the status_display_program is called.


DanielRS:
I'm going to clean the branch and submit a cleaner pull request.

PR #324, opened 2015-08-23
Allow multiple arguments in status_display_program

DanielRS:
Additional arguments can now be specified, e.g.

`set status_display_program=cmus-status --changed --capitalize`


DanielRS:
This changes certainly introduced some undefined behaviour. As `spawn` would sometimes fail when this patch is used. Ignoring this.

PR #325, opened 2015-08-23
Calls status_display_program on exit

DanielRS:
status_display_program is called with status "exiting" during cmus exit process.


mahkoh:
This has been open for over a year. I think the idea is good but I'm not sure about the implementation. Adding the additional player state is a bit of a hack. I would prefer to add this to the dbus implementation but mpris doesn't have a signal for "player exit".

mahkoh:
Closing. You can listen for the "NameAcquired" and "NameLost" signals sent by the dbus daemon instead. Those will be sent whenever cmus starts or exits with the "org.mpris.MediaPlayer2.cmus" argument.

mahkoh:
Listening for "NameOwnerChanged" is probably more appropriate and combines both cases in one.

jolange:
@mahkoh I have looked into the dbus interface and managed to receive property changes (in python) like this:
```
bus.add_signal_receiver(receiver_fun, bus_name = "org.mpris.MediaPlayer2.cmus", dbus_interface = "org.freedesktop.DBus.Properties")
```
Trying something similar for detecting cmus exiting did not work:
```
bus.add_signal_receiver(receiver_fun, bus_name = "org.mpris.MediaPlayer2.cmus", dbus_interface = "org.freedesktop.DBus.NameOwnerChanged")
```
Is this what you had in mind?

mahkoh:
```
~$ dbus-monitor | grep -A3 NameOwnerChanged
signal time=1488729505.775120 sender=org.freedesktop.DBus -> destination=(null destination) serial=9019 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string "org.mpris.MediaPlayer2.cmus"
   string ""
   string ":1.4145"
--
signal time=1488729506.664161 sender=org.freedesktop.DBus -> destination=(null destination) serial=9020 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string "org.mpris.MediaPlayer2.cmus"
   string ":1.4145"
   string ""
```

jolange:
Great, thanks :-)

Issue #319, opened 2015-08-19
TAK audio playback


d125q:
[TAK is a lossless audio codec](http://wiki.hydrogenaud.io/index.php?title=TAK) and it is supported by FFmpeg. Is there a way to force cmus to load such files? Would it be possible to add support for them in the future?

flyingmutant:
Hi @d125q, sorry for delayed reply — I was on vacation. 

 Can you try to write a PR similar to [this commit](https://github.com/cmus/cmus/commit/2d47150af78e918c77d0129d91d8663b3c9884ee)?

d125q:
Hi. Sorry I have also been away. I will write a PR as soon as I can. Thanks a lot!

flyingmutant:
Fixed in #330.

PR #330, opened 2015-09-03
Add support for *.tak files as per cmus/cmus#319

d125q:
I am not sure if we should check if FFmpeg's version supports TAK (as some relatively old versions do not). Let me know if that's the case.


flyingmutant:
I think the patch is fine as-is. Thanks a lot for contributing! Merged.

PR #333, opened 2015-09-16
Add support for *.aa files

inAudible-NG:
FFmpeg >= 2.8 (libavformat >=56.40.101) is required
for playing *.aa files.

A sample *.aa file can be downloaded from,
http://samples.ffmpeg.org/audible/


flyingmutant:
Thanks a lot! Merged.

PR #335, opened 2015-09-19
Fix HTTP-URI parsing bugs

akretz:
http_parse_uri doesn't work correctly when the URI's path contains an "at" or a colon. For example, the valid URI http://somedomain.com/path:withcolon fails to parse. Also, the proxy-value is undefined when http_parse_uri fails (e.g.: adding the stream http://somedomain.com:/ segfaults on my machine when free is called).


flyingmutant:
Thanks a lot! Merged.

PR #339, opened 2015-10-05
Fix missing ffmpeg metadata

### LINKED ISSUE #313: ogg/opus tags not recognized
I have some ogg/opus files whose tags are not available inside cmus, i.e., they are placed in the `<No Name>` section. The tags are shown when I play the file with mpv.

ogg/opus files are read with ffmpeg so I deleted all other input plugins and the cache and restarted cmus. But once again all but those ogg/opus files did have their tags read properly. This includes some ogg/vorbis files.

~~I'm providing one of the affected files for the purpose of fixing this problem: http://www.mediafire.com/listen/9pl59t1dzzcdj98/Oh_Wonder_-_Landslide.ogg~~

Here is an example: http://www.mediafire.com/listen/uy98p2123ushh45/example.ogg

mahkoh:
Certain formats (such as ogg/opus) contain metadata in more than one
part of the file. ffmpeg does not automatically merge this metadata so
we have to manually collect it from all streams.

Closes #313

### DIFF
--- a/ffmpeg.c
+++ b/ffmpeg.c
@@ -505,6 +505,16 @@ static int set_comment(struct keyval *comment, int i, const char *key, const cha
 }
 #endif
 
+static void ffmpeg_read_metadata(struct growing_keyvals *c, AVDictionary *metadata)
+{
+	AVDictionaryEntry *tag = NULL;
+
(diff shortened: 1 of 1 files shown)


flyingmutant:
Looks good! Merged, thanks a lot.

PR #352, opened 2015-10-31
Fix for short terminal

srang:
Previously, if you resize your terminal to less than 8 lines, there was weird behavior when moving around or when the track changed such as the current track line disappearing or partially disappearing and then reappearing. This changes the min height to be 2 lines so that as long as there is space to show the current track line it does but if there is only one line display current selection.


flyingmutant:
Thanks! Merged.

PR #353, opened 2015-11-02
adds gruvbox theme

elliottetzkorn:
Added a gruvbox theme for myself that might be worth including in the cmus master since it's a popular vim theme. Screenshot shows my cmus next to vim. I tried to make style choices so that it would look similar to vim.
screen shot 2015-11-01 at 16 58 53


flyingmutant:
Thanks a lot! Merged.

PR #360, opened 2015-11-15
Link against libavcodec

sebastinas:
The ffmpeg plugin uses functions from libavcodec, so ensure that it is linked
against libavcodec.

### DIFF
--- a/configure
+++ b/configure
@@ -378,7 +378,7 @@ check_aac()
 check_ffmpeg()
 {
 	HAVE_FFMPEG_AVCODEC_H=y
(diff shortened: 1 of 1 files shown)


flyingmutant:
Probably better to place this near `libavformat` [here](https://github.com/cmus/cmus/blob/master/configure#L381).

sebastinas:
Like that?

flyingmutant:
Yep :-)

flyingmutant:
Thanks for the PR!

PR #373, opened 2015-12-18
Enable user to set device interface when using libao plugin.

Wang-Yue:
Tested on CoreAudio.


flyingmutant:
@Wang-Yue, can you please add the documentation for the new option?

Wang-Yue:
PR is updated.

flyingmutant:
Thanks a lot — merged!

Issue #331, opened 2015-09-11
Mac OS X 10.11 El Capitan: Using deprecated Carbon Component Manager


herrbischoff:
While compiling and running `cmus` via `homebrew` on Mac OS X 10.11 GM Candidate with Xcode 7, the following notice is displayed whenever first starting playback of audio:

    This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.

I believe this should be addressed to keep compatibility with future OS releases.
Let me know if you need more details and what about.

flyingmutant:
You are using libao (`ao` output plugin), right? This notice is probably about it, and there is not much we can do aside from hoping libao will update in the near future.

Can you please check if `jack` output plugin is OK on 10.11?

herrbischoff:
Correct, the standard for Homebrew appears to be `ao`.

I tried to use the `jack` plugin, I remembered that from using `muc` 
back then. Unfortunately I was not able to activate it. `jack` is 
installed, I rebuilt `cmus` from the master branch afterwards with `brew 
reinstall cmus --HEAD` but when trying to set the output plugin via 
`:set output_plugin=jack`, all I get is the following error:

    Error: selecting output plugin 'jack': no such plugin

Any idea what is wrong?

nefthy:
The jack plugin is probably not installed. Do you have the jack header files installed?

herrbischoff:
@nefthy: I double-checked just a minute ago. They are indeed installed in `/usr/local/Cellar/jack/0.124.1/include/jack`, which appear to be identical to those found [here](https://github.com/jackaudio/headers).

Interestingly, compiling by hand (not via homebrew) finds those during the `./configure` run:
```
checking for JACK_LIBS (pkg-config)... -L/usr/local/Cellar/jack/0.124.1/lib -ljack -lpthread -framework CoreAudio -framework CoreServices -framework AudioUnit
checking for JACK_CFLAGS (pkg-config)... -I/usr/local/Cellar/jack/0.124.1/include
```

This does indeed result in the creation of the `jack.so` plugin. Manually copying it to `/usr/local/Cellar/cmus/HEAD/lib/cmus/op` makes it available and it runs as expected.

I should probably go though the homebrew build process to fix it and submit a pull request adding jack support.

[...]

PR #374, opened 2015-12-21
native CoreAudio plugin

Wang-Yue:
native CoreAudio plugin, with interesting features such as hog mode, auhal, device selection, volume setting, native audio format, and more!

Comparing to libao, this plugin reduces core audio daemon CPU utilization at least to 5% (sometimes to 1%). In addition you're not going to see deprecate warnings populated onto your terminal.


flyingmutant:
@Wang-Yue, why did you close this PR? Sorry for not reviewing it yet — I simply have very little time right now, but I think I'll be able to do this on weekend.

configure
```diff
@@ -28,6 +28,17 @@ check_sndio()
 	return 1
 }
 
+check_coreaudio()
+{
+	case `uname -s` in
+	Darwin)
```
nefthy: Is that the only posible system name for MacOS?

coreaudio.c
```diff
@@ -0,0 +1,857 @@
...
+typedef struct coreaudio_ring_buffer_t {
+	size_t buffer_size;
+	size_t write_index;
+	size_t read_index;
+	size_t big_mask;
+	size_t small_mask;
+	char *buffer;
+} coreaudio_ring_buffer_t;
+
+int coreaudio_ring_buffer_init(coreaudio_ring_buffer_t *rbuf, size_t num_of_bytes);
```
nefthy: the coreaudio_ring_buffer_* functions should be static, as they are not used outside this file.

Also, there seems to be a ring buffer implementation called CARingBuffer in CoreAudio. Is there any reason not to use it?

coreaudio.c
```diff
@@ -0,0 +1,857 @@
...
+		return -OP_ERROR_ERRNO;
+	}
+
+	if (coreaudio_device_name)
+	{
+		property_address.mSelector = kAudioHardwarePropertyDevices;
+		err = AudioObjectGetPropertyDataSize(
+			kAudioObjectSystemObject,
+			&property_address, 0, NULL, &property_size);
+		if (err == noErr)
```
nefthy: what if err != noErr?

coreaudio.c
```diff
@@ -0,0 +1,857 @@
...
+	{
+		AudioDeviceDestroyIOProcID(coreaudio_device_id, coreaudio_proc_id);
+	}
+	coreaudio_proc_id = NULL;
+	coreaudio_device_id = kAudioDeviceUnknown;
+	coreaudio_ring_buffer_destroy(&coreaudio_ring_buffer);
+	return OP_ERROR_SUCCESS;
+}
+
+static int coreaudio_open(sample_format_t sf, const channel_position_t *channel_map)
```
nefthy: this function is quite long, consider splitting it in smaller parts.

Doc/cmus.txt
```diff
@@ -1230,6 +1230,12 @@ dsp.ao.wav_dir
 	Output directory for libao plugin, default is your home directory.
 	Used only if *dsp.ao.driver* is "wav".
 
+dsp.coreaudio.device
+	Device for Core Audio output. Leave empty for default.
+
+dsp.coreaudio.enable_hog_mode
+	Set true to set the device to hog mode. The default value is false.
```
nefthy: what is hog mode?

nefthy:
Other than the comments above, it seems well written to me. Can't compile/test though, as I don't have a Mac.

coreaudio.c
Wang-Yue: CARingBuffer is demo code. It's not part of CoreAudio.

Doc/cmus.txt
Wang-Yue: Hog mode is a CoreAudio feature that you gain direct access to the hardware. No other process can read from/write to the audio hardware. This feature is useful when you want to disable all other sound (such as system notification sound) send to the speaker. 

Audiophile players such as Vox or Foobar2000 have similar features.

Wang-Yue:
@flyingmutant I am still actively enhancing its feature set.  I will send a new PR when it's fully ready. 

If you really want to merge something now, a8c0cad is the golden version, and we can always add more features later.

configure
Wang-Yue: OpenDarwin/PureDarwin also use this name. But I doubt if there's anyone using these OSes.

coreaudio.c
Wang-Yue: Doesn't matter. We then fallback to use the default system output. See line 309. We can do a d_print, though.

Wang-Yue:
Please review commit  b3901e9. It's has pretty good CoreAudio feature set support (Comparable to the VOX player).

flyingmutant:
Thanks a lot — I'll try to test and review it as soon as I can.

jwohlert:
I just built the newest version of your PR and was not able to get cmus to recognise the coreaudio plugin.

When i execute `set output-plugin=coreaudio`, I get the following message: `selecting output plugin 'coreaudio': no such plugin`.

However, listing the plugins, coreaudio.so is present.

```bash
% ls /usr/local/lib/cmus/op
ao.so        coreaudio.so
```

flyingmutant:
That's weird — the plugin does work for me.

Wang-Yue:
The plugin should work.

could you check its permission? Maybe the executable doesn't have permission to visit the plugin?
alternatively you can enable DEBUG=2 and try to see what's the error message output by 
https://github.com/cmus/cmus/blob/2bfc4624ddc8961b8d1b892fb5236b874cfdfb0d/output.c#L82

wohlert:
Running with `DEBUG=2` yielded the following:

```bash
cat ~/cmus-debug.txt
main: charset = 'UTF-8'
op_load_plugins: /usr/local/lib/cmus/op/coreaudio.so: dlopen(/usr/local/lib/cmus/op/coreaudio.so, 2): Symbol not found: _AudioComponentFindNext
  Referenced from: /usr/local/lib/cmus/op/coreaudio.so
  Expected in: flat namespace
 in /usr/local/lib/cmus/op/coreaudio.so
```

flyingmutant:
Merged, thanks a lot — CoreAudio plugin is a big and welcome addition!

@Wang-Yue — please take a look at how the [coreaudio.c](https://github.com/cmus/cmus/blob/master/coreaudio.c) looks now. I've tried to limit the scope of use of globals, and massage the code a bit. Hope I did not break anything with my changes :-)

flyingmutant:
@wohlert — can you please create a separate issue for the problem you are experiencing?

Wang-Yue:
@wohlert what version of OS X you are using? Can you try to link it with AudioUnit in addition to CoreAudio? That should solve the problem.  
You can do this by adding -framework AudioUnit after -framework CoreAudio at 
https://github.com/cmus/cmus/blob/8b8b83da5dd5b480c118a005e82940d74c72e053/configure#L35

Wang-Yue:
@flyingmutant Thanks! This does lot a lot readable.

Sorry I was not very familiar with the Linux kernel coding style... and I don't know how to edit/debug automake project in Xcode (So it actually took > 12 hours, which is much more than I originally thought about...) So my code is messy. thanks for the cleaning up!

Wang-Yue:
Oh we also need to send a PR to homebrew when a new version is released,as we are not depending on libao anymore:)

Doc/cmus.txt
nefthy: would be nice if you add that to the documentation. I don't think every end user understands the term hog mode.

wohlert:
@Wang-Yue yes, it works now. Appending `-framework AudioUnit` did the trick. I am running 10.11.1 so it might be an El Capitan related problem?

Anyway, I am submitting a PR with your solution as a bug fix.

Wang-Yue:
@wohlert I have no idea as I am using El Capitan too. Yep Please do a PR. Thanks.

jaredbeck:
>  [with] this plugin ... you're not going to see deprecate warnings populated onto your terminal. 
> .. we also need to send a PR to homebrew ..

Did this happen? I'm looking forward to not seeing the ".. [deprecated Carbon Component Manager][1] .." anymore. Thanks for writing this plugin!

[1]: https://github.com/cmus/cmus/issues/331

Wang-Yue:
@jaredbeck 
You shouldn't see issue 331 when using this plugin. You would see it if you're using jack or libao for output.

jaredbeck:
Sorry, I don't know how to use plugins. All I know how to do is `brew upgrade cmus`. Is there a new `brew` release yet?

Wang-Yue:
@jaredbeck no. you have to build it and manually enable it.

Wang-Yue:
@jaredbeck  I think brew can build from head. you need to pass in the right parameter.

flyingmutant:
Yeah, `brew install --HEAD cmus` should work.

jaredbeck:
Thanks, I'll just wait for an actual brew release, unless you want my help testing before then?

herrbischoff:
This is absolutely awesome! Thank you very much @Wang-Yue.

Wang-Yue:
@herrbischoff glad you like it.

squfrans:
useful info in here, thank you.  until `brew` catches up with the version including `coreaudio`, here are the manual install steps:
```sh
$ brew uninstall cmus
$ brew install --HEAD cmus
$ brew cleanup
```

inside `cmus`:
`:set output_plugin=coreaudio`

papertigers:
@Wang-Yue is there a way to get the device id for the particular coreaudio device I want to use.  Right now I have just changed it in the sound preferences and that seems to have worked but that means all of my mac's audio is going to this device now.

PR #377, opened 2015-12-25
Added check for AudioUnit when configuring

wohlert:
This is a small commit to to check for the presence of AudioUnit framework when configuring the build. This PR is in relation to #374.

### DIFF
--- a/configure
+++ b/configure
@@ -32,7 +32,7 @@ check_coreaudio()
 {
 	case `uname -s` in
 	Darwin)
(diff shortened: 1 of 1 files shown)


Wang-Yue:
This is indeed needed. 

I neglected this since I have ffmpeg plugin installed. and ffmpeg is already linked to AudioUnit.

flyingmutant:
Thanks!

PR #383, opened 2016-01-09
Replace deprecated FFmpeg API

sebastinas:
We have received the included patch in Debian. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810557 for details.

### DIFF
--- a/ffmpeg.c
+++ b/ffmpeg.c
@@ -39,7 +39,11 @@
 #include <libavformat/avio.h>
 #include <libswresample/swresample.h>
 #include <libavutil/opt.h>
+#if LIBAVUTIL_VERSION_MAJOR >= 53
+#include <libavutil/channel_layout.h>
+#else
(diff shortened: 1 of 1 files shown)


flyingmutant:
I wonder, maybe we should just drop the ton of ifdefs in ffmpeg.c, and only support the versions shipped in latest Debian/Ubuntu/Redhat releases? Maintaining the ifdef magic seems almost impossible right now (e.g. this patch changes some of the ifdefed code without introducing new ifdefs — so it *can* break build with *some* version of ffmpeg, right?).

sebastinas:
The patch now included updates to the #ifdefs according to Andreas this should work with all releases from the last two years. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810557#17

flyingmutant:
Thanks a lot for your work over at Debian! Hope ffmpeg will be problem-free for a while.

PR #384, opened 2016-01-10
Libcue2

Wang-Yue:
libcue2 has a changed header, and unknown length is -1 rather than 0.

### DIFF
--- a/cue.c
+++ b/cue.c
@@ -150,7 +150,11 @@ static int cue_open(struct input_plugin_data *ip_data)
 	if (rc)
 		goto ip_open_failed;
 
+#ifdef CONFIG_CUE2
 	if (track_get_length(t) != -1)
+#else
+	if (track_get_length(t) != 0)
+#endif
 		priv->end_offset = priv->start_offset + _to_seconds(track_get_length(t));
 	else
 		priv->end_offset = ip_duration(priv->child);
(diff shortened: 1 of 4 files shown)


cue.c
```diff
@@ -254,7 +257,6 @@ static int cue_read_comments(struct input_plugin_data *ip_data, struct keyval **
 	Track *t;
 	Rem *track_rem;
 	Cdtext *track_cdtext;
-	char *val;
```
Smattr: @flyingmutant can probably say for certain, but my guess is that the previous forward declarations of these variables was to maintain C90 compatibility. I have no idea if this is a priority or not, but you can achieve this with less churn in this diff by just marking this as `const` here.

Apologies for butting in on this PR if C90 compatibility is not relevant.

Wang-Yue: I've revert back to c90. Thanks. 

The reason we need to have another variable (path) in addition to child_filename is because free() takes a non-const pointer.

flyingmutant:
Merged as 68d650c8fada236582a6c2fbc80d8dff065c630d (I've excluded all non-libcue2 related changes), thanks a lot!

Wang-Yue:
Thanks @flyingmutant 

libcue2 changed a few return pointers to const pointers. So the compiler will give a few warnings if these pointers are not const. It can still compile, though.

PR #389, opened 2016-01-24
read BPM from metadata

jayay:
Hello :octocat:!

These changes allow cmus to read the BPM value from your file's metadata.

Reading BPM works for both FLAC using libflac and MP3 using LibMAD. It doesn't work yet when the input plugin is FFMPEG, as they don't seem to have an interface for that.

Using `bpm` in filters or sorting is also possible, for filtering one might use `bpm > 0` to get every track which has a BPM value in its metadata, if it hasn't one it's set to `-1`.

The value is also being displayed in the status bar, as you can see on this screenshot:
screenshot from 2016-01-24 14-14-29

jayay


jarv:
The indentation is off when I look at the diff, maybe you are using spaces instead of tabs?

jayay:
Thanks! Also changed my editor settings for the future.

expr.c
```diff
@@ -427,6 +427,7 @@ static const struct {
 	{ 's',	"stream"	},
 	{ 't',	"title"		},
 	{ 'y',	"date"		},
+	{ 'b',  "bpm"           },
 	{ '\0',	NULL		},
```
flyingmutant: I'd prefer to have only the most frequently used tags in this map.

jayay:
Okay.

flyingmutant:
Also, can you please mention new formatting option in the docs?

jayay:
Yes! Could not find any more relevant references.

flyingmutant:
Merged as 976c10d0e42c9ecd7389b28dd7c5b560a1702821, thanks!

jayay:
:clap: cool!

PR #390, opened 2016-01-25
add move view command

frkwy:
It added a command to the window movement of the left and right quickly.
The key bindings for example
rc

```
unbind common ^L
bind common ^L right-view
unbind common ^H
bind common ^H left-view
```
I think that it is easier to use.


flyingmutant:
Can you add the documentation for the new commands?

PR #394, opened 2016-02-03
add natural sort

mahkoh:
As an optional feature.


mahkoh:
@jayay Since this was originally your code, I'm fine with you taking this PR over if that's something you care about.

jayay:
Haha, thanks. But it's fine as it is.

flyingmutant:
Looking at the PR, what I do not like is that it introduces new optional feature that defaults to false. That means that: a) most users will not benefit from it b) we introduce rarely used code which should be maintained nevertheless.

So, I think the best is not to merge any natural sort implementation, unless it is a straight improvement over regular sort (and thus does not need any new options).

I'll close this PR, but if anybody will find time to make natural sort implementation mergeable per definition above —I'll be very glad to review the attempt.

PR #402, opened 2016-02-20
Multiple fixes

yne:
This PR fix multiple kinds of:
  - NULL pointer freeing
  - missing variable initialization
  - buffer under-run
  - memory leak (missing free)
  - duplicate/useless assignment

I'm not particularly proud of af413d7 but I can't found a better way to know if he the `old` track was the last referenced, and if so, avoid calling editable_update_track then track_info_unref again on it :/

Any advice would be appreciated

### DIFF
--- a/http.c
+++ b/http.c
@@ -121,7 +121,9 @@ void http_free_uri(struct http_uri *u)
 {
 	free(u->uri);
 	free(u->user);
-	free(u->pass);
+	if (u->pass) {
+		free(u->pass);
+	}
 	free(u->host);
 	free(u->path);
 
--- a/http.c
+++ b/http.c
@@ -513,6 +513,6 @@ char *base64_encode(const char *str)
 	case 0:
 		break;
 	}
-	buf[d++] = 0;
+	buf[d] = 0;
 	return buf;
 }
--- a/job.c
+++ b/job.c
@@ -331,7 +331,7 @@ void do_update_job(void *data)
 
 	for (i = 0; i < d->used; i++) {
 		struct track_info *ti = d->ti[i];
-		struct stat s;
+		struct stat s = {};
 		int rc;
 
 		/* stat follows symlinks, lstat does not */
--- a/job.c
+++ b/job.c
@@ -57,7 +57,7 @@ static void flush_ti_buffer(void)
 	editable_lock();
 	for (i = 0; i < ti_buffer_fill; i++) {
 		jd->add(ti_buffer[i]);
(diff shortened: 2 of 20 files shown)


http.c
```diff
@@ -121,7 +121,9 @@ void http_free_uri(struct http_uri *u)
 {
 	free(u->uri);
 	free(u->user);
-	free(u->pass);
+	if (u->pass) {
+		free(u->pass);
+	}
```
flyingmutant: `free(NULL)` is a noop, no need to change anything.

main.c
```diff
@@ -388,7 +391,7 @@ int main(int argc, char *argv[])
 		send_cmd("player-play\n");
 	if (flags[FLAG_PAUSE])
 		send_cmd("player-pause\n");
-	if (flags[FLAG_FILE])
+	if (flags[FLAG_FILE] && play_file)
```
flyingmutant: `play_file` is assigned at line 318 (`case FLAG_FILE:`), I believe this change is redundant.

job.c
```diff
@@ -331,7 +331,7 @@ void do_update_job(void *data)
 
 	for (i = 0; i < d->used; i++) {
 		struct track_info *ti = d->ti[i];
-		struct stat s;
+		struct stat s = {};
 		int rc;
```
flyingmutant: Since `rc` is checked below, this is purely cosmetic change, right?

flyingmutant:
Out of curiosity, how did you find all those – just reading the code, or using some kind of static analysis tool?

flyingmutant:
As for af413d70f6d704e18a0306319d6ee6a05591b743 — yeah, it does not seem right. `editable_update_track` only derefs the reference that was «owned» by the editable, so I believe there is no need to change anything at all. Either the track info was referenced by an editable, then it is correctly dereferenced, or it was not referenced — then nothing changes.

main.c
```diff
@@ -349,8 +350,10 @@ int main(int argc, char *argv[])
 	if (server == NULL)
 		server = xstrdup(cmus_socket_path);
 
-	if (!remote_connect(server))
+	if (!remote_connect(server)){
+		free(server);
```
flyingmutant: `server` can also come from argv, in that case this is actually wrong. Since not freeing stuff before returning from `main()` is fine, no need to change anything here.

flyingmutant:
I've merged what I believe fixed actual errors. Thanks a lot!

Issue #403, opened 2016-02-25
cmus-remote --pause toggles playback, not pause it.


kingy9000:
I am trying to use `cmus-remote -u` to only pause playback when I lock my screen, but it can only toggle playback status. The problem with using `cmus-remote -s` is that it starts the current song from the beginning when I play it again, which is annoying for a long mix of songs.

I have made a work around for this by checking `cmus-remote -Q` to check if cmus is even playing before running `cmus-remote -u` but this sounds like a feature that should be implemented into cmus itself since that would be more efficient and neater.

I understand that `cmus-remote -u` needs to toggle playback instead of only pausing to be able to use cmus with media keys, but this could be worked around by adding an extra option that only pauses cmus.

Thanks for reading.

jessebolson:
+1

jolange:
I also think there should be separate commands for `pause` and `pause/play toggle`. I could try to implement this, but I'm not sure if one should break the current `pause` behaviour and add a further `toggle` option, or keep the current `pause` behaviour and add a `pause-only` option.

Since this would be a change in cmus itself and not only for `cmus-remote`, I think breaking the current behaviour would be annoying to most users. @flyingmutant what do you think?

flyingmutant:
As for compatibility with status quo – we should definitely preserve current behavior. PR adding new commands would be very welcome.

PR #415, opened 2016-03-16
adding a player-pause-no-resume command

jolange:
which only pauses if playing, but does not resume when stopped/paused

implements #403 

I'll add a dedicated cmus-remote command and the relevant documentation once you think the implementation is ok and worth integrating.

### DIFF
--- a/Doc/cmus-remote.txt
+++ b/Doc/cmus-remote.txt
@@ -41,6 +41,9 @@ When *-C* is given all command line arguments are treated as raw commands.
 -u, --pause
 	Toggle pause.
 
+-U, --pause-playback
+	Pause when currently playing.
+
 -s, --stop
 	Stop playing.
 
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -463,6 +463,9 @@ player-next (*b*)
 player-pause (*c*)
 	Toggle pause.
 
+player-pause-playback
+	Pause when currently playing.
+
 player-play [filename] (*x*)
 	Play the given track, or, if none is specified, [re]play the current
 	track from the beginning.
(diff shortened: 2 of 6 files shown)


flyingmutant:
Sorry for delayed reply! The PR looks like a good start. It'd be good if someone can come up with a better name — but I can't suggest anything decent enough myself.

jolange:
No problem, I'm in no hurry.
Right, the natural name for this function is certainly simply "pause", but I agree that this should not be changed. Another alternative would be "pause-only", but I like the more descritpive version better... Good suggestions are of course welcome!

Anyway, I'll then continue with what I said and see if we can improve the name.

jolange:
Ok, I think I'm ready now. Please tell me should I have overlooked something.
In any case we should maybe wait to see if someone has a better idea for a name.
I'll rebase the changes when things are settled if you like.

jayay:
How about `--rest`?

nefthy:
I think breaking backward compatibility is the right thing to do here. Every cd/dvd/mp3-player out there implements pause as "stop here and continiou later from this point". I realy doubt that there are mission critical scripts out there, that would break with this change, out there.

jolange:
I don't think we should introduce a new terminology for this, this is still "pausing", just under different conditions.

flyingmutant:
@jolange thanks a lot for the PR. Just to let you know — I think I'll only be able to take a closer look at it not sooner than Sunday, unfortunately.

jolange:
No problem, take your time.

flyingmutant:
How about `--pause-playback`?

jolange:
Yes, that sounds a bit better.

flyingmutant:
Great! Can you please update the PR?

jolange:
Sure, I'll do that later today. If you are satisfied apart from that, I can also squash and rebase the changes.

jolange:
@flyingmutant ok, done.

PR #416, opened 2016-03-17
issue template

jayay:
Hey!

Often, when a bug is submitted, we don't immediately get all the information needed. Sometimes something was simply forgot to add, such as the compiler, plugins or cmus version. But to prevent that, I thought it might be a good idea to create an issue template.

You can try it by creating a [new issue](https://github.com/jayay/cmus/issues/new) in my fork. I'm pretty sure it's not complete yet, so please tell me what's missing or add it yourself.

I hope this is useful and saves some time in future.

Cheers!

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,19 @@
+<!-- Please fill the following if you found an issue, delete if otherwise. -->
+
+Expected behaviour: 
+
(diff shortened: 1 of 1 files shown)


flyingmutant:
Thanks a lot for the idea! I've added a slightly less bureaucratic version in  634503a

jayay:
Haha! Great! :v:

PR #428, opened 2016-04-02
fix compiler warnings

nefthy:


flyingmutant:
What are platform/compiler/warnings?

nefthy:
gcc 4.9.3  / linux

```
cache.c: In function 'read_cache':
cache.c:212:9: warning: missing initializer for field 'st_dev' of 'struct stat' [-Wmissing-field-initializers]
  struct stat st = {};
         ^
In file included from /usr/include/sys/stat.h:104:0,
                 from utils.h:31,
                 from cache.c:24:
/usr/include/bits/stat.h:48:13: note: 'st_dev' declared here
     __dev_t st_dev;  /* Device.  */
             ^
```

```
track.c: In function 'shuffle_list_get_next':
track.c:91:9: warning: declaration of 'filter' shadows a global declaration [-Wshadow]
   int (*filter)(const struct simple_track *))
         ^
In file included from search_mode.h:28:0,
                 from track.c:21:
/usr/include/ncursesw/curses.h:620:29: warning: shadowed declaration is here [-Wshadow]
 extern NCURSES_EXPORT(void) filter (void);    /* implemented */
                             ^
track.c: In function 'shuffle_list_get_prev':
track.c:117:9: warning: declaration of 'filter' shadows a global declaration [-Wshadow]
   int (*filter)(const struct simple_track *))
         ^
In file included from search_mode.h:28:0,
                 from track.c:21:
/usr/include/ncursesw/curses.h:620:29: warning: shadowed declaration is here [-Wshadow]
 extern NCURSES_EXPORT(void) filter (void);    /* implemented */
                             ^
track.c: In function 'simple_list_get_next':
track.c:143:9: warning: declaration of 'filter' shadows a global declaration [-Wshadow]
   int (*filter)(const struct simple_track *))
         ^
In file included from search_mode.h:28:0,
                 from track.c:21:
/usr/include/ncursesw/curses.h:620:29: warning: shadowed declaration is here [-Wshadow]
 extern NCURSES_EXPORT(void) filter (void);    /* implemented */
                             ^
track.c: In function 'simple_list_get_prev':
track.c:166:9: warning: declaration of 'filter' shadows a global declaration [-Wshadow]
   int (*filter)(const struct simple_track *))
         ^
In file included from search_mode.h:28:0,
                 from track.c:21:
/usr/include/ncursesw/curses.h:620:29: warning: shadowed declaration is here [-Wshadow]
 extern NCURSES_EXPORT(void) filter (void);    /* implemented */
                             ^
```

nefthy:
clang does not warn about the shadowing.

nefthy:
gcc 5.3 does not warn about the initialization.

flyingmutant:
Thanks! Merged.

cache.c
```diff
@@ -209,7 +209,7 @@ void cache_remove_ti(struct track_info *ti)
 static int read_cache(void)
 {
 	unsigned int size, offset = 0;
-	struct stat st = {};
+	struct stat st;
```
mahkoh: This should instead be changed to `{ 0 }`.

flyingmutant: Good catch! Thanks.

mahkoh: Not that it matters since it works either way, but the initialization was added some months ago because the lack of an initializer also causes a warning. `{ }` is the same as `{ 0 }` but is a GNU extension.

flyingmutant: Heh, no:
```
cache.c:212:21: warning: missing field 'st_mode' initializer [-Wmissing-field-initializers]
        struct stat st = {0};
                           ^
1 warning generated.
```

mahkoh: Then that is a bug in the compiler you are using. Section 6.7.8 paragraph 19 says that

>all subobjects that are not initialized explicitly shall be initialized implicitly the same as
>objects that have static storage duration.

mahkoh: That being said, we might have explicitly enabled a warning that requires every field to be explicitly initialized. However, since newer versions of gcc don't produce this warning, the gcc developers seem to have decided that this was indeed a mistake.

flyingmutant: This is the current OS X clang version (`Apple LLVM version 7.3.0 (clang-703.0.29)`). Let's just stick with `{}`, modern GCC should be OK with this (https://github.com/cmus/cmus/pull/428#issuecomment-204972047).

flyingmutant: Done in https://github.com/cmus/cmus/commit/7b5297a5b9b0690901b2c129056b877ed40dc160.

mahkoh: That's odd because we're explicitly passing `-Wno-missing-field-initializers` to the compiler. So either way you should never see that warning.

mahkoh: `= { 0 }` was explicitly exempted from the `missing-field-initializers` warning in gcc in 2011: https://gcc.gnu.org/viewcvs/gcc?view=revision&revision=172857

nefthy:
Its fine either way, the struct is initialized by the subsequent stats call. `= { 0 }` only initializes the first member IIRC. `= {}` works with clang and gcc 5.3, so its fine with me.

mahkoh:
`= { 0 }` initializes all members. And the initializer was added because `stat` doesn't necessarily initialize the structure. At least this was the justification given at that time.

nefthy:
the example in the stat manpage on my system looks like this:
```C
           struct stat sb;
[...]
           if (stat(argv[1], &sb) == -1) {
               perror("stat");
               exit(EXIT_FAILURE);
           }
```

nefthy:
Ah, now I see what was the original problem. The return value of `fstat` is not checked, instead `st.st_size < sizeof(cache_header)`, which could contain junk if the call failed, dependen on system/compiler/alignmentofstars.

PR #433, opened 2016-04-13
some s/when/if/

avoidr:


flyingmutant:
Thanks for the PR!

Are you a native speaker? Some of the "when" uses you've changed look OK to me — but English is foreign language for me.

avoidr:
I am not native, but I'm pretty sure that it's now correct. I didn't change anything I wasn't sure of. Some examples are more clear than others.

What's PR here? Either way, you're welcome, I guess :P

>Some of the "when" uses you've changed look OK to me

Do you mean that the new version looks now correct, or do you think that I may have changed something which was correct before?

flyingmutant:
PR means Pull Request :-)

avoidr:
Ah, right. I kept thinking "Public Relations", lol. At least the first two letters in each word were correct :-D

PR #442, opened 2016-05-09
make IP priority an option

mahkoh:


flyingmutant:
Thanks! Looks good, I'll try to test and merge this tomorrow.

flyingmutant:
There is an issue: changing the priority should invalidate the cache, or at least warn the user that the cache is no longer consistent with current configuration.

mahkoh:
Is a comment in the documentation good enough?

flyingmutant:
The best would be to require confirmation when setting the option from inside cmus as well — but only if it is trivial to add.

mahkoh:
What do you want the warning to say?

flyingmutant:
Something along the lines that «metadata cache will be inconsistent with new configuration until you run `:update-cache -f` manually. continue?», probably.

mahkoh:
I've made it so that it warns the first time you change a priority.

flyingmutant:
Thanks for the info — I'll temporarily revert the PR.

PR #446, opened 2016-05-10
Split ffmpeg plugin into legacy and new

mahkoh:
The ffmpeg plugin is a mess.

Since the ffmpeg API changes relatively often, the plugin is littered with ifdefs that make it hard to reason about the code. In the next ffmpeg release, more changes will be included that amplify this problem: [[1]](https://ffmpeg.org/pipermail/ffmpeg-cvslog/2016-April/099152.html), [[2]](https://ffmpeg.org/pipermail/ffmpeg-devel/2016-March/190525.html).

This patch aims to fix this problem as follows:

1) The ffmpeg plugin is split into two plugins, ffmpeg_legacy and ffmpeg_new. Each installation contains exactly one of these plugins which is then simply called ffmpeg. When choosing which plugin to compile, ffmpeg_new is preferred.

2) ffmpeg_new drops support for all ffmpeg versions prior to 1.0 which was released in 2012. 1.0 is one of the earliest versions which have API documentation exists and for which all subsequent API changes are documented. I assume that almost all users will be able to use ffmpeg_new.

3) ffmpeg_new removes all ifdefs from the program logic. Fortunately most ifdefs are only needed because functions were renamed. The decoding logic is rewritten to use the logic of the new ffmpeg API linked above. When an older ffmpeg version is used, the new API is simulated with helper functions.

I've tested the new plugin with the latest ffmpeg version on arch and the git master. Since I had to make some changes to the logic, it would be good if some people could test the new plugin.


ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
+/*
+ * Copyright 2008-2013 Various Authors
+ * Copyright 2007 Kevin Ko <kevin.s.ko@gmail.com>
```
flyingmutant: Copyright info probably needs to be updated.

ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
...
+
+static inline void ffmpeg_buffer_flush(struct ffmpeg_output *output)
+{
+	output->buffer_pos = output->buffer;
+	output->buffer_used_len = 0;
+}
+
+static void ffmpeg_init(void)
+{
+	static int inited = 0;
```
flyingmutant: Better to use a true global variable.

ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
...
+	av_register_all();
+}
+
+static int ffmpeg_open(struct input_plugin_data *ip_data)
+{
+	struct ffmpeg_private *priv;
+	int err = 0;
+	int stream_index = -1;
+	int64_t channel_layout = 0;
+	AVCodec *codec;
```
flyingmutant: `= NULL` as well, for consistency.

I'd say it is almost always better to explicitly initialize variables. This way, reader of the code does not need to double-check «is this always properly initialized later?» all the time.

ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
...
+	case AV_SAMPLE_FMT_U8:
+		ip_data->sf |= sf_bits(8) | sf_signed(0);
+		av_opt_set_sample_fmt(swr, "out_sample_fmt", AV_SAMPLE_FMT_U8,  0);
+		break;
+	case AV_SAMPLE_FMT_S32:
+		ip_data->sf |= sf_bits(32) | sf_signed(1);
+		av_opt_set_sample_fmt(swr, "out_sample_fmt", AV_SAMPLE_FMT_S32,  0);
+		break;
+	/* AV_SAMPLE_FMT_S16 */
+	default:
```
flyingmutant: Better to have an explicit `case AV_SAMPLE_FMT_S16` and `default: BUG()`.

ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
...
+	return 0;
+}
+
+static int ffmpeg_fill_buffer(AVFormatContext *ic, AVCodecContext *cc,
+		struct ffmpeg_input *input, struct ffmpeg_output *output, SwrContext *swr)
+{
+	AVFrame *frame = ffmpeg_frame_alloc();
+	int res;
+
+again:
```
flyingmutant: Any reason for `goto` instead of a loop here?

ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
...
+	}
+}
+
+static void ffmpeg_read_metadata(struct growing_keyvals *c, AVDictionary *metadata)
+{
+	AVDictionaryEntry *tag = NULL;
+
+	while ((tag = av_dict_get(metadata, "", tag, AV_DICT_IGNORE_SUFFIX))) {
+		if (tag->value[0])
+			comments_add_const(c, tag->key, tag->value);
```
flyingmutant: Hm, do ffmpeg metadata labels map 1-to-1 to cmus' comment ones?

ffmpeg_new.c
```diff
@@ -0,0 +1,555 @@
...
+	.seek = ffmpeg_seek,
+	.read_comments = ffmpeg_read_comments,
+	.duration = ffmpeg_duration,
+	.bitrate = ffmpeg_bitrate,
+	.bitrate_current = ffmpeg_current_bitrate,
+	.codec = ffmpeg_codec,
+	.codec_profile = ffmpeg_codec_profile
+};
+
+const int ip_priority = 300;
```
flyingmutant: Whoa! I'd suggest we keep the old priority, until we will be *very* confident that the plugin does indeed improve the status quo.

Also, changing the priority should probably cause a cache version bump, to force metadata re-reading.

flyingmutant:
Can I ask why you believe that a) using ffmpeg will be better cmus users; b) the new plugin will not repeat the fate of the «legacy» one?

I'll try to do a more thorough review (and some basic testing) tomorrow.

P.S. In May 2012, @weisslj wrote a script to check the compilation of the plugin against different versions of ffmpeg. If you are serious about maintaining this plugin, something similar may come handy in the future.

ffmpeg_new.c
mahkoh: Why?

ffmpeg_new.c
mahkoh: I don't know enough about audio to say if that's correct or not. But I think the intention here is to use AV_SAMPLE_FMT_S16 as the fallback.

ffmpeg_new.c
mahkoh: I don't think ffmpeg has such a thing. It merely passes on what it finds in the files. I haven't seen any bad behavior in my library.

ffmpeg_new.c
mahkoh: If the priority is too low, people can't test it. I'll reset the priority once it's ready to be merged.

mahkoh:
>a) using ffmpeg will be better cmus users

See my comment above.

>b) the new plugin will not repeat the fate of the «legacy» one?

I can't make such a promise but it's my intention to keep it clean.

>In May 2012, @weisslj wrote a script to check the compilation of the plugin against different versions of ffmpeg. If you are serious about maintaining this plugin, something similar may come handy in the future.

The only reliable solution would be to add this script to travis.

ffmpeg_new.c
flyingmutant: Certainly not «bad behavior», but https://github.com/cmus/cmus/blob/master/comment.c#L216 may not catch everything.

ffmpeg_new.c
flyingmutant: The `default` clause does not check for the sample format, only assumes it (in the form of comment). Better to explicitly check for `AV_SAMPLE_FMT_S16`, and have a default which should be unused (for now), with something BUG-like in it.

This way, we will be protected from `cc->sample_fmt` containing something new in the future; and the logic will be expressed in code, not comments.

ffmpeg_new.c
mahkoh: Keep in mind that this commit does not change either the priority of the plugin nor its comment handling code.

ffmpeg_new.c
mahkoh: I think you're missing the point of this code. We're checking the sample format of the audio the decoder gives us, `cc->sample_fmt`, and set an appropriate target format for resampling that can be handled by the rest of cmus. In fact, I assume that the code would be just fine if we simply replaced this whole `switch` by the `default` case.

ffmpeg_new.c
flyingmutant: Static local is the same global mutable state, only «hidden» inside a function. This «hiding» does not have any real benefits — but makes it harder for reader to know about global mutable state used by the plugin. Placing it as a global on top of file makes it easier to read the code: more explicit and more in line with other cmus (or C, in general) code.

flyingmutant:
> The only reliable solution would be to add this script to travis.

Good idea — only need to make sure it fits in Travis build time limits.

ffmpeg_new.c
mahkoh: >more explicit and more in line with other cmus (or C, in general) code.

I'm not sure if this is the case. 1)l I think that limiting the scope of a variable is almost always desirable. ￼2) There quite a few scoped static variables currently in the master branch:

```
ack "\tstatic" *.c *.h | wc -l
27
```

ffmpeg_new.c
flyingmutant: Sorry – I was just skimming the code and not really reading it, as I should. Good point, thanks.

ffmpeg_new.c
flyingmutant: >  I think that limiting the scope of a variable is almost always desirable

Then we disagree in what a limiting is — IMO static local is just a weird syntax for a global, so the scope is global anyway.

P.S. Let's stop arguing about this one, it is not that important.

ffmpeg_new.c
flyingmutant: Yeah — I was just reading the code more from the «one plugin to rule them all» than «updated ffmpeg plugin» point of view.

flyingmutant:
```
checking for FFMPEG_LIBS (pkg-config)... -L/usr/local/Cellar/ffmpeg/3.0.2/lib -lavformat -lavcodec
checking for FFMPEG_CFLAGS (pkg-config)... -I/usr/local/Cellar/ffmpeg/3.0.2/include
checking for successful build of ffmpeg_new.c... no
checking for header <libavcodec/avcodec.h>... yes
checking for successful build of ffmpeg_legacy.c... yes
```

```
gp@MacBook-Pro-Gregory ~/d/cmus (mahkoh-ffmpeg_legacy) [1]> clang -std=gnu99 ffmpeg_new.c -I/usr/local/Cellar/ffmpeg/3.0.2/include -L/usr/local/Cellar/ffmpeg/3.0.2/lib -lavformat -lavcodec
ffmpeg_new.c:128:43: error: use of undeclared identifier 'got_frame'; did you mean 'got_fram'?
        len = avcodec_decode_audio4(cc, frame, &got_frame, &avpkt);
                                                ^~~~~~~~~
                                                got_fram
ffmpeg_new.c:119:6: note: 'got_fram' declared here
        int got_fram = 0;
            ^
ffmpeg_new.c:139:10: error: use of undeclared identifier 'got_frame'; did you mean 'got_fram'?
        return !got_frame;
                ^~~~~~~~~
                got_fram
ffmpeg_new.c:119:6: note: 'got_fram' declared here
        int got_fram = 0;
            ^
2 errors generated.
```

flyingmutant:
Also, it might make sense to add `ffmpeg.c` to `.gitignore`.

mahkoh:
Can you check why ffmpeg_new doesn't compile? (Just create the soft-link manually.)

mahkoh:
Oh, sorry. I see that you already did that. I thought the error was from ffmpeg_legacy.

mahkoh:
Fixed.

flyingmutant:
I think we need to have something like `ffmpeg (new)` / `ffmeg (legacy)` in `cmus --plugins` output; otherwise it will be hard to know which version of plugin is used.

Also, can you please use the new `sf_hostendian()` stuff here as well?

flyingmutant:
Can you please address https://github.com/cmus/cmus/pull/446#issuecomment-218897111 and rebase the PR on top of master?

mahkoh:
I'm no longer happy with this. I think it's better to just drop support for old ffmpeg versions.

flyingmutant:
I am not against this — we just need to find a reasonable starting version, which will cover the majority of distributions in the wild.

Closing this one?

mahkoh:
The previous Debian version (released in 2013) does not come with 1.0 but has it in its backport repos. I'm not sure if such an old Debian version is used on systems which would not want to enable backports.

The previous Ubuntu LTS 14.04 does have 1.0 in its repos I believe. This Ubuntu version is still used in some places and should probably be supported.

Closing for now.

mahkoh:
In any case I don't think such a large change should be made before the next release.

flyingmutant:
Yeah, sure. Stabilization is the most important thing right now; all major features will be reviewed/merged only after the release.

Issue #448, opened 2016-05-19
Tree view not sorting by album year?


joaquinlpereyra:
---

1) `uname -a`:
```
Linux bombadil 4.5.4-1-ARCH #1 SMP PREEMPT Wed May 11 22:21:28 CEST 2016 x86_64 GNU/Linux

```

2) `cmus --version`:
```
cmus v2.7.1
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg

Output Plugins: /usr/lib/cmus/op
  pulse
  alsa
  oss

```

According to the documentation, the Tree View should sort the albums by year. This is not happening. There's a screenshot attached where you can see the problem. I've tried checking if the actual tags displayed as 'Year' on cmus are the ones on the files, and indeed they were. 

Screenshot: http://i.imgur.com/dx55E98.png

flyingmutant:
Here is the screenshot, for easier navigation:

![dx55e98](https://cloud.githubusercontent.com/assets/73059/15416604/e457a928-1e55-11e6-9b47-3a29dfa59346.png)

flyingmutant:
Thanks for reporting! Looks like a bug indeed.

Can you please post what `mutagen-inspect` says about first track of each album?

joaquinlpereyra:
Sure. Here you go. This information is about the albums showed on the screenshot, although the problem happens also with other artists. 

```
~/.music ❯❯❯ mutagen-inspect Variations\ on\ a\ Theme/01\ On\ the\ Mountain\ at\ Dawn.mp3
-- Variations on a Theme/01 On the Mountain at Dawn.mp3 
[...]

PR #450, opened 2016-05-21
Now year sorting corresponds to year displayed on tree view

joaquinlpereyra:
In reference to issue 448: https://github.com/cmus/cmus/issues/448


flyingmutant:
This is a wrong fix: `%{maxdate}` is «album's» date. While displaying it «fixes» the sorting, you dismiss the date metadata from the individual tracks, which is not good.

PR #456, opened 2016-06-04
remove device stream sync code

Wang-Yue:
Only perform sample rate synchronization when user opt in sync_sample_rate. Other values in the format should be transparent for CoreAudio conversion. Particularly, If user have high end DACs (such as 24 bit DAC) it's always better to select higher bit depth rather than sync it with the audio file. This gives user more freedom adjusting software volume without losing precision.


flyingmutant:
Less code is always good! Thanks.

Issue #453, opened 2016-05-27
Separation of plugin-specific code


vit1-irk:
Certain parts of plugins' code are stored in the player itself (especially cue and cd), so I have started to write a layer to provide these functions apart from main code.

Code (this commit separates just cue support for start) compiles successfully and doesn't make trouble, but I'm not yet sure to open pull request.

https://github.com/cmus/cmus/compare/master...vit1-irk:special_handlers?expand=1

Please watch the code and leave a comment (_да и на русском тоже можно_) if I'm doing right and which details are required to know about plugin system and these functions (to handle file names, playlist entries and special links).

Thanks in advance!

flyingmutant:
Hi @vit1-irk,

First of all, sorry for super-delayed response — it is really hard for me to find any free time as of late :-(

The thing you do certainly makes sense, and the code looks OK in general (I have some comments, but better to leave them for the PR).

I'll close this issue — please feel free to open a PR once you will be ready. Can't promise timely review on that, unfortunately.

P.S. Let's keep the discussion in English, because this way it would be accessible to the vast majority of users/contributors.

PR #460, opened 2016-06-13
Special handlers

vit1-irk:
That's continuation of closed issue #453 

I've finally managed to separate playlist addition functions for cue and cd into their own plugins.
Then I've done support of init actions for any plugin (deadbeef has something like that too) to perform registering of those handlers.

This code compiles successfully (no warnings), and GDB shows that special_handlers initialization passed OK, but I think that additional testing is required anyway.

Waiting for response.


special_handlers.c
```diff
@@ -0,0 +1,55 @@
+#include "utils.h"
+#include "special_handlers.h"
+#include "xmalloc.h"
+
```
mahkoh: Please `typedef` every single function pointer type.

vit1-irk: Yeah, I should really have done this before :)
Thanks.

utils.h
```diff
@@ -176,7 +175,12 @@ static inline int is_cue_url(const char *name)
 
 static inline int is_url(const char *name)
 {
-	return is_http_url(name) || is_cdda_url(name) || is_cue_url(name);
+	return is_http_url(name) || is_cdda_url(name);
+}
+
+static inline int is_partial_url(const char *name)
```
flyingmutant: What does this function do, exactly? And why some urls (cue) are "partial_url", but not "url"?

special_handlers.h
```diff
@@ -0,0 +1,24 @@
+typedef void (*adder)(const char*, int);
```
flyingmutant: Include guard is missing.

special_handlers.c
```diff
@@ -0,0 +1,55 @@
...
+		if (mimetype_handlers.detectors[i](filename))
+			return mimetype_handlers.strings[i];
+	}
+	return NULL;
+}
+
+void register_special_filename_handler(identify_func detector, special_adder add_file_special) {
+	int count = filename_handlers.count;
+	filename_handlers.count++;
+	filename_handlers.detectors = xrealloc(filename_handlers.detectors, sizeof(int(*)(const char* filename)) * (count + 1));
```
flyingmutant: Can `typedef` be used here as well?

special_handlers.c
```diff
@@ -0,0 +1,55 @@
+#include "utils.h"
+#include "special_handlers.h"
+#include "xmalloc.h"
+
+special_adder get_special_filename_handler (const char* filename) {
+	int i;
+	for (i=0; i<filename_handlers.count; i++) {
```
flyingmutant: `for (int i = …` is legal C.

cue_utils.c
```diff
@@ -23,16 +23,22 @@
 
 #include <stdio.h>
 
+int is_cue_filename(const char *name)
+{
+	const char *ext = get_extension(name);
+
+	if (ext != NULL && strcmp(ext, "cue") == 0)
+		return 1;
+	else return 0;
```
flyingmutant: `return ext != NULL && strcmp(ext, "cue") == 0)`

cue_utils.c
```diff
@@ -23,16 +23,22 @@
...
 {
 	FILE *fp;
-	const char *ext;
 	char buf[4096] = {0};
 	const char *dot;
 
-	ext = get_extension(filename);
-	if (ext != NULL && strcmp(ext, "cue") == 0)
-		return NULL;
+	if (is_cue_filename(filename)) return xstrdup(filename);
```
flyingmutant: Please split statements like this into two lines.

cue_utils.h
```diff
@@ -33,9 +33,8 @@
 #endif
 
 
+int is_cue_filename(const char *name);
 char *associated_cue(const char *filename);
 int cue_get_ntracks(const char *filename);
 char *construct_cue_url(const char *cue_filename, int track_n);
-
-
```
flyingmutant: No need to remote the whitespace.

cdio.c
```diff
@@ -563,3 +584,8 @@ const char * const ip_options[] = {
 const int ip_priority = 50;
 const char * const ip_extensions[] = { NULL };
 const char * const ip_mime_types[] = { "x-content/audio-cdda", NULL };
+
+extern void ip_init_function(void) {
+	register_special_filename_handler(is_cdda_url, add_cdda);
+	register_special_mimetype_handler(is_cdda_url, "x-content/audio-cdda");
+}
```
flyingmutant: Missing newline at the end of file.

input.c
```diff
@@ -496,6 +500,8 @@ void ip_load_plugins(void)
 		ip->name = xstrndup(d->d_name, ext - d->d_name);
 		ip->handle = so;
 
+		if (ip->init_function) ip->init_function();
```
flyingmutant: Again, please split into multiple lines.

flyingmutant:
Thanks for the PR – and sorry for super-delayed review.

utils.h
flyingmutant: @vit1-irk can you please comment on this?

vit1-irk: Yes, I've been thinking about usefulness of these functions.
They have no real purpose: It's just an alias to checkers. At first, it was needed to avoid "cue" and "cdda" mentions in the main code. Readability, you know...

But this code (is_url, is_partial_url) isn't nice for me. Maybe there is another way to get rid of those bad "link" hacks?

flyingmutant: Yes, I do not like these functions either; I'd even say that their existence makes this PR feel unfinished.

vit1-irk: I can revert it back to the previous state. It's impossible to kick out `is_cue_url` and `is_cdda_url` in the current state of codebase. Filetype detection (made using awkward enum) and other specific properties hardly depend on them.

mahkoh:
I don't think this can work. You're removing this code:

```c
#ifdef CONFIG_CUE
       if (!is_cue_url(filename)) {
               if (force || lookup_cache_entry(filename, hash_str(filename)) == NULL) {
                       int done = add_file_cue(filename);
                       if (done)
                               return;
               }
       }
#endif
```

and replace it with a special handler. But this code is the opposite of special. It handles all cases where your special handler does not take action. Namely those cases where the filename does not end in .cue.

Files ending in .cue cannot be added in the first place because of this inversion. Closing.

vit1-irk:
@mahkoh this code was removed because of really confusing recursive calls.

> It handles all cases where your special handler does not take action

Do you mean `lookup_cache_entry`?

mahkoh:
No I mean add_file_cue. The behavior is as follows:

For any file that does not end in .cue: Try to add `file.cue`. If this works ignore the original file.

PR #463, opened 2016-06-22
mpris support

mahkoh:
At long last this PR continues the work started in #90. Thanks to the new sd-bus API it's now much easier to implement MPRIS support in plain C. This allows cmus to integrate with various desktop environments. See the screenshot below. Some minor changes had to be made to the existing code but nothing major. The `track_info` structure now has a `uid` field which is a unique integer and allows each track to be identified by MPRIS clients. Not every feature is implemented. For example, the CanGoNext property always returns true even if there is no next track to play. Implementing this properly would require more refactoring. This can be done in the future.

screenshot from 2016-06-22 16-40-06

### DIFF
--- a/mpris.c
+++ b/mpris.c
@@ -1,0 +1,558 @@
+/*
+ * Copyright 2016 Various Authors
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <systemd/sd-bus.h>
+
+#include "mpris.h"
+#include "ui_curses.h"
+#include "cmus.h"
+#include "player.h"
+#include "options.h"
+#include "output.h"
+#include "track_info.h"
+#include "utils.h"
+
+#define CK(v) \
+do { \
+	int tmp = (v); \
+	if (tmp < 0) \
+		return tmp; \
+} while (0)
+
+static sd_bus *bus;
+int mpris_fd = -1;
+
+static struct track_info *mpris_get_ti(void)
+{
+	player_info_lock();
+	struct track_info *ti = player_info.ti;
+	if (ti)
+		track_info_ref(ti);
+	player_info_unlock();
+	return ti;
+}
+
+static int mpris_msg_ignore(sd_bus_message *m, void *_userdata,
+		sd_bus_error *_ret_error)
+{
+	return sd_bus_reply_method_return(m, "");
+}
+
+static int mpris_read_false(sd_bus *_bus, const char *_path,
+		const char *_interface, const char *_property,
+		sd_bus_message *reply, void *_userdata,
+		sd_bus_error *_ret_error)
+{
+	uint32_t b = 0;
+	return sd_bus_message_append_basic(reply, 'b', &b);
+}
+
+static int mpris_read_true(sd_bus *_bus, const char *_path,
+		const char *_interface, const char *_property,
+		sd_bus_message *reply, void *_userdata,
+		sd_bus_error *_ret_error)
+{
+	uint32_t b = 1;
+	return sd_bus_message_append_basic(reply, 'b', &b);
+}
+
+static int mpris_write_ignore(sd_bus *_bus, const char *_path,
+		const char *_interface, const char *_property,
+		sd_bus_message *value, void *_userdata,
+		sd_bus_error *_ret_error)
+{
+	return sd_bus_reply_method_return(value, "");
+}
+
+static int mpris_identity(sd_bus *_bus, const char *_path,
+		const char *_interface, const char *_property,
+		sd_bus_message *reply, void *_userdata,
+		sd_bus_error *_ret_error)
+{
+	const char *id = "cmus";
+	return sd_bus_message_append_basic(reply, 's', id);
+}
+
+static int mpris_uri_schemes(sd_bus *_bus, const char *_path,
+		const char *_interface, const char *_property,
+		sd_bus_message *reply, void *_userdata,
+		sd_bus_error *_ret_error)
+{
+	static const char * const schemes[] = { "file", "http", NULL };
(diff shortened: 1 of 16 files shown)


mpris.c
```diff
@@ -0,0 +1,558 @@
+/*
+ * Copyright 2016 Various Authors
```
flyingmutant: Don't you want your copyright here?

mpris.c
```diff
@@ -0,0 +1,558 @@
...
+#include "output.h"
+#include "track_info.h"
+#include "utils.h"
+
+#define CK(v) \
+do { \
+	int tmp = (v); \
+	if (tmp < 0) \
+		return tmp; \
+} while (0)
```
flyingmutant: Nitpick, but I am not sure this macro is warranted, given its usage here.

mpris.c
```diff
@@ -0,0 +1,558 @@
...
+static int mpris_msg_append_simple_dict(sd_bus_message *m, const char *tag,
+		char type, const void *val)
+{
+	const char s[] = { type, 0 };
+	CK(sd_bus_message_open_container(m, 'e', "sv"));
+	CK(sd_bus_message_append_basic(m, 's', tag));
+	CK(sd_bus_message_open_container(m, 'v', s));
+	CK(sd_bus_message_append_basic(m, type, val));
+	CK(sd_bus_message_close_container(m));
+	CK(sd_bus_message_close_container(m));
```
flyingmutant: Should't we try to close the containers on the error path, too?

mpris.c
```diff
@@ -0,0 +1,558 @@
...
+		const char *b)
+{
+	CK(sd_bus_message_open_container(m, 'e', "sv"));
+	CK(sd_bus_message_append_basic(m, 's', a));
+	CK(sd_bus_message_open_container(m, 'v', "as"));
+	CK(sd_bus_message_open_container(m, 'a', "s"));
+	CK(sd_bus_message_append_basic(m, 's', b));
+	CK(sd_bus_message_close_container(m));
+	CK(sd_bus_message_close_container(m));
+	CK(sd_bus_message_close_container(m));
```
flyingmutant: Same here.

mpris.h
```diff
@@ -0,0 +1,52 @@
...
+void mpris_playback_status_changed(void);
+void mpris_loop_status_changed(void);
+void mpris_shuffle_changed(void);
+void mpris_volume_changed(void);
+void mpris_metadata_changed(void);
+void mpris_seeked(void);
+
+#else
+
+__attribute__((unused)) static int mpris_fd = -1;
```
flyingmutant: `__attribute__` is GCC-specific, and can be avoided here.

player.c
```diff
@@ -1018,6 +1020,8 @@ void player_init(const struct player_callbacks *callbacks)
 #endif
 	pthread_attr_t *attrp = NULL;
 
+	cmus_mutex_init_recursive(&player_info.mutex);
```
flyingmutant: Can you please explain the need for `_init_recursive` here, as a comment?

utils.h
```diff
@@ -53,6 +53,7 @@
 #define STATIC_ASSERT(cond) \
 	static uint8_t CONCATENATE(_cmus_unused_, __LINE__)[2*(cond) - 1] UNUSED
 
+#define CLEANUP(f) __attribute__((cleanup(f)))
```
flyingmutant: Again, `__attribute__` is GCC-specific. Better avoid it, and use standard `goto err` pattern.

track_info.h
```diff
@@ -128,6 +130,7 @@ void track_info_set_comments(struct track_info *ti, struct keyval *comments);
 
 void track_info_ref(struct track_info *ti);
 void track_info_unref(struct track_info *ti);
+void track_info_unrefp(struct track_info **ti);
```
flyingmutant: With `CLEANUP` gone, there is no need for this, too.

track_info.c
```diff
@@ -46,8 +46,12 @@ static void track_info_free(struct track_info *ti)
 
 struct track_info *track_info_new(const char *filename)
 {
+	static uint64_t cur_uid = 0;
+	cur_uid++;
```
flyingmutant: Does this follow the spec?

> Each track in the tracklist has a unique identifier. The intention is that this uniquely identifies the track within the scope of the tracklist. In particular, if a media item (a particular music file, say) occurs twice in the track list, each occurrence should have a different identifier. If a track is removed from the middle of the playlist, it should not affect the track ids of any other tracks in the tracklist.

flyingmutant:
Thanks a lot for the PR, by the way! I am sure this is a feature many people will enjoy.

mpris.c
mahkoh: These lines have no legal effect. All they do is advertise who wrote the code. And since most cmus files simply use "Various Authors" I'm fine with it here too.

mpris.c
mahkoh: It's unnecessary. The library will discard the reply if we return a negative error code and free all contents.

mpris.h
mahkoh: What is the best way to avoid "unused static" errors in this case?

utils.h
mahkoh: In general, yes. However, the mpris.c file will only be compiled on platforms which have the libsystemd library. This library uses the cleanup attribute extensively internally and sd-bus is designed to be used with this attribute. Hence I think that using it inside mpris.c is appropriate.

track_info.c
mahkoh: cmus does not implement the tracklist feature at the moment. If it ever gets implemented one might have to revisit this point.

mpris.h
flyingmutant: I think the usual trick is `define`. Alternatively, you can create a `mpris_noop.c` file with noop implementations for the `!CONFIG_MPRIS` case.

track_info.c
flyingmutant: Why not simply use integer value of `track_info` pointer then?

mahkoh: That's not race-free. The one point where it's used is in the seek request to ensure that the track hasn't changed in the meantime.

flyingmutant: Sorry, I don't get it: what race is possible when using the pointer value, but not possible with `static int` approach?

mahkoh: If you remove a track, free the track_info, add another track, allocate a new track_info, then that track_info can have the same address as the old one. A 64 bit integer cannot realistically overflow.

flyingmutant: Good point, thanks. And why not full path to track as ID?

mahkoh: dbus paths are restricted to certain ascii characters. So we would have to encode the paths and then ensure that they are still unique. It's possible but not trivial.

player.c
flyingmutant: @mahkoh please do not forget about this one, as the code is non-obvious without the comment.

PR #469, opened 2016-07-01
use a void ptr instead of an unsigned for option multiplexing

mahkoh:
Multiple options can be handled by a single function via an opaque argument that is passed to the handler. This PR changes the type of this argument from `unsigned` to `void *`.

This has the following advantages:

* The code is smaller: This PR removes 200 lines.
* The code is more robust: Plugins used to figure which option to handle by comparing the passed integer to the position of the option name in an array. This meant that the array layout was significant and had to be kept in sync with the option handler functions. The new version uses the same design used inside the `option.c` file where each option gets its own function.
* The new version allows #442 to be implemented with a single list since the position of a plugin in the list is no longer significant and the list can be reordered at runtime.

### DIFF
--- a/aac.c
+++ b/aac.c
@@ -534,4 +534,4 @@ const struct input_plugin_ops ip_ops = {
 const int ip_priority = 50;
 const char * const ip_extensions[] = { "aac", NULL };
 const char * const ip_mime_types[] = { "audio/aac", "audio/aacp", NULL };
-const char * const ip_options[] = { NULL };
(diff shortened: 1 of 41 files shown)


nefthy:
I like it. Switch statements for figuring out which option is past are annoying.

flyingmutant:
@mahkoh can you please rebase this (I've merged #442)?

mahkoh:
This should work now.

flyingmutant:
Excellent!

PR #472, opened 2016-07-02
Locking rework

mahkoh:
This PR removes the editable_lock function.

This has several advantages:

- You no longer have to worry about calling editable_lock
- At least one deadlock has been removed by this change
- ~~The code is a about 20 lines smaler~~

To do this I've changed the way jobs modify the structures previously guarded by this mutex:

1. Jobs are scheduled as before
2. Once the jobs are done, they store the result in a list and notify the ui thread of this
3. The ui thread fetches the result and performs the necessary modifications


job.c
```diff
@@ -311,74 +404,147 @@ void do_add_job(void *data)
...
+static void free_update_job(void *data)
 {
 	struct update_data *d = data;
 
-	free(d->ti);
+	if (d->ti) {
+		for (size_t i = 0; i < d->used; i++)
+			track_info_unref(d->ti[i]);
+		free(d->ti);
+	}
```
mahkoh: this also fixes an unrelated memory leak

mahkoh:
I used a static analysis tool to make sure that editable_lock is only called from the places I fixed here. Unfortunately I missed one instance because it is called through a function pointer there. I've already written a patch and will push it tomorrow.

mahkoh:
This should be ready now.

cmus.c
```diff
@@ -57,33 +57,25 @@ int cmus_init(void)
...
 {
-	worker_remove_jobs(JOB_TYPE_ANY);
-	worker_exit();
+	job_exit();
 	if (cache_close())
 		d_print("error: %s\n", strerror(errno));
 }
 
+static struct track_info *cmus_get_next_from_main_thread(void);
+
```
flyingmutant: I think it is better to not have this forward declaration and simply use `cmus_get_next_track` here.

cmus.c
```diff
@@ -57,33 +57,25 @@ int cmus_init(void)
...
-	info = play_queue_remove();
-	if (info == NULL) {
-		if (play_library) {
-			info = lib_goto_next();
-		} else {
-			info = pl_goto_next();
-		}
-	}
-	editable_unlock();
+	info = cmus_get_next_from_main_thread();
```
flyingmutant: This can be merged with `info` declaration above.

cmus.c
```diff
@@ -164,15 +154,17 @@ enum file_type cmus_detect_ft(const char *name, char **ret)
 	return FILE_TYPE_FILE;
 }
 
-void cmus_add(add_ti_cb add, const char *name, enum file_type ft, int jt, int force)
+void cmus_add(add_ti_cb add, const char *name, enum file_type ft, int jt,
+		int force)
```
flyingmutant: I think it is better to try to keep things like this as single line – even if it crosses some (arbitrary) line length limit.

cmus.c
```diff
@@ -388,3 +376,72 @@ int cmus_playlist_for_each(const char *buf, int size, int reverse,
 	}
 	return 0;
 }
+
+/* multit-hreaded next track requests */
```
flyingmutant: Typo.

cmus.c
```diff
@@ -388,3 +376,72 @@ int cmus_playlist_for_each(const char *buf, int size, int reverse,
...
+
+int cmus_next_track_request_fd;
+static int cmus_next_track_request_fd_priv;
+static pthread_mutex_t cmus_next_file_mutex = CMUS_MUTEX_INITIALIZER;
+static pthread_cond_t cmus_next_file_cond = CMUS_COND_INITIALIZER;
+static int cmus_next_file_provided;
+static struct track_info *cmus_next_file;
```
flyingmutant: I'd prefer to keep globals at the top of the file.

job.c
```diff
@@ -46,27 +47,116 @@
...
+			struct track_info **update_cache_ti;
+		};
+	};
+};
+
+int job_fd;
+static int job_fd_priv;
+
+static LIST_HEAD(job_result_head);
+static pthread_mutex_t job_mutex = CMUS_MUTEX_INITIALIZER;
```
flyingmutant: Please keep globals at the top of the file.

cmus.c
```diff
@@ -388,3 +376,72 @@ int cmus_playlist_for_each(const char *buf, int size, int reverse,
...
+}
+
+static struct track_info *cmus_get_next_from_other_thread(void)
+{
+	static pthread_mutex_t mutex = CMUS_MUTEX_INITIALIZER;
+
+	cmus_mutex_lock(&mutex);
+
+	char buf = 0;
+	write(cmus_next_track_request_fd_priv, &buf, 1);
```
flyingmutant: I'd prefer to have some `notify`-style wrappers around pipe-related read/write calls – it would make the code much more readable.

cmus.c
mahkoh: Some line length limit should be used. If you think 80 is too little, what about 90?

mahkoh:
I've merged the other locking PR into this. All commits are now self-contained, compile, and can be reviewed individually.

cmus.c
flyingmutant: 90 is probably OK.

mahkoh:
I've addressed the style issues and fixed another bug. Namely https://github.com/cmus/cmus/pull/472/commits/b9c641c658af7510426d23a7c872eb09cc67f274

The rest of the commit history is unchanged.

flyingmutant:
Thanks *a lot* for cleanly separated commits — this made reviewing the PR much easier.

PR #475, opened 2016-07-04
make IP priority an option

mahkoh:
This is now based on #469 which should be merged first.


flyingmutant:
FYI, clang now reports lots of
```
   CC     flac.lo
flac.c:591:55: warning: missing field 'set' initializer [-Wmissing-field-initializers]
const struct input_plugin_opt ip_options[] = { { NULL } };
                                                      ^
1 warning generated.
```
warnings.

mahkoh:
Clang is wrong: https://llvm.org/bugs/show_bug.cgi?id=21689

PR #476, opened 2016-07-04
remove public use of player_info_lock

mahkoh:
This PR removes the player_info_lock function. With this change and #472 (which needs to be merged first) almost all cross-file locking has been removed.

This is done as follows:

- The metadata array (4kb) has been moved out of the player_info structure. It is only used in one place and still requires separate locking for each access.
- The player.c file operates exclusively on a new symbol called player_info_priv.
- At the start of each main-loop iteration, the main thread requests a snapshot of the player_info_priv object. This snapshot is then stored in player_info which is owned by the main thread.
- The main thread can then access this symbol without locking.

### DIFF
--- a/cmus.c
+++ b/cmus.c
@@ -23,7 +23,6 @@
 #include "player.h"
 #include "input.h"
 #include "play_queue.h"
-#include "worker.h"
 #include "cache.h"
 #include "misc.h"
 #include "file.h"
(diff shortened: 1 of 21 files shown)


mahkoh:
I realize that this is probably a bit hard to review due to its size and complicated nature. I'll split it up tomorrow into smaller, self-contained commits which will hopefully make it easier.

Issue #108, opened 2014-02-16
[Feature request] multiple playlists/treat playlists like directories


losingkeys:
I'm opening this issue to discuss the option of multiple playlists. If they were stored in a central spot (like ~/.cmus/playlists/), then the third view could just show the directory and have an option to make a new one.

Then if you open a playlist, it would work just like it does now, except there'd be a `../` at the top, like in the file browser, so you could go back and look at more playlists.

Thanks for this awesome music player!

toogy:
Who goes for it?

fin0:
+1 Would be so great. Managing multiple playlists in cmus is just a pain in the ass. I love that cmus has only the basic functionality and no dumb features but playlist support is a very very basic functionality..after that cmus would be close to perfection :)

ald0n:
Totally agree. Zhora zapili, ochen' nado.

toogy:
I switched to mpd + ncmpcpp in the meantime because I was really missing this feature :/

mahkoh:
Friendly reminder that "+1", "me too", and "I'm leaving the fandom :/" won't make it happen. Someone who actually wants this has to sit down for a few hours and implement it. It's probably not even that hard.

losingkeys:
I'm happy to take a crack at it with someone. It's been a few years since I've done C, but I'm sure it'll come back to me quickly :smile:. Email me if you want to pair on this, otherwise I'll open a pull if/when I think I've figured it out (though don't hold your breath if you'd rather solo).

toogy:
@losingkeys, why not simply forking?

losingkeys:
Because this thread is still open for discussion. We just need to come up
with some code :)

z3t0:
Hello, is anyone still interested in working on this? I really like cmus but the lack of multiplaylist support really breaks my workflow. Although I am only an intermediate C programmer I would like to work on this feature if anyone else wants to join let me know!

losingkeys:
I am! I also really want playlists in cmus. Let's discuss the design here
quick.

So far I think the plan is just to have a playlists directory; how about
.cmus/playlists/? To display them we could almost copy the library view,
but the sidebar would show all the playlist names from the playlists
folder, maybe showing a * next to the sidebar selected one? This might be
confusing because you could add to one playlist and play from another.
We'll also need to figure out how to deal with editing, but I'd like to
keep it super simple (like the rest of cmus, just a clean and fast
interface :). Maybe just create/rename/delete for the playlist, and keys to
[...]

PR #485, opened 2016-07-09
Add multiple playlists feature

mahkoh:


losingkeys:
Hmm, If I:

    make clean && ./configure prefix=. DEBUG=2 CONFIG_ROAR=n CONFIG_ALSA=y && make && make install
At 0c687d152e1bf5f7e61c82f7297865817e5fe4cb, I still can't get it to restore the selected playlist on startup. Am I building it wrong?

mahkoh:
You have to enable the resume option.

losingkeys:
Ohh I wasn't talking about the last played _track_, but where the `*` marker is. Is that also tied to the resume option? It seems like they're similar, but different enough that this could just be the default (restoring where the `*`), and the resume option could work across playlists/the file browser/the library/etc.

mahkoh:
There is no mechanism in cmus to store such state except for the resume feature.

mahkoh:
There are still some problems that need to be fixed:

- When activating a playlist and follow is enabled, the cursor moves into the track window. This should not happen.

- When adding tracks from the browser window the cursor does not move.

flyingmutant:
And I am still to review the whole thing — only got around 50% so far.

mahkoh:
Rebased on top of master.

flyingmutant:
Even if there are some unsolved issues (https://github.com/cmus/cmus/pull/485#issuecomment-236263927), this is a *fantastic* PR.

Thanks a lot for your hard work (and especially for splitting everything in nice self-contained commits) — and sorry it took me so long to review and merge.

Fixes #108.

ayancey:
Is there any documentation on this feature? I could not find any.

Yodzeity:
I'm sorry, is this abandoned?

criztovyl:
@Yodzeity Not exactly, [it's merged](https://github.com/cmus/cmus/pull/485#event-758921691) :)

PR #487, opened 2016-07-12
Refactor the short filter expansion

mahkoh:
The old version was written as a single, large state machine-based function that made some questionable assumptions about memory management and was also not particularly fast (not that it matters in this case.) This new version is a more conventional recursive descent parser.

This is part of a larger refactoring of expr.c that decouples it from filter expressions and thus allows it to be used properly in format_print.c.

### DIFF
--- a/short_expr.c
+++ b/short_expr.c
@@ -198,15 +198,12 @@ static int short_expr_quoted_str_arg(struct short_expr *e)
 		else if (c == '\\')
 			esc = true;
 		if (c == 0)
-			break;
-		else
-			e->input++;
-	}
-	size_t len = e->input - start;
-	if (*e->input == '"')
+			return short_expr_err(e, "unterminated string");
 		e->input++;
+	}
+	e->input++;
+	size_t len = e->input - start;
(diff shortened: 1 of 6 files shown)


short_expr.c
```diff
@@ -0,0 +1,381 @@
...
+}
+
+static void short_expr_push_strn(struct short_expr *e, const char *s, size_t n)
+{
+	gbuf_add_bytes(e->output, s, n);
+}
+
+static void short_expr_skip_spaces(struct short_expr *e)
+{
+	while (*e->input == ' ')
```
flyingmutant: Better to use `isspace`.

nefthy:
Is there a reason why this was not merged @flyingmutant? It seems good to me

flyingmutant:
Without "larger refactoring of expr.c" etc. what problem does it solve? The old code is probably much more battle-tested.

nefthy:
The old code is a mess, but I can see your point.

flyingmutant:
It might be unconventional, but definitely not a mess.

PR #493, opened 2016-07-30
Support 24-bit/32-bit encodings in sndio output plugin

t6:
I was trying to play a 24-bit encoded FLAC file in cmus using the sndio output plugin, which failed with an error message saying that it's unsupported.  With just a little bit of extra code it works.

I also removed `sys/audioio.h` which is not needed to compile cmus on OpenBSD 5.9 and does not exist on FreeBSD.

While here I also enabled checking for sndio on FreeBSD in the configure script.

### DIFF
--- a/configure
+++ b/configure
@@ -21,7 +21,7 @@ check_cflags()
 check_sndio()
 {
 	case `uname -s` in
-	OpenBSD)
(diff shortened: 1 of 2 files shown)


flyingmutant:
Thanks a lot!

PR #502, opened 2016-08-21
Add lib_add_filter option

tsujamin:
This patch allows for a filter to be applied when adding files to the library view. An example use-case would be music libraries with multiple formats, or those with cue files resulting in files being added twice.

I'm open to any feedback on the structure of the change; in particular I think my {get,set}_lib_add_filter functions are a bit logic heavy for where they are in the code base.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -991,6 +991,9 @@ icecast_default_charset (ISO-8859-1)
 
 	NOTE:  This is used only if the metadata is not valid UTF-8.
 
+lib_add_filter (`Filter`)
+	Apply filter when adding to library view. See *FILTERS*.
+
 lib_sort (artist album discnumber tracknumber title filename) [`Sort Keys`]
 	Sort keys for the sorted library view (2).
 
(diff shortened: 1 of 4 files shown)


Doc/cmus.txt
```diff
@@ -991,6 +991,9 @@ icecast_default_charset (ISO-8859-1)
 
 	NOTE:  This is used only if the metadata is not valid UTF-8.
 
+lib_add_filter (`Filter`)
+	Apply filter when adding to library view. See *FILTERS*.
```
flyingmutant: IMO «… when adding *files* to *the library*.» is a bit better.

options.c
```diff
@@ -1079,6 +1080,32 @@ static void toggle_mpris(void *data)
...
+
+		if (!expr)
+			return;
+	}
+
+	if (lib_add_filter != NULL)
+		free(lib_add_filter);
+
+	lib_add_filter = malloc(strlen(buf) + 1);
+	strcpy(lib_add_filter, buf);
```
flyingmutant: `xstrdup`

mahkoh:
@tsujamin Can you please address the comments?

tsujamin:
Sorry, been off the radar, will address when next on my laptop

tsujamin:
@mahkoh addressed

mahkoh:
Needs rebase.

tsujamin:
No problem, rebased and tested against current master

mahkoh:
Merged as 8435f19d4aeb8dd588e6fd3c96d184b810004062 with some fixes.

Issue #467, opened 2016-06-28
Add "browser" playback mode


avoidr:
Would be cool if there was this mode, so when you're in the browser view and start playing a track, it keeps playing tracks from the folder. Like playing a song from a playlist would switch to playlist mode and play tracks from there.

tsujamin:
I'm interested in doing some work on this. 

I'll start by refactoring the play_library option to cycle through "play modes", such as library, playlist and eventually browser, as I don't think the current implementation would allow for a 3rd play mode.

Does this seem like a reasonable place to start?

avoidr:
Thank you for being interested in doing some work on this.

mahkoh:
What I'd like to see here is some kind of "track stream" abstraction. A "track stream" is simply a circular list of tracks that determines what happens when the next/the previous track is requested.

Currently such a thing does not exist. Instead, whenever the next track is requested, the currently playing view is told to go to the next track and to return the track info of that track. This means that it's impossible to see what the next/previous tracks are going to be without actually playing them.

Any implementation of this feature should first refactor the old code to add such an abstraction.

bitwombat:
This seems like a really useful feature. My mp3's on disk are all organized by album, so often I just want to play from the browser view, without messing with the library, queues, or playlists.

agenbite:
I'm also interested in this, although my programming skills are not so strong.

It could also be solved with a chain of commands (`clear -q`, `win-add-q {}`, `player-next`), but there's no way to do this, afaik.

agenbite:
I've just learnt [an obvious way](https://medium.com/@johnnymatthews/how-to-update-your-cmus-library-73b1282f7d51) to chain commands: creating a script that  launches them sequentially in their remote version.

Thus, my script to play files from the browser view is:

```
#!/bin/bash

cmus-remote -C "clear -q"
cmus-remote -C "win-add-q {}"
sleep 0.2
cmus-remote -C player-next
```

Note the `sleep` command, which prevents race conditions.

I've binded this script to ENTER like so:

```
:bind browser enter shell /path/to/script/playFromBrowser.sh
```

It works. :)

PR #504, opened 2016-08-21
play_mode refactor

tsujamin:
These patches refactor the _play_library_ option to a new _play_mode_ enumeration. 

This removes the limitation on the number play modes, due to the boolean nature of _play_library_, and allows for the addition of future play modes such as those requested in #467.

The options exposed by the server component have been changed by these patches. Is this an issue?

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1012,6 +1012,10 @@ pl_sort () [`Sort Keys`]
 	Sort keys for the playlist view (3).  Empty value disables sorting and
 	enables manually moving tracks.
 
+play_library (true)
+	Play tracks from the library instead of playlist. This setting changes
+	*play_mode* and is kept for backwards compatability.
+
 play_mode (library playlist)
 	Play tracks from the library or from the playlist.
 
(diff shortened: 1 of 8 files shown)


mahkoh:
Doesn't this break every existing installation?

tsujamin:
Wow, apparently that went entirely over my head as I was writing it 😔 my bad.

I'll have a bit of a think how this could be done without modifying existing settings, but I'm guessing that might be difficult given play_library presently acts as a Boolean value...

mahkoh:
You can just leave play_library as is and add play_mode as a new option. Changing play_library then simply changes play_mode.

tsujamin:
Alright, will update tomorrow.

tsujamin:
Implemented your advice, appears to be working. 

Is this something you're interested in merging as-is, or after a use case for it (the browser play_mode in #467) has been implemented?

flyingmutant:
I'd prefer to wait until the followup PR, as this one by itself is not «complete».

nefthy:
closing due to old age.

PR #510, opened 2016-09-02
Make plugins debuggable without installation

mahkoh:
1. Move plugins to `ip`/`op` directories

2. Have cmus check if certain environment variables are set which determine the location of plugins/data.

3. Add a small wrapper that sets those variables in the source directory.


Makefile
```diff
@@ -263,15 +267,16 @@ quiet_cmd_ttman = MAN    $@
...
 
-main: cmus cmus-remote
+main: cmus cmus-remote cmus-bin
 plugins: $(ip-y) $(op-y)
 man: $(man1) $(man7)
 
 install-main: main
-	$(INSTALL) -m755 $(bindir) cmus cmus-remote
+	$(INSTALL) -m755 $(bindir) cmus-bin cmus-remote
+	@mv $(bindir)/cmus-bin $(bindir)/cmus
```
mahkoh: Not super clean.

mahkoh:
Not really a fan of this btw. But I don't know a better solution.

flyingmutant:
I don't like e79741efbebbc5bf601978f547b2700f58da2868 and `cmus-local`/`cmus-bin` names.

`CMUS_LIB_DIR=./ CMUS_DATA_DIR=./data ./cmus` is not that hard to type by hand during development (esp. with shell history).

mahkoh:
The idea is that the right thing happens automatically for people who don't know about those variables.

flyingmutant:
As this is only for development, I don't think e79741e is worth the additional complexity.

Without the variables specified, the behavior should be the same as it is now. And mentioning them in `README` should be enough for developers IMO.

nefthy:
I agree with @flyingmutant. I like the ip/op directories though.

mahkoh:
>And mentioning them in README should be enough for developers IMO.

For developers. But I've seen users struggling with this several times. I.e. they compile from source and try to run the local version without installing it first.

Anyway, I've dropped the last commit for now.

nefthy:
A shell script in the source dir that does not get installed, should do the trick for this usecase. Something like run-cmus-debug.sh:

- check if cmus binary exists in the same die as the shell script or print error and compile instructions
- run cmus form current dir with environment variables set accordingly.

mahkoh:
Unfortunately that does not solve the problem.

1. People still have to know that they have to run the script.

2. Scripts can't be used as debugger arguments.

flyingmutant:
LGTM (after resolving the conflicts).

PR #513, opened 2016-09-06
Some main_loop improvements

mahkoh:


player.c
```diff
@@ -512,6 +523,7 @@ static void _producer_buffer_fill_update(void)
 /* 		d_print("\n"); */
 		player_info_priv.buffer_fill = fill;
 		player_info_priv.buffer_fill_changed = 1;
+		player_info_changed();
```
mahkoh: This is not so great because it happens too often.

flyingmutant: Yeah, looks like a hack to be honest. And I think this can happen both too fast and too slow. Why did you decide to remove the timeout?

mahkoh: >too slow

I'm not sure what you mean by this.

>Why did you decide to remove the timeout?

Let's for the moment ignore the buffer information which I didn't consider at the time (and which the main_loop comment does not mention.) The main loop iterates ten times per second which is nine times too often because the highest resolution UI element is the playback time which only needs to be updated once per second. On the other hand the UI can lag behind the real playback time by up to 100ms.

An implementation in which the player informs the UI that something has changed is both cleaner and more efficient.

With the buffer information it's a bit more complicated. Maybe it's sufficient to move this call out of this function (which is called in a loop elsewhere until the buffer is filled.)

mahkoh:
I don't see a good way to do this so I've drooped the last three commits.

PR #519, opened 2016-09-14
Fix manpage formatting

mahkoh:


flyingmutant:
Thanks!

By the way, I am terribly sorry for lots of PR/issues not getting attention they require right now. Because of both personal and work-related things, it is very hard for me to find the time needed.

mahkoh:
It doesn't matter to me how long it takes.

PR #522, opened 2016-09-16
Skip the TI cache if it's locked for over 100ms

### LINKED ISSUE #521: Cache updates should not hold the cache lock indefinitely
Trying to play a track in the browser view during a cache update will block the UI until the update has been completed.

mahkoh:
Fixes #521


mahkoh:
This needs to use the "-lrt" flag but I'm not what's the best way to add this.

flyingmutant:
To be honest I don't like this — it feels like a bit like hack/workaround.  Solving #521 (by not holding the cache lock too long) should make this unnecessary, too.

mahkoh:
>Solving #521 (by not holding the cache lock too long)

That alone won't solve the problem. E.g.

```
loop {
  // work
  if we've run for over 100ms:
    release lock
    acquire lock
}
```

does not guarantee that the UI thread will actually get the lock. Instead we'd have to write a new synchronization primitive that has the desired FIFO properties.

mahkoh:
Closing in favor of #534.

PR #526, opened 2016-09-20
support 24bit playback for flac

Wang-Yue:
There's currently a bug in player. Seeking 24bit doesn't work for now.

@flyingmutant  any idea why?

### DIFF
--- a/flac.c
+++ b/flac.c
@@ -190,12 +190,10 @@ static int eof_cb(const Dec *dec, void *data)
 
 #if defined(WORDS_BIGENDIAN)
 
-#define LE16(x) swap_uint16(x)
 #define LE32(x) swap_uint32(x)
 
 #else
 
-#define LE16(x)	(x)
 #define LE32(x)	(x)
 
 #endif
@@ -207,6 +205,7 @@ static FLAC__StreamDecoderWriteStatus write_cb(const Dec *dec, const FLAC__Frame
 	struct flac_private *priv = ip_data->private;
 	int frames, bytes, size, channels, bits, depth;
 	int ch, nch, i, j = 0;
+	char *dest; int32_t src; const char *src_ptr;
 
 	if (ip_data->sf == 0) {
 		return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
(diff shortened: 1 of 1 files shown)


Wang-Yue:
comment:

1. write_cb performance is slightly worse than the original, which is expected --- previous version just copy 32 bit integer, but this version has to convert 32bit to 24. other bits (8, 16, 32) playback is also slightly worse. but the performance is not a bottleneck, as flac decoding takes much longer time than write_cb. 
2. while write_cb takes a little bit more time, it takes less time for output plugin, and system audio daemon to run (coreaudiod on Mac OS X for instance).
3. seek doesn't work. I don't know why. perhaps we have the wrong read/write position? 
4. On Linux when selecting raw hardware (hw:1:0 for instance), previous version doesn't work since my sound card doesn't support 32 bit. Now it works.

mahkoh:
>There's currently a bug in player. Seeking 24bit doesn't work for now.

Works for me.

Wang-Yue:
@mahkoh what output/input plugin and os version are you using? I want to isolate the problem.

mahkoh:
Pulse 9.0, linux, Flac with your patch applied. I converted an mp3 to 16/24 bit flac and didn't notice a difference in seeking behavior.

Wang-Yue:
OK, I just confirmed it's not a flac plugin bug. 

24bit wav has same issue on my Mac, which I will fix it soon. 

The PR is ready to merge.

flac.c
```diff
@@ -235,50 +234,17 @@ static FLAC__StreamDecoderWriteStatus write_cb(const Dec *dec, const FLAC__Frame
...
+	dest = priv->buf + priv->buf_wpos;
+	for (i = 0; i < frames; i++) {
+		for (ch = 0; ch < channels; ch++) {
+			src = LE32(buf[ch % nch][i] << (bits - depth));
+			src_ptr = (void *)&src;
+			for (j = 0; j < bits / 8; j++) {
+				*dest = *src_ptr;
+				dest++;
+				src_ptr++;
+			}
```
mahkoh: Why not just memcpy?

Wang-Yue: done.

PR #528, opened 2016-09-20
support bit swapping on Mac

Wang-Yue:


mahkoh:
Both clang and GCC know how to optimize

```c
unsigned swap(unsigned x)
{
	return ((((x) & 0xff000000) >> 24)
		| (((x) & 0x00ff0000) >>  8)
		| (((x) & 0x0000ff00) <<  8)
		| (((x) & 0x000000ff) << 24));
}
```

to the corresponding CPU instruction. Namely

```c
swap:
	movl	%edi, %eax
	bswap	%eax
	ret
```

If anything I would remove the special cases.

Wang-Yue:
@mahkoh I definitely believes CMUS should embrace modern compilers and modern C standard. It's a huge hassle to develop nowadays,  as old school things like forward declaration drives me mad. 

Old systems will have old versions of cmus in their software repo. We need to say goodbye to those systems in HEAD. Migrating to C99 or C11. We should drop old ffmpeg, flac library as well. 

I also think moving to C++ is a reasonable move --- input/output plugins have strong OO design patterns. C++ has various standard/defacto standard library for various verbose operations in CMUS (such as atomicity support, data structures). I believe it would make the project a lot smaller, more maintainable this way.

mahkoh:
Can this be closed then?

Wang-Yue:
> Can this be closed then?

For completeness sake, no.
I don't think we should relies on compiler implementation either. If OS provides the function, just use it.

flyingmutant:
While this PR is «harmless» in the sense that it stays consistent with the way code is written, and does not break anything, it does not make anything better either. It adds (little) complexity for no value.

I'd rather merge a PR which provides a single `bswap` definition for all platforms, and drops all `#ifdef` magic.

PR #511, opened 2016-09-03
Release fixes

mahkoh:
Addresses the issues raised in #507 and some more.

Coincidentally 3a8091b might (or might not) fix #505.

PR #534, opened 2016-09-22
Yield cache mutex during cache updates

mahkoh:


mahkoh:
This still needs some work to avoid double frees.

cache.c
```diff
@@ -467,12 +467,21 @@ struct track_info **cache_refresh(int *count, int force)
...
 	for (i = 0; i < n; i++) {
 		unsigned int hash;
 		struct track_info *ti = tis[i];
 		struct stat st;
 		int rc = 0;
+		uint64_t now;
+
+		now = time_in_ms();
+		if (now - last_break > 100) {
+			cache_unlock();
```
mahkoh: On linux at the very least pthread_mutex_unlock reduces to a single atomic operation if no other thread is waiting.

mahkoh:
This should be ready up to a rebase on #511.

cache.c
```diff
@@ -465,14 +468,22 @@ struct track_info *cache_get_ti(const char *filename, int force)
...
 		struct track_info *ti = tis[i];
 		struct stat st;
 		int rc = 0;
+		uint64_t now;
+
+		now = time_in_ms();
+		if (now - last_break > 100) {
+			cache_yield();
+			last_break = time_in_ms();
+		}
```
mahkoh: Maybe it's better to yield unconditionally here. cache_yield is pretty efficient.

flyingmutant:
Aren't you trying to be a little bit too clever with all stdatomic stuff? It probably could all be done with old boring pthread stuff, with performance being same for all practical purposes.

Another important thing is that Travis builds should keep working. It is a huge timesaver for the maintainer -- you don't have to check for yourself if code compiles or not. It may be the case that we'll need to fit Docker inside Travis in order to use more up to date compilers (if we will require those).

flyingmutant:
Oh, I see, you've updated Travis config in #511, great.

mahkoh:
>Aren't you trying to be a little bit too clever with all stdatomic stuff?

I don't see another way to do this while guaranteeing the fifo property.

>It is a huge timesaver for the maintainer -- you don't have to check for yourself if code compiles or not.

Oh but you do. For example: This morning I had to manually fetch PR #502 and fix it because GCC and clang just issued a warning that would have resulted in UB. Luckily I remembered that `get_lib_add_filter` has to take a `size` argument.

nefthy:
I think this issue can be solve by setting the appropriate scheduling policies for the threads. SCHED_BATCH  for the cache_refresh task and SCHED_FIFO or SCHED_RR for the other tasks. Then a simple unlock/lock every so often (maybe even after each record) should do the trick.

mahkoh:
Please explain.

mahkoh:
To be clear: I think the code is perfectly fine. I realize that dealing with atomics and implementing synchronization structures is not something most people do. But if you've ever implemented a condition variable then the code here doesn't seem to be far fetched.

The structure as written has three nice properties:

1. When a thread tries to acquire the lock, no other thread can prevent it from getting a place in the queue.
2. No thread can jump ahead in the queue.
3. The queue moves forward whenever the lock is unlocked or yielded.

Using a mutex to manage the queue will destroy the first property.

Messing with scheduling policies is a huge overkill and I don't even know if it would solve the problem.

nefthy:
> Please explain.

with the default policy the thread releasing the lock can acquire it again in the next instance. By changing the policy that can be prevented. When the cache_refresh thread is set to SCHED_BATCH it will only get the lock if no higher priority thread is asking for it. At least that is my understanding. 

> Messing with scheduling policies is a huge overkill and I don't even know if it would solve the problem.

Its part of the pthreads POSIX API, see pthread_attr_setschedpolicy(3). Sure this can get messy, but as is, we have the cache_refresh thread that does a batch job and everything else that has priority over it.

I personally implementing synchronizing primitives quite a messy and tricky business and would prefer a solution that avoids this sort of thing. However I'm not religious about it either.

mahkoh:
>When the cache_refresh thread is set to SCHED_BATCH it will only get the lock if no higher priority thread is asking for it.

Where does it say that?

>Its part of the pthreads POSIX API

SCHED_BATCH is not part of POSIX. 

>we have the cache_refresh thread that does a batch job and everything else that has priority over it.

There is no cache_refresh thread. There is only a worker thread and you're trying to apply a global solution to a local problem that only exists in one function.

>I personally implementing synchronizing primitives quite a messy and tricky business

I wouldn't call 21 lines of code with lots of whitespace messy.

nefthy:
well I just wanted to propose an alternative solution, if you don't think it is any good, that's fine with me.

track_info.c
```diff
@@ -52,16 +48,18 @@ struct track_info *track_info_new(const char *filename)
 	uint64_t uid = atomic_fetch_add_explicit(&cur_uid, 1, memory_order_relaxed);
 	BUG_ON(uid == 0);
 
-	struct track_info *ti;
-	ti = xnew(struct track_info, 1);
-	ti->uid = uid;
+	struct track_info_priv *priv = xnew(struct track_info_priv, 1);
+	atomic_store_explicit(&priv->ref_count, 1, memory_order_relaxed);
```
flyingmutant: Why a simple assignment is not enough here? Compiler will complain?

mahkoh: A simple assignment is equivalent to

```c
atomic_store_explicit(&priv->ref_count, 1, memory_order_seq_cst);
```

flyingmutant: It will do the job, right? As I understand it, we can't have any races here, so a simple assignment is good enough. Is there any reason to use explicit atomic instructions?

mahkoh: >As I understand it, we can't have any races here, so a simple assignment is good enough.

You're misunderstanding. A simple assignment has stronger synchronization properties than the code above. Compare:

```c
#include <stdatomic.h>

struct X {
	_Atomic int j;
};

void f(struct X *x)
{
	x->j = 0;
}

void g(struct X *x)
{
	atomic_store_explicit(&x->j, 0, memory_order_relaxed);
}
```

```asm
f:
	xorl	%eax, %eax
	xchgl	%eax, (%rdi)
	retq
g:
	movl	$0, (%rdi)
	retq
```

cache.c
mahkoh: Dropped.

track_info.c
mahkoh: I don't think atomics should be used with plain operations it's unnecessary magic that's also a bit misleading. E.g. in the functions above you get a synchronization barrier, but consider:

```c
void h(struct X *x)
{
	*x = (struct X) { 0 };
}
```

```asm
h:
	movl	$0, (%rdi)
	retq
```

No barrier if the plain assignment is made as part of a compound assignment.

mahkoh: I've changed the code so that this issue is no longer relevant.

flyingmutant:
LGTM (given that I convince myself that `fifo_mutex_relock` is working :-), do you consider this finished?

mahkoh:
Yes.

PR #538, opened 2016-09-25
cache device_stereo_channels information, and add mixer_fds to coreaudio

Wang-Yue:
which will make volume querying a little bit faster


flyingmutant:
This fixes

> It happens that CoreAudio's AudioObjectGetPropertyData is pretty slow and heavy. So we end up with a lot higher CPU usage. On my machine (a low end Macbook Pro) it went from 0.3% usage to 0.5%.

right?

Wang-Yue:
Right.

Wang-Yue:
This PR reduced coreaudiod cpu usage from 0.5% to 0.1%

PR #539, opened 2016-09-25
Drop old flac version support.

Wang-Yue:


flyingmutant:
What distributions will be affected by this change?

Wang-Yue:
FLAC_API_VERSION_CURRENT was added 10 years ago, See https://github.com/xiph/flac/commit/833f832ab1eba3b5784009c5d3f49e5c2adaed4f

1.1.3 was released in Nov 2006.

A distribution release has to be having a flac version  at least 10 years old from now, and is still being maintained, to cause breakage. I just can't imagine who would do that. 

The maximum support time for a ubuntu distribution is 5 years. The longest known support ever is windows xp, whose mainstream support date from 2001 to 2009, and extended support end at 2014. But XP is different is each server pack is a significant update from previous versions, and XP without server pack has not been supported since 2005. Also XP does not bundle flac libraries.

Wang-Yue:
also please take a look at https://github.com/cmus/cmus/commit/379dff769de03c5fbf6e5e083e55ac3f36985d7d where these bunch of stuffs were added.

PR #540, opened 2016-09-28
Improve fifo_mutex performance

mahkoh:
This makes the code both simpler and faster.

### DIFF
--- a/locking.c
+++ b/locking.c
@@ -65,7 +65,7 @@ void cmus_rwlock_unlock(pthread_rwlock_t *lock)
 		BUG("error unlocking mutex: %s\n", strerror(rc));
 }
 
-static void fifo_mutex_relock(struct fifo_mutex *fifo, bool holding_lock)
+void fifo_mutex_lock(struct fifo_mutex *fifo)
 {
 	struct fifo_waiter self = {
 		.cond = PTHREAD_COND_INITIALIZER,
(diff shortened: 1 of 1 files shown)


mahkoh:
Fun fact: In the uncontended case fifo_yield is now faster than a simple cmus_mutex_unlock+cmus_mutex_lock.

flyingmutant:
How do you check the correctness (other than reading the source, of course)?

mahkoh:
I go by the following quote:

>Write down the problem. Think real hard. Write down the solution.

Jokes aside I only ready the code and think about it for some time. Maybe try to reformulate the code in terms of mutexes. The code here can be simplified to

```c
void fifo_mutex_lock(struct fifo_mutex *fifo)
{
	struct fifo_waiter self = { 0 };

	struct fifo_waiter *old_tail = swap(&fifo->tail, &self);
	if (old_tail)
		old_tail->next = &self;

	lock(fifo->mutex);
	if (old_tail) {
		while (fifo->head != &self)
			wait();
	}

	fifo->head = NULL;
	if (!cmpxchg(&fifo->tail, &self, NULL)) {
		while (!(fifo->head = self.next))
			;
	}
}
```

and with locks

```c
void fifo_mutex_lock(struct fifo_mutex *fifo)
{
	struct fifo_waiter self = { 0 };

	lock();
	bool old_tail = fifo->tail != NULL;
	if (old_tail)
		fifo->tail->next = &self;
	fifo->tail = &self;
	unlock();

	lock(fifo->mutex);
	if (old_tail) {
		while (fifo->head != &self)
			wait();
	}

	lock();
	fifo->head = self.next;
	if (!fifo->head)
		fifo->tail = NULL;
	unlock();
}
```

Where `lock()` protects `fifo->tail` and `self.next`. `fifo->head` is protected by `lock(fifo->mutex)`.

mahkoh:
This can actually be even simpler:

```c
void fifo_mutex_yield(struct fifo_mutex *fifo)
{
	if (fifo->head || atomic_load_explicit(&fifo->tail, memory_order_relaxed)) {
		fifo_mutex_unlock(fifo);
		fifo_mutex_lock(fifo);
	}
}
```

The check `fifo->head` is superfluous because `fifo->head != NULL` implies `fifo->tail != NULL`.

PR #542, opened 2016-09-29
Adds 24bit support to op/jack

nefthy:


op/jack.c
```diff
@@ -91,6 +91,14 @@ static jack_default_audio_sample_t read_sample_le16(const char *buffer)
 	return (jack_default_audio_sample_t)s / (jack_default_audio_sample_t)upper_bound;
 }
 
+static jack_default_audio_sample_t read_sample_le24(const char *buffer)
+{
+	uint32_t u = read_le24(buffer);
+	int32_t s =  u & 0x800000 ? u | 0xFF000000 : u;
```
mahkoh: I don't like it. Better use something like `read_le24i`.

op/jack.c
```diff
@@ -403,6 +418,10 @@ static int op_jack_open(sample_format_t sf, const channel_position_t *cm)
 	if (bits == 16) {
 		sample_bytes = 2;
 		read_sample = sf_get_signed(sf) ? &read_sample_le16 : &read_sample_le16u;
+	} else if (bits == 24) {
+		d_print("signed: %d", sf_get_signed(sf));
```
mahkoh: That doesn't belong here.

utils.h
```diff
@@ -245,6 +245,14 @@ static inline uint32_t read_le32(const char *buf)
 	return b[0] | (b[1] << 8) | (b[2] << 16) | (b[3] << 24);
 }
 
+static inline uint32_t read_le24(const char *buf)
+{
+	const unsigned char *b = (const unsigned char *)buf;
+
+	uint32_t a = b[0] | (b[1] << 8) | (b[2] << 16);
+	return a;
```
mahkoh: The temporary should be removed.

utils.h
```diff
@@ -245,6 +245,21 @@ static inline uint32_t read_le32(const char *buf)
...
+	const unsigned char *b = (const unsigned char *)buf;
+
+	return b[0] | (b[1] << 8) | (b[2] << 16);
+}
+
+static inline int32_t read_le24i(const char *buf)
+{
+	const unsigned char *b = (const unsigned char *)buf;
+
+	uint32_t a = b[0] | (b[1] << 8) | (b[2] << 16);
```
mahkoh: This can use read_le24.

nefthy:
@mahkoh I addressed everything, thanks for reviewing.

Issue #251, opened 2015-02-16
Incorrect determination of the duration of the tracks in some CUE


komarevtseff:
For example CUE:
https://bpaste.net/show/89d2ed623967
In the first three tracks duration faithful, and then there is no.

For clarity screenshots for this album in DeaDBeeF (true) and cmus (false):
![2015-02-16-062548_1366x751_scrot](https://cloud.githubusercontent.com/assets/3647789/6206903/8c52db0e-b5a5-11e4-8d66-4e71f96a9f87.png)
![2015-02-16-062532_1088x456_scrot](https://cloud.githubusercontent.com/assets/3647789/6206919/157b2daa-b5a6-11e4-9216-18e153244252.png)

flyingmutant:
Thanks for reporting! We use `libcue` for CUE parsing — and it looks like this is another little problem with it.

flyingmutant:
Pasting the CUE in question here for reference:

```
REM ACCURATERIPID 0016fab2-00f67491-c30ac90e
REM GENRE "Alternative Rock/Nu Metal/Rap Rock"
REM DATE 2000
PERFORMER "Linkin Park"
TITLE "Hybrid Theory"
FILE "Linkin Park - Hybrid Theory.wv" WAVE
  TRACK 01 AUDIO
    TITLE "Papercut"
    PERFORMER "Linkin Park"
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    TITLE "One Step Closer"
    PERFORMER "Linkin Park"
    INDEX 01 03:05:40
  TRACK 03 AUDIO
    TITLE "With You"
    PERFORMER "Linkin Park"
    INDEX 01 05:41:17
  TRACK 04 AUDIO
    TITLE "Points of Authority"
    PERFORMER "Linkin Park"
    INDEX 01 09:04:35
  TRACK 05 AUDIO
    TITLE "Crawling"
    PERFORMER "Linkin Park"
    INDEX 00 12:15:59
    INDEX 01 12:24:55
  TRACK 06 AUDIO
    TITLE "Runaway"
    PERFORMER "Linkin Park"
    INDEX 00 15:46:41
    INDEX 01 15:53:55
  TRACK 07 AUDIO
    TITLE "By Myself"
    PERFORMER "Linkin Park"
    INDEX 00 18:57:49
    INDEX 01 18:57:50
  TRACK 08 AUDIO
    TITLE "In the End"
    PERFORMER "Linkin Park"
[...]

PR #544, opened 2016-10-05
Remove libcue dependency

mahkoh:
Not ready but can be tested.

### DIFF
--- a/cue.c
+++ b/cue.c
@@ -1,0 +1,497 @@
+/*
+ * Copyright 2016 Various Authors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdlib.h>
+#include <stdbool.h>
+#include <ctype.h>
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+#include "cue.h"
+#include "xmalloc.h"
+#include "file.h"
+#include "list.h"
+
+#define ASCII_LOWER_TO_UPPER(c) ((c) & ~0x20)
+
+struct cue_track_proto {
+	struct list_head node;
+
+	uint32_t nr;
+	int32_t pregap;
+	int32_t postgap;
+	int32_t index0;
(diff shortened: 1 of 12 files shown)


Wang-Yue:
does this handle index 00 case?

mahkoh:
It handles all cases in which index0 or index1. If one of them is missing and pregap is defined then the other one is calculated. It also adjusts for postgap. The results are somewhat different from what libcue produces. Maybe libcue ignores postgap. I have not looked into this.

Wang-Yue:
LGTM

Wang-Yue:
I tested your PR with many of my files and all seem fine by now.

mahkoh:
Call, thanks. This won't go in before the next stable release, though. Unless @flyingmutant disagrees.

flyingmutant:
I think it is OK to include it in the next release. It is a relatively localised change, which is unlikely to break much. And we'll probably catch the majority  of the errors and corner cases during -rc.
Do you consider it ready for review/merge?

mahkoh:
No, I still have some local changes and need to split the commits.

mahkoh:
Should be ready.

mahkoh:
It should also be confirmed that this is in line with #251.

cue.c
```diff
@@ -0,0 +1,495 @@
...
+	for (size_t i = 0; i < a_len; i++) {
+		if ((a[i] & ~0x20) != (b[i] & ~0x20))
+			return false;
+	}
+	return true;
+}
+
+static void cue_skip_spaces(struct cue_parser *p)
+{
+	while (p->len > 0 && (*p->src == ' ' || *p->src == '\t'))
```
flyingmutant: Why not `isspace`?

cue.c
```diff
@@ -0,0 +1,495 @@
...
+{
+	p->err = true;
+}
+
+static bool cue_str_eq(const char *a, size_t a_len, const char *b, size_t b_len)
+{
+	if (a_len != b_len)
+		return false;
+	for (size_t i = 0; i < a_len; i++) {
+		if ((a[i] & ~0x20) != (b[i] & ~0x20))
```
flyingmutant: Comment or named constant for `~0x20` would make the reading a bit easier.

cue.c
```diff
@@ -0,0 +1,495 @@
...
+
+	if (p->len > 0) {
+		char c = *p->src;
+		cue_consume(p);
+		if (p->len > 0 && c == '\r' && *p->src == '\n')
+			cue_consume(p);
+	}
+}
+
+static char *cue_strdup(const char *start, size_t len)
```
flyingmutant: Why?

cue.c
mahkoh: Because isspace matches \n.

cue.c
mahkoh: Why not?

cue.c
mahkoh: Updated.

Wang-Yue:
you also need to modify .travis.yml and https://github.com/cmus/cmus.github.io/blob/4d1d49fae9bf01fd6a9950a172b5e610e17916be/index.html

Wang-Yue:
lgtm

Issue #554, opened 2016-10-12
factivate outside of FILTERS_VIEW


s3rb31:
Hello,

according to the man page the console command **factivate** should disable all the filters, if no argument is passed:  "If no arguments given then all filters are unactivated."

But I noticed that due to the nature of it's implementation, it would try to edit one filter after another, if no filter has changed between the calls to the function (e.g. no filter activated via view 6 or keybindings).

That of course is utterly annoying. When I am in view 1 or 2 and I am not sure if there are active filters, I will press a keybinding (factivate with no args) to disable all the filters. But if there was no filter active, this would take me to the console to edit a filter. This of course blocks any further interaction with CMus because everything will end up in the console. 

I looked into the the code and found the corresponding function which is responsible for the odd behaviour. A relatively small patch fixes the issue:

```
--- a/filters.c	2016-10-12 16:57:59.859825214 +0200
+++ b/filters.c	2016-10-12 17:03:02.147423231 +0200
@@ -27,6 +27,7 @@
 #include "file.h"
 #include "ui_curses.h"
 #include "xmalloc.h"
+#include "options.h"
 
 #include <stdio.h>
 #include <ctype.h>
@@ -142,8 +143,14 @@
 		if (f->act_stat != f->sel_stat)
 			unchanged = 0;
 	}
-	if (unchanged)
-		edit_sel_filter();
+
+    if (unchanged)
+    {
+        if (cur_view == FILTERS_VIEW)
+		    edit_sel_filter();
+        else
+            return;
+    }
 
 	/* mark visited and AND together all selected filters
 	 * mark any other filters unvisited */
```

If wanted i will create a pull request, but anyone might feel free to implement this fix. 

Thanks for the great software and regards,

s3rb31

S3RB31:
This is not a feature request. There is no feature requested but rather one found. xD

The behaviour of the _factivate_ command is caused by the fact that both the _cmd_factivate_ function and _cmd_win_activate_ eventually will call _filters_activate_. 

_cmd_win_activate_ of course will only do it, if the current view is equal to FILTERS_VIEW. But factivate works in all views.

I don't think that it is intended to call _edit_sel_filter_ from any view but from the FILTERS_VIEW. If this is intentional, than please lets state this somewhere in the manpage.

mahkoh:
I can't reproduce the issue. Please provide a sequence of commands that causes this.

S3RB31:
Oh, you were slightly too fast. ;)

```
[...]

PR #555, opened 2016-10-13
Fix for issue #554

s3rb31:
I looked through the issues listed in #554 and this should resolve them.

Just some minor fixes to differentiate between the two commands within _filters_activate_. Would be nicer with _bool_ from my point of view, but on the other hand there is really no need for the additional include.

Regards


mahkoh:
Looks fine except that it doesn't follow the code style.

S3RB31:
Yeah sorry, my editor made it all look nice and uniform. Now it should be fine.

PR #556, opened 2016-10-18
Make filepath affect album separation.

forgottenswitch:
Previously, directories 'a/c' and 'b/c'
would be considered the same 'c' album:

  c:------------
  a/c/1.wav
  b/c/1.wav
  a/c/2.wav
  b/c/2.wav

### DIFF
--- a/options.c
+++ b/options.c
@@ -1284,15 +1284,13 @@ static void set_format(void *data, const char *buf)
 	update_full();
 }
 
-static void print_re_error(int errcode, regex_t *re, const char *re_src) {
-	size_t len;
-	char *buf;
+static void print_re_error(int errcode, regex_t *re, const char *re_src)
+{
+	size_t len = regerror(errcode, re, NULL, 0);
+	char *buf = xmalloc(len+1);
 
-	len = regerror(errcode, re, NULL, 0);
-	buf = xmalloc(len+1);
 	regerror(errcode, re, buf, len+1);
 	fprintf(stderr, "Failed to compile regex '%s': %s\n", re_src, buf);
-	exit(1);
 }
 
 static void get_album_path_ignore_re(void *data, char *buf, size_t size)
--- a/options.c
+++ b/options.c
@@ -1311,8 +1311,10 @@ static void set_album_path_ignore_re(void *data, const char *buf)
 	int rc;
 
 	fprintf(stderr, "HERE %d %p\n", ever_called, patternp);
-	if (ever_called)
+	if (ever_called) {
 		free(*patternp);
+		regfree(&album_path_ignore_re);
+	}
 	fprintf(stderr, "HERE freed\n");
 	*patternp = xstrdup(buf);
 	fprintf(stderr, "HERE '%s' %p\n", *patternp, *patternp);
--- a/options.c
+++ b/options.c
@@ -56,9 +56,9 @@ const char DEFAULT_ALBUM_PATH_IGNORE_RE[] =
 		/* The regexp being intended:
 		 *  "[^[:alnum:]]*([cC][dD]|[dD][iI][sS][cCkK])[^[:alnum:]]*[[:alnum:]]+[^[:alnum:]]*"
 		 * The [^...]* construct, however, does not seem to work with either POSIX or glibc.
+		 * Do a subset of this instead. "([-_]*|.)" at beginning or end also doesn't work.
 		 * */
-		"([cC][dD]|[dD][iI][sS][cCkK])[[:alnum:]]+";
-
+		"([cC][dD]|[dD][iI][sS][cCkK])([-_. \t]*|.)[[:alnum:]]+";
 
 /* initialized option variables */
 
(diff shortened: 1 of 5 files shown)


mahkoh:
I don't understand what you're doing.

mahkoh:
This is a breaking change that I very much doubt is going to happen. @flyingmutant should also comment on this.

nefthy:
You assume, that each album is located in a separate directory. That is probably not true for some users. While I would very much like this to be merged, I don't think it is such a great Idea. Mabee you can make this behaviour optional, through a config variable defaulting to current behaviour.

mahkoh:
>You assume, that each album is located in a separate directory.

I think it's the opposite. The assumption is that each album is contained in a single directory. Structures of the form Artist/Album/Cd{1,2} already break this.

nefthy:
> I think it's the opposite. The assumption is that each album is contained in a single directory. 

We mean the same thing. The intention is probably to not have dupplicate tracks directly next to each other in one album if for some reason the album happens to exist two times in the collection, which is quite annoing.

Its probably better to implement some deduplication strategy, like displaying only the file with the highest quality or something. But that is quite a bottomless pit...

> Structures of the form Artist/Album/Cd{1,2} already break this.

You're absolutely right on this.

forgottenswitch:
Make it optional, add a regexp for excluding path components.
- depends only on libc, yet currently adds two includes into `options.h`
- `strdup`-then-`free` in `album_path_disagrees` could be `static`+`realloc`
- `options_add` change needs factoring (not currently done, as it looks to be the only case)
- options names probably could be shorter

forgottenswitch:
So this should be it.

options.c
```diff
@@ -1267,6 +1284,44 @@ static void set_format(void *data, const char *buf)
 	update_full();
 }
 
+static void print_re_error(int errcode, regex_t *re, const char *re_src) {
```
flyingmutant: `{` should be on the line by itestlf

options.c
```diff
@@ -1267,6 +1284,44 @@ static void set_format(void *data, const char *buf)
...
+	size_t len;
+	char *buf;
```
flyingmutant: IMO it is better to not declare variables separately; this looks like very old C.

options.c
```diff
@@ -1267,6 +1284,44 @@ static void set_format(void *data, const char *buf)
...
+
+	len = regerror(errcode, re, NULL, 0);
+	buf = xmalloc(len+1);
+	regerror(errcode, re, buf, len+1);
+	fprintf(stderr, "Failed to compile regex '%s': %s\n", re_src, buf);
+	exit(1);
```
flyingmutant: You already haave `exit` after call to `print_re_error`.

options.c
```diff
@@ -1442,6 +1499,18 @@ void options_add(void)
...
+	{
+		struct cmus_opt *opt;
+		opt = option_find("album_path_ignore_re");
+		opt->data = xmalloc(sizeof(void*));
+		set_album_path_ignore_re(opt->data,
+				/* The regexp being intended:
+				 *  "[^[:alnum:]]*([cC][dD]|[dD][iI][sS][cCkK])[^[:alnum:]]*[[:alnum:]]+[^[:alnum:]]*"
+				 * The [^...]* construct, however, does not seem to work with either POSIX or glibc.
+				 * */
+				"([cC][dD]|[dD][iI][sS][cCkK])[[:alnum:]]+");
```
flyingmutant: Better to move this to a toplevel named constant.

options.h
```diff
@@ -21,6 +21,9 @@
 
 #include "list.h"
 
+#include <sys/types.h>
+#include <regex.h>
```
flyingmutant: Isn't `<regex.h>` enough?

tree.c
```diff
@@ -656,9 +728,25 @@ static inline const char *album_sort_collkey(const struct album *a)
...
+{
+	int a_dup = a->dup_id;
+	int b_dup = b->dup_id;
+	if (a_dup <= 0 || b_dup <= 0)
+		return 0;
+	if (a_dup < b_dup)
+		return -1;
+	else if (a_dup > b_dup)
+		return +1;
+	return 0;
```
flyingmutant: `return a_dup - b_dup`?

tree.c
```diff
@@ -568,6 +572,74 @@ static void album_free(struct album *album)
...
+static int special_assign_album_filename(void *data, struct track_info *ti) {
+	char **album_filename = (char**) data;
+	*album_filename = ti->filename;
+	return 1;
+}
+
+static int album_path_disagrees(struct album *album, const char *filename) {
+	char *album_filename;
+	char *album_filename_1, *filename_1;
+	char *album_filename_2, *filename_2;
```
flyingmutant: Bad names.

tree.c
```diff
@@ -950,11 +1049,18 @@ void tree_add_track(struct tree_track *track)
...
-		if (album)
-			album_free(new_album);
-	} else
+		album = find_album(new_album, ti->filename, &album_dup_count);
+		new_album->dup_id = album_dup_count + 1;
+		if (album) {
+			if (album_dup_count < 0)
+				album_free(new_album);
+			else
+				album = NULL;
```
flyingmutant: Honestly, all this dup logic is hard to follow.

PR #558, opened 2016-10-18
added option to show or hide leading zero in minutes

teearremm:
# 553

With this change users can now use the option `time_show_leading_zero=true` to always show two digits for the minutes (this is the default):
leading_zero
And `time_show_leading_zero=false` to only show 2 digits padded with a space if the duration is more than 10 minutes:
no_leading_zero

### DIFF
--- a/format_print.c
+++ b/format_print.c
@@ -142,8 +142,6 @@ static void print_time(int t)
 	stack[p++] = m % 10 + '0';
 	if (m / 10 || h || time_show_leading_zero)
 		stack[p++] = m / 10 + '0';
-	else
-		stack[p++] = ' ';
 	if (h) {
 		stack[p++] = ':';
 		do {
--- a/format_print.c
+++ b/format_print.c
@@ -140,7 +140,10 @@ static void print_time(int t)
 	stack[p++] = s / 10 + '0';
 	stack[p++] = ':';
 	stack[p++] = m % 10 + '0';
-	stack[p++] = m / 10 + '0';
+	if (m / 10 || h || time_show_leading_zero)
+		stack[p++] = m / 10 + '0';
+	else
+		stack[p++] = ' ';
 	if (h) {
 		stack[p++] = ':';
 		do {
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1096,7 +1096,7 @@ status_display_program () [command]
 	`/usr/share/doc/cmus/examples/cmus-status-display`.
 
 time_show_leading_zero (true)
-	Pad shorter durations with a leading 0 instead of a leading space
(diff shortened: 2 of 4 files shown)


flyingmutant:
Thanks a lot for the PR. However, I think a better way to go would be to add an option to the relevant formatting flag, similar to printf.

flyingmutant:
@mahkoh what do you think?

mahkoh:
Something like this: `%{duration:_}` where `_` stands for space.

mahkoh:
@teearremm: To implement this you should look at the format_parse function. Currently it only looks for `}` inside `{`. Instead you should also look for `:`, parse the flags that come afterwards, and pass the flags to the format functions.

teearremm:
@mahkoh I can implement it that way, my concern is that that will make it more difficult to switch between the display modes (the user will have to replace all of their `%d`s and `%{duration}`s with `%{duration:_}` in all of the format strings if they want to change modes).

I imagine that users might want to switch back and forth between modes, but probably want all of the times to be displayed the same way (I don't see, for example, why someone would want leading zeros to be displayed in the playlist, but not in the status bar or library view). Having it as a configuration option allows people to easily switch and updates all the times everywhere.

Up to you, though. If you feel strongly that it should be a format string I'll implement it that way.

mahkoh:
I've no strong feelings on the issue. I'll leave it to @flyingmutant.

mahkoh:
This is certainly backwards compatible. An option, which sets the default when no flags are given, can always be added later.

flyingmutant:
@teearremm, please add the docs for the new flag. Otherwise, LGTM.

flyingmutant:
@teearremm, do you have any updates on this one?

teearremm:
Sorry, it's been a busy week. I'll update the documentation shortly

format_print.c
```diff
@@ -140,7 +140,10 @@ static void print_time(int t)
 	stack[p++] = s / 10 + '0';
 	stack[p++] = ':';
 	stack[p++] = m % 10 + '0';
-	stack[p++] = m / 10 + '0';
+	if (m / 10 || h || time_show_leading_zero)
+		stack[p++] = m / 10 + '0';
+	else
+		stack[p++] = ' ';
```
teearremm: Good catch, I don't use `show_remaining_time` and didn't think to test the combination of the two flags.

Come to think of it, does it make sense to pad the time with a space at all? I was originally thinking that users would want the durations to always take up the same amount of space, but that already won't be the case for durations of more than 1 hour, and anyway if a constant width is required it can be set with a format string.

mahkoh: I don't care.

teearremm: Ok, I'll update the pull request to have no padding instead of padding with a space

Issue #567, opened 2016-10-28
CUE timing issue


Wang-Yue:
flac with the new cue plugin may start dozens of ms before the index. 

My testing file is Glenn Gould's Goldberg Variations (https://www.amazon.com/Glenn-Gould-Complete-Goldberg-Variations/dp/B00006FI7C)

Disc 2. track 30, 31 show the behavior. 

My other audio tools (VOX player, XLD converter) show the correct behavior with same cue and flac. 

I found nothing strange with the cue
```
  TRACK 30 AUDIO
    TITLE "Bach: Goldberg Variations - Variation 29"
    PERFORMER "Glenn Gould"
    INDEX 01 44:57:45
  TRACK 31 AUDIO
    TITLE "Bach: Goldberg Variations - Variation 30"
    PERFORMER "Glenn Gould"
    INDEX 01 45:59:37
```

cmus version and plugin info:

```
23:29:33 ~ $ cmus --plu
Input Plugins: /usr/local/lib/cmus/ip
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  wav:
    Priority: 50
    File Types: wav
    MIME Types:

Output Plugins: /usr/local/lib/cmus/op
  coreaudio
23:29:35 ~ $ cmus --ver
cmus v2.7.0-267-g6fcb53a
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2013 Various Authors
```

Wang-Yue:
OK after some debugging I can isolate the problem. 
I put a break point at flac_seek, and the offset passed in is 2697 for track 30. However it should be 44*60+57+45/75 = 2697.6. 
If I manually set it to be 2697.6 the track will start at the right time.

So cue plugin is sending the wrong value.

PR #568, opened 2016-10-28
use floating point number for cue track timing.

Wang-Yue:


Wang-Yue:
should solve #567

Wang-Yue:
The alternative is to keep using integer in cue_track, and do the 75 division in ip/cue.c.

PR #572, opened 2016-10-28
Fix flac plugin for small files

### LINKED ISSUE #440: FLAC plugin does not work for small files
Using version 2.7 from GitHub, but this also happened on version 2.5 from Debian stable repository.

I have two tracks from the same album that are being sorted under <No Name> in Cmus.  The album was ripped by me and the tags were edited by me with the same program, so they are identical.

The only think I can think of that is different, is the tracks are less than 10 seconds long.  1 is 10 seconds long and the other is 9.  Is Cmus having an issue with tracks this small?  BTW, they are FLAC files.

mahkoh:
The flac library does not output decoded data until it has processed a
certain amount of input data or until the end of the stream is reached.
The old version of the code did not take this into account. In
particular, us having inputted all of our data into the flac plugin does
not mean that the flac plugin has outputted all of its data yet.

Hence we remove the `priv->pos == priv-len` check and instead ask the
flac library directly if it's done.
---

Fixes #440.

### DIFF
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -357,11 +357,9 @@ static int flac_open(struct input_plugin_data *ip_data)
 	}
 
 	ip_data->sf = 0;
-	while (priv->buf_wpos == 0 && priv->pos < priv->len) {
-		if (!F(process_single)(priv->dec)) {
-			free_priv(ip_data);
-			return -IP_ERROR_ERRNO;
-		}
+	if (!F(process_until_end_of_metadata)(priv->dec)) {
+		free_priv(ip_data);
+		return -IP_ERROR_ERRNO;
 	}
 
 	if (!ip_data->sf) {
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -397,9 +397,11 @@ static int flac_read(struct input_plugin_data *ip_data, char *buffer, int count)
 		BUG_ON(avail < 0);
 		if (avail > 0)
 			break;
-		if (priv->pos == priv->len)
+		FLAC__bool internal_error = !F(process_single)(priv->dec);
+		FLAC__StreamDecoderState state = F(get_state)(priv->dec);
+		if (state == E(END_OF_STREAM))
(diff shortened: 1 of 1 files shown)


mahkoh:
Please hold off on merging this. While I think this works in almost all cases, after reading the libflac source code I think we'll have to check a few more cases to make sure that this is 100% robust.

mahkoh:
I think this should be ready now but it would be good if someone with a larger FLAC library could run a few casual tests.

jayay:
I'm testing this. I got a segfault when typing `:clear` and `:add ~/Music/` using the old cache file.

```
(gdb) bt                                     │
#0  __strcmp_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:32
#1  0x000000000040d8cc in lookup_cache_entry (filename=<optimized out>, hash=<optimized out>) at cache.c:175
#2  0x00000000004244e4 in add_file (filename=<optimized out>, force=<optimized out>) at job.c:171
#3  0x000000000042442e in add_dir (dirname=<optimized out>, root=<optimized out>) at job.c:333
#4  0x0000000000424422 in add_dir (dirname=<optimized out>, root=<optimized out>) at job.c:331
#5  0x00000000004238d6 in do_add_job (data=<optimized out>) at job.c:392
#6  0x000000000043ecbf in worker_loop (arg=<optimized out>) at worker.c:86
#7  0x00007ffff7bc45ca in start_thread (arg=0x7fffe9021700) at pthread_create.c:333
#8  0x00007ffff6f8bf6d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

Deleting the cache file and generating a new one and repeating `:clear` and `:add ~/Music/` works.

jayay:
Sorry, that must have been something different. I've changed to master, rebuilt it and deleted my cache file. Then I switched back to your new branch and tried to reload my library again. Works.

Issue #337, opened 2015-09-24
No output on terminal (PulseAudio 7.0)


vtj:
After launching cmus from the terminal, top/ps shows cmus running, but there is no output nor error messages on the terminal. Happens with both X server + rxvt, and Linux console. I tried compiling with DEBUG=2, which makes cmus write a few lines to ~/cmus-debug.txt, contents at http://pastebin.com/QdAS0Sug. I'm using v2.7.0-17-g2bfc462, tried both building directly from source, and using the binaries from Arch Linux repos. I've no custom config files under $HOME.

flyingmutant:
Here is the pastebin content, for the reference:

```
main: charset = 'UTF-8'
player_init: using real-time scheduling
player_init: setting priority to 99
player_set_op: selecting first initialized op
init_plugin: initialized mixer for pulse
init_plugin: initialized pcm for pulse
_pa_context_running_cb: pulse: context state has changed to PA_CONTEXT_CONNECTING
```

Can you please make sure that PulseAudio is configured properly and running?

mxci:
I encountered the same behavior as described above. Downgrading to libpulse-6.0 solves the problem. I don't use pulseaudio, libpulse was installed as a mpv dependency.

inca2000:
I have encountered the same problem on a fresh install. No output. Also running arch. Is this an arch problem or a bug that should be fixed?

inca2000:
Also fixed by downgrading to libpulse-6.0.

flyingmutant:
Hm, [pulse.c](https://github.com/cmus/cmus/blob/master/pulse.c) has not seen any significant changes since 2011 at least, and [PulseAudio 7.0 Release Notes](https://wiki.freedesktop.org/www/Software/PulseAudio/Notes/7.0/) says:

> #### Notes for Application Developers
> (nothing important happened)

It seems that either they are wrong, or Arch package for libpulse has changes something.

Can somebody on Arch please try to get a backtrace after the hang, using either lldb or gdb?

mahkoh:
Works for me with libpulse=7.0-2.

Possibly related: I just restarted cmus after installing pa and got an error that was something like "libncurses5.0.so not found". So I recompiled cmus and the problem went away.

mxci:
@mahkoh 
Are you on x86_64 ?

mahkoh:
Yes.

flyingmutant:
@mxci does Ctrl-C during the hang, followed by `backtrace` work?

[...]

PR #578, opened 2016-10-29
Reduce pulse plugin priority

mahkoh:
The pulse plugin might be much better than the alsa plugin but it doesn't matter if it simply gives the user a blank screen. With this change the alsa plugin is the new default on an alsa/pulse machine.

### DIFF
--- a/op/pulse.c
+++ b/op/pulse.c
@@ -270,7 +270,7 @@ static int _pa_create_context(void)
 
 	pa_context_set_state_callback(pa_ctx, _pa_context_running_cb, NULL);
 
-	rc = pa_context_connect(pa_ctx, NULL, PA_CONTEXT_NOFAIL, NULL);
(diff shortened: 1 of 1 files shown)


flyingmutant:
What do you mean by "gives the user a blank screen"?

mahkoh:
https://bugs.archlinux.org/task/46447, #337

flyingmutant:
It is unclear if this a problem with PulseAudio, Arch's PulseAudio config or cmus' output plugin, right?

I'd rather find a problem and fix it; going back to ALSA is a very crude workaround that might introduce more problems than it solves (e.g. are you sure that ALSA plugin/configuration works for the majority of users, and not only Arch?). PulseAudio plugin was born because ALSA one was not working great.

mahkoh:
I've updated the PR. I don't know why this flag was there in the first place.

flyingmutant:
Good catch, thanks!

PR #592, opened 2016-12-03
Implement BASS input plugin

nsoudee:
BASS is a library which provides highly accurate reproduction
of MOD, S3M, MTM, XM, IT, UMX playback. It plays more modules
correctly than both modplug and mikmod (subtle nuances in
effects, etc). This library is capable of more, but my main
goal was to get modules playing.

The library can be found here: http://www.un4seen.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ flac-objs		:= ip/flac.lo
 mad-objs		:= ip/mad.lo ip/nomad.lo
 mikmod-objs		:= ip/mikmod.lo
 modplug-objs		:= ip/modplug.lo
+bass-objs		:= ip/bass.lo
 mpc-objs		:= ip/mpc.lo
 vorbis-objs		:= ip/vorbis.lo
 opus-objs		:= ip/opus.lo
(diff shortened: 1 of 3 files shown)


mahkoh:
Looks fine to me modulo some style issues. But I won't test this because the library is proprietary. I'll merge this if someone else is willing to test it.

flyingmutant:
Any testers for this PR?

nefthy:
I don't think it is a good idea to include a plugin upstream that none of the maintainers can use. A separate repository with a Makefile to build against cmus is a better alternative.

mahkoh:
It's off by default so it won't cause any harm for now. Maybe someone will even test it.

cybersnow02:
libBASS not only support "xm", "it", "s3m", "mod", "mtm", "umx", 

But also, MP3, WAV, FLAC, OGG, OPUS.

It seems BASS is a nice library for fast seek and use 30% less CPU with best quality output. CMus's Dev team should do some tests with it. On their website, there is an already compiled version of 'libbass.so' for Linux x64. Can you guys link common audio files with BASS so we can start using it? I will try to do some test on my side too...

PR #595, opened 2016-12-22
Maint 2.6 win set sel

puellavulnerata:
Add a win-set-sel command to set the active track to the selected one, plus documentation and a keybinding.  This is useful for flow between library/playlist/queue: i.e., queue several tracks, then use win-set-sel to pick the point in the library to resume (and continue playing in order) from after the queue is exhausted.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -246,6 +246,7 @@ tab            win-next
 ^Y             win-scroll-up
 D, delete      win-remove
 i              win-sel-cur
+\@              win-set-sel
 space          win-toggle
 g, home        win-top
 k, up          win-up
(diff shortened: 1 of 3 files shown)


mahkoh:
It took me a while to understand what this actually does. If I understand you correctly, you want to add tracks to the queue and then immediately set the track in the library that will be played after the queue is exhausted. I've never thought about this but it seems like a reasonable feature. Documentation is a bit unclear but if you get to use it in the real program it should be easy enough to figure out.

In any case PRs must be made against master.

mahkoh:
Closing for the reasons mentioned above. Feel free to reopen with the issues addressed.

Issue #515, opened 2016-09-11
Make search rebindable


elvis-sik:
I have not found a way to do this. I want to use other keys, and not the default `/`, to search. Is there a way to do this currently or to implement it easily?

elvis-sik:
Thanks!

PR #596, opened 2016-12-29
made (backward) search a bindable command

jolange:
Implements #515.

Lets you bind the search commands, which are hard coded to `/` and `?` up to now.

The search mode in the command line is still indicated by `/`/`?`. This could be changed to e.g. `search: `/`backward search: `. Thoughts?

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -200,6 +200,8 @@ x                           player-play
 z                           player-prev
 v                           player-stop
 ^L                          refresh
+/                           search-start
+?                           search-b-start
 n                           search-next
 N                           search-prev
 .                           seek +1m
--- a/command_mode.c
+++ b/command_mode.c
@@ -1400,6 +1400,16 @@ static void cmd_search_prev(char *arg)
 	}
 }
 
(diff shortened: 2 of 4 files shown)


mahkoh:
LGTM. I'll test this over the weekend.

jolange:
Any news, @mahkoh?

mahkoh:
Sorry for the delay. Work has been very stressful lately.

I've tested your patch now and there is an issue that needs to be fixed before it can be merged: There is no migration of existing users, search simply stops working for them because `/` and `?` are not bound.

jolange:
No problem, I'm in no hurry ;-)

You are right, that's a problem. One option would be to re-add the old hard coded bindings, which provide a fallback solution for existing installations/configurations. Users can create custom search bindings on top of that and rebind `/`/`?`. The only catch is that you cannot completely unbind `/`/`?`, but that has also been the case before.

What do you think?

mahkoh:
Sounds fine.

jolange:
Ok, done.

flyingmutant:
@mahkoh good to merge?

PR #597, opened 2016-12-29
remove trailing spaces from key in cmd_unbind() and cmd_showbind()

jolange:
Resolving minor FIXMEs.

### DIFF
--- a/command_mode.c
+++ b/command_mode.c
@@ -706,7 +706,12 @@ static void cmd_unbind(char *arg)
 	if (*key == 0)
 		goto err;
 
-	/* FIXME: remove spaces at end */
+	/* remove trailing spaces from key */
+	char *key_end = key + strlen(key);
+	while (isspace(*(key_end-1))) {
+		key_end--;
+	}
+	*key_end = 0;
 
 	key_unbind(arg, key, flag == 'f');
 	return;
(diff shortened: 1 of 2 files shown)


command_mode.c
```diff
@@ -727,7 +732,12 @@ static void cmd_showbind(char *arg)
 	if (*key == 0)
 		goto err;
 
-	/* FIXME: remove spaces at end */
+	/* remove trailing spaces from key */
+	char *key_end = key + strlen(key);
+	while (isspace(*(key_end-1))) {
+		key_end--;
+	}
+	*key_end = 0;
```
mahkoh: 1. Please no duplicate code.

2. `isspace` is broken. Write a small utility that checks for ' ' and '\t'.

jolange: Both addressed (will squash things later to clean up, if everyone is satisfied).

Just for my own education: what is wrong with `isspace()`?

mahkoh: It accepts only a single byte which is not enough to determine if something is a space. E.g. "a ̈b" contains ' '.

flyingmutant:
@mahkoh, do you mean by "broken isspace" that it does not check for various Unicode (outside ASCII) spaces?

mahkoh:
There are at least three issues:

1. isspace depends on the locale. If a particular locale defines isspace = true for all inputs, then this patch leads to segfaults.

2. I do not know of an official algorithm to determine what a space in unicode is. There are various code points which are referred to as spaces, but those can be combined with other characters so that no space actually appears on the display. Hence the question if something is a space is not well-defined for any unit smaller than grapheme clusters. As I demonstrated above, the ascii space can be combined with  ̊ to form a freestanding  ̊ (usually used to decorate other letters such as å), so this issue is not restricted to code points above 127.

3. isspace is stateless and only accepts single bytes so it cannot be used to detect spaces above code points 127.

PR #601, opened 2017-01-06
Add next/prev playback controls to list in step 3

rydgeback:
Also made a small change to sentence syntax

### DIFF
--- a/Doc/cmus-tutorial.txt
+++ b/Doc/cmus-tutorial.txt
@@ -112,11 +112,12 @@ this:
 @endpre
 
 Use the up and down arrow keys to select a track you'd like to hear, and press
-Enter to play it. Here's some keys to control play:
+Enter to play it. Here are some keys to control playback:
 
 Press *c* to pause/unpause
 Press right/left to seek by 10 seconds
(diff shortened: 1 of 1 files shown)


Doc/cmus-tutorial.txt
```diff
@@ -112,11 +112,12 @@ this:
 @endpre
 
 Use the up and down arrow keys to select a track you'd like to hear, and press
-Enter to play it. Here's some keys to control play:
+Enter to play it. Here are some keys to control playback:
 
 Press *c* to pause/unpause
 Press right/left to seek by 10 seconds
 Press *<*/*>* seek by one minute
+Press *z* to play the previous track and *b* to play the next track (if in a queue)
```
flyingmutant: I think `(if in a queue)` bit is not needed -- it can cause confusion with cmus' queue view.

rydgeback: You're right, especially considering you can still go to the previous or next track without having a queue. At the time I wasn't aware of that, the change has been made.

mahkoh:
Thanks.

PR #610, opened 2017-01-18
Added option to select starting view

AlexKolpa:
Allow overriding of the starting view when cmus opens. Default is still TREE_VIEW.


mahkoh:
You need to add documentation for the new option.

AlexKolpa:
@mahkoh Added the documentation, and fixed an issue where a non-existing view could be selected

PR #614, opened 2017-01-21
Use correct format for playing 24-bit files with op/oss

t6:
Adding support for AFMT_{S24,U24}_{LE,BE} will make the format
selection even more difficult to read, so while here replace it with a
simple lookup table.

Tested on FreeBSD.  The symptoms of using the wrong format were
the same as with the sndio plugin (#613).

### DIFF
--- a/op/oss.c
+++ b/op/oss.c
@@ -20,6 +20,7 @@
 #include "sf.h"
 #include "xmalloc.h"
 #include "debug.h"
+#include "utils.h"
 
 #if defined(__OpenBSD__)
 #include <soundcard.h>
@@ -48,6 +49,7 @@ static int oss_reset(void)
 	return 0;
 }
 
+#if defined(__linux__)
 /* defined only in OSSv4, but seem to work in OSSv3 (Linux) */
 #ifndef AFMT_S32_LE
 #define AFMT_S32_LE	0x00001000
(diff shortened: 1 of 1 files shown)


op/oss.c
```diff
@@ -55,13 +55,47 @@ static int oss_reset(void)
...
-#ifndef AFMT_S24_PACKED
-#define AFMT_S24_PACKED	0x00040000
+#ifndef AFMT_S24_LE
+#define AFMT_S24_LE	0x00010000
+#endif
+#ifndef AFMT_S24_BE
+#define AFMT_S24_BE	0x00020000
+#endif
+#ifndef AFMT_U24_LE
+#define AFMT_U24_LE	0x00040000
```
mahkoh: This seems rather dubious. Why was this constant S24_PACKED before?

t6: I was under the impression that the constants are the same across different OSS implementation, but that's apparently not true. S24_PACKED isn't even defined on FreeBSD and U24_LE isn't defined in OSSv4.

I don't really understand why the constants need to be added here at all to be honest.  I've amended the patch a little.

op/oss.c
```diff
@@ -76,35 +105,17 @@ static int oss_set_sf(sample_format_t sf)
...
-				tmp = AFMT_U16_LE;
-			}
-		}
-	} else if (sf_get_bits(oss_sf) == 8) {
-		if (sf_get_signed(oss_sf)) {
-			tmp = AFMT_S8;
-		} else {
-			tmp = AFMT_U8;
+	found = 0;
+	for (i = 0; i < sizeof(oss_fmts)/sizeof(oss_fmts[0]); i++) {
```
mahkoh: There is a macro for this in utils.h

mahkoh:
This seems to remove support for OSSv3 which requires some justification.

t6:
That wasn't my intent. On what system do you see problems?

mahkoh:
```
-/* defined only in OSSv4, but seem to work in OSSv3 (Linux) */
```

t6:
I've re-added the constant-definining-block under an `#ifdef __linux__`, but I still don't understand where exactly the magic numbers come from.

mahkoh:
I'll merge this for now. If anyone still uses OSS, we'll deal with breakage when it happens.

PR #620, opened 2017-01-23
Canonicalize path when adding files to collection

### LINKED ISSUE #423: Duplicate files while scanning library
While scanning my library **in a symlinked folder** cmus constantly finds duplicate files, like this:

Sometimes there are only two of them, other times three or even four (like in this example).

I probably do have some duplicate files, but it finds duplicates for practically every file, which is impossible — I have searched for many of supposedly duplicate files and found only one of each. I am unsure what info I should provide to clarify the issue :(, so please ask questions if you care to help.

```
Darwin Kernel Version 15.3.0, cmus v2.7.1
```

On the other note, how am I supposed to deal with duplicates anyway? I read the man, but couldn't find any way to display, e.g. the file path of the selected file (in pane 1). Duplicates make listening in album mode difficult. Have you considered providing the user with some option to deal with them, e.g. delete the duplicates, or add an additional level of grouping (e.g file path) that would remove the duplicates?

nefthy:
When there are symlinks involved, a file can be rechable via multiple filenames. In that case a file can be added multiple times to the library.

This PR uses realpath to resolve symlinks before adding files to the collection, mitigating this issue.

fixes #423

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1040,6 +1040,11 @@ replaygain_limit (true)
 replaygain_preamp (0.0)
 	Replay gain preamplification in decibels.
 
+resolve_symlinks (true)
+	Resolve any symlinks filenames before adding them to the library. This
+	prevents duplicates resulting from files reachable by multiple paths.
+	The path displayed by %f format is the real path of the added file.
+
 resume (false)
 	Resume playback on startup.
 
--- a/job.c
+++ b/job.c
@@ -42,6 +42,8 @@
 #include <errno.h>
 #include <sys/mman.h>
 
+extern int resolve_symlinks;
+
 enum job_result_var {
 	JOB_RES_ADD,
 	JOB_RES_UPDATE,
@@ -166,6 +168,11 @@ static int add_file_cue(const char *filename);
 static void add_file(const char *filename, int force)
 {
 	struct track_info *ti;
+	char buf[PATH_MAX];
+
+	if (resolve_symlinks) {
+		filename = realpath(filename, buf);
+	}
 
 	if (!is_cue_url(filename)) {
 		if (force || lookup_cache_entry(filename, hash_str(filename)) == NULL) {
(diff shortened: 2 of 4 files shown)


mahkoh:
I'm not so sure about this. `~/music` is symlinked to `/home/share/music` on my system and I don't really want cmus to know about this.

nefthy:
There are not that many places where the filename appears in the interface. The issues described in #423 seem reasonable, so it is a matter of judgment. Why don't you want cmus to know the real location of the files? What is the use case?

mahkoh:
>There are not that many places where the filename appears in the interface. 

The path appears whenever the `%f` formatter is used.

As I said before it would be good to know how other players deal with this.

nefthy:
Amarok resolves
clementine resolves
exaile is configurable

configurable more the spirit of cmus. I propose we add resolve_symlinks to the config defaulting to true.

nefthy:
There are other solutions I can think of, but they are all way more complicated. What is your opinion on this?

mahkoh:
I don't like it and I especially don't like it being optional. The only thing that I might consider acceptable is storing the real path additionally to the normal path. But then you have to consider the additional storage.

I will probably not accept anything that replaces normal paths by real paths simply because I don't know if I know all the consequences of such a change.

nefthy:
ok, since it works for me as is, someone else will have to do that, if there is interest.

xeverse:
@nefthy how can one use this solution  
& still be up-to-date with the up-stream?

nefthy:
@xeverse its posible to maintain local patches on top of a remote repository with git, but if you don't already know how to do that, this topic is probably quite advanced. I would rather consider some other form of de-duplication if I where you.

xeverse:
@nefthy. no way, please    
share how you do it.   
i'd like to learn..

nefthy:
I don't use this patch, as I am not affected.On Feb 3, 2019 11:09 AM, xeverse <notifications@github.com> wrote:@nefthy. no way, please
share how you do it.
i'd like to learn..

PR #621, opened 2017-01-23
set curses escdelay

### LINKED ISSUE #618: Abort command-line mode on ESC
When pressing ESC after having entered something in command line mode in vim, the command line is cleared and the mode is exited.

I've missed that quite often in cmus and would like to implement it. I can't think of any backwards compatibility issues here.

nefthy:
unless the ESCDELAY environment variable is set.

fixes #618

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -147,6 +147,8 @@ static int show_cursor;
 static int cursor_x;
 static int cursor_y;
 
+static const int default_esc_delay = 25;
+
 static char *title_buf = NULL;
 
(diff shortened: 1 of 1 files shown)


mahkoh:
Why not let the user configure it via the environment variable?

nefthy:
The user can configure it via the environment variable. The patch just set it to 25 if the environment variable is not set. 25ms is what vim also uses so that seems a good default.

Also ESCDELAY seems to be undocumented.

Issue #626, opened 2017-01-27
No test for a working C11 compiler


dwatteau:
This is with cmus v2.8.0-rc0 on OpenBSD -current, trying to build with GCC 4.2.1, which is still the default compiler at the moment (it's possible to build ports with a newer compiler when needed, however).

cmus now uses `<stdatomic.h>`, which requires a C11 compiler, AFAIK.

However, `configure` doesn't reject non-C11 compilers, such as GCC 4.2.1.

It justs removes a lot of flags (since `-std=gnu11` and the warnings flags are tested at the same time), and the `ffmpeg.c` build test fails, since it also requires a working `-std=gnu11`:

```
checking for CFLAGS -std=gnu11 -pipe -Wall -Wshadow -Wcast-align -Wpointer-arith -Wwrite-strings -Wundef -Wmissing-prototypes -Wredundant-decls -Wextra -Wno-sign-compare -Wformat-security... no

[...]

checking for successful build of ffmpeg.c... no
```

The build then fails this way:
```
cc -c -O2 -pipe [some flags, but no -std=XXX] -o ape.o ape.c
In file included from ape.c:24:
utils.h: In function 'strscpy':
utils.h:322: error: 'for' loop initial declaration used outside C99 mode
```

Of course, if you force `-std=gnu99`, it's not going to be enough:
```
cc -c -O2 -pipe -std=gnu99 [some flags] -o buffer.o buffer.c
In file included from buffer.c:21:
locking.h:23:23: warning: stdatomic.h: No such file or directory
In file included from buffer.c:21:
locking.h:26: error: expected ':', ',', ';', '}' or '__attribute__' before 'tail'
```

So there's **no problem** requiring a C11 compiler here, but I think `configure` should just test for a working C11 compiler, and then abort at the earliest possible stage if it's not found.

mahkoh:
PR's fixing this and #625 would be accepted.

>GCC 4.2.1, which is still the default compiler at the moment

For what purpose?

dwatteau:
> PRs fixing this and #625 would be accepted.

See  #629 for this. Hope it's good.

> For what purpose?

On OpenBSD, GCC 4.2.1 is still the *base* system compiler, at the moment. The reason for this is that the project doesn't want GPLv3 code in its base system, so newer GCC releases are excluded. There's some ongoing work to switch to clang, but GCC 4.2.1 will probably remain the base compiler for a long time for the platforms where clang is not usable.

[...]

Issue #625, opened 2017-01-27
-I${topdir} included too late in the compiler flags


dwatteau:
This is with cmus v2.8.0-rc0 on OpenBSD -current, building with clang-3.9.1.

cmus currently fails to build if the id3lib library (http://id3lib.sourceforge.net) is installed on the system:
```
ip/mad.c:125:16: error: variable has incomplete type 'struct id3tag'
        struct id3tag id3;
                      ^
```

That's because we are trying to `#include` our own `id3.h` header, but `/usr/local/include/id3.h`, from the external id3lib library, is actually included first, if it's installed when you're building cmus.

The compiler is invoked this way:
```
clang -c -O2 -pipe -Wno-system-headers -I/usr/local/include -DHAVE_DISCID -DHAVE_CONFIG -std=gnu11 -pipe [various -W flags] -MMD -MP -MF .dep-ip-mad.lo -DREALTIME_SCHEDULING -D_FILE_OFFSET_BITS=64 -I"/home/ports/cmus-2.8.0-rc0" -I/usr/local/include -fPIC -o ip/mad.lo ip/mad.c
```
`-I/usr/local/include` is set first, so it takes precedence over the following `-I` flags. It is added by `checks.sh`:128, on BSD systems.

Injecting `-I"${topdir}"` at the earliest place in the compiler flags (I'd say even before `$CFLAGS`) _might_ be a way of making sure that our own header files take precedence.

flyingmutant:
Fixed in #629.

PR #629, opened 2017-01-31
Fixes for issues #625 and #626

dwatteau:
Here's my attempt at fixing the issues #625 and #626 I've previously reported.

### DIFF
--- a/configure
+++ b/configure
@@ -2,9 +2,47 @@
 
 . scripts/configure.sh || exit 1
 
+c11_code="
+#include <stdatomic.h>
+
+int main(void)
+{
+#ifdef __STDC_NO_ATOMICS__
+#error No C11 atomics
+#endif
+	_Atomic int res = ATOMIC_VAR_INIT(0);
+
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -2,9 +2,47 @@
...
+
+	return res;
+}
+"
+
+check_c11()
+{
+	msg_checking "for C11 (with atomics support)"
+
+	for flag in -std=gnu11 -std=gnu1x -std=c11 ""
```
mahkoh: gnu11 should be sufficient

dwatteau: I was thinking of potential compilers where gnu11 isn't available, but other C11 options (or none at all) would be enough to build cmus.

I don't think these fallbacks hurt, unless we really only want gnu11 compatibility (I could be fine with that).

mahkoh: gnu1x is certainly unnecessary.

dwatteau: Here you go.

mahkoh:
Merged manually.

PR #633, opened 2017-02-05
Fix segfault in format_skip_cond_expr

### LINKED ISSUE #632: cmus crash due to malformed format string
1) `uname -a`:
Linux kern 4.9.4-kern #3 SMP Thu Jan 19 22:34:45 CET 2017 x86_64 Intel(R) Core(TM) i5-4590 CPU @ 3.30GHz GenuineIntel GNU/Linux
(vanilla sources kernel)

2) `cmus --version`:
cmus 2.8.0-rc0

By accident I created a format string that made cmus crash. I tried to remove single parts of it to identify the culprit, but I could only reproduce it with the full string. In the end I found out I was missing a "}" right before the first bullet character to close the %{?A? statement. I attached the backtrace as a separate file for better readbility. The malformed format string is right at the top of it.

The desired output should print a lot of information, but no empty placeholders if a field is not set, for example empty braces. Like so:
> Various Artists [Oxford Camerata] · Early Music Greatest Hits 1100-1600 · 16. Missa Papae Marcelli - Kyrie (Palestrina, Giovanni da)

[cmus-gdb.txt](https://github.com/cmus/cmus/files/752962/cmus-gdb.txt)

mahkoh:
Fixes #632 

@flyingmutant

### DIFF
--- a/format_print.c
+++ b/format_print.c
@@ -379,7 +379,7 @@ static struct expr *format_parse_cond(const char* format, int size)
 static uchar format_skip_cond_expr(const char *format, int *s)
 {
 	uchar r = 0;
-	while (*format) {
+	while (format[*s]) {
 		uchar u = u_get_char(format, s);
 		if (u == '}' || u == '?') {
 			return u;


format_print.c
```diff
@@ -379,7 +379,7 @@ static struct expr *format_parse_cond(const char* format, int size)
 static uchar format_skip_cond_expr(const char *format, int *s)
 {
 	uchar r = 0;
```
mahkoh: What the hell is this variable anyway. It's not even used anywhere. This code seems seriously broken.

jolange: It was obviously used when it was introduced in fe4efbb, but was abandoned cd2e7d7.

Issue #449, opened 2016-05-21
Album duration


iliavolyova:
Hi,

is there a way to get the cumulative running time of all the tracks on an album? It would be nice if it could be stored in a variable like the total library duration is stored in total, so we could include it in format_treewin. 

Thanks!

pgaskin:
#639 was merged.

PR #639, opened 2017-02-13
initial attempt at album duration

zandrmartin:
Hi there! This is an initial attempt at adding an album duration format option as mentioned in issue #449. I am sure this will require more work, but I wanted to get feedback on it before I go any farther. This is what it looks like:

2017-02-13-082200_swaygrab

Please let me know what I need to do to get this into a state where it can be merged. Thank you!

### DIFF
--- a/options.c
+++ b/options.c
@@ -239,7 +239,7 @@ static const struct {
 	[FMT_PLAYLIST_VA]	= { "format_playlist_va"	, " %-21%A %3n. %t (%a)%= %y %d %{?X!=0?%3X ?    }"	},
 	[FMT_TITLE_ALT]		= { "altformat_title"		, "%f"							},
 	[FMT_TITLE]		= { "format_title"		, "%a - %l - %t (%y)"					},
-	[FMT_TRACKWIN_ALBUM]	= { "format_trackwin_album"	, " %l "						},
+	[FMT_TRACKWIN_ALBUM]	= { "format_trackwin_album"	, " %l %= %{albumduration} "				},
 	[FMT_TRACKWIN_ALT]	= { "altformat_trackwin"	, " %f%= %d "						},
 	[FMT_TRACKWIN]		= { "format_trackwin"		, "%3n. %t%= %y %d "					},
 	[FMT_TRACKWIN_VA]	= { "format_trackwin_va"	, "%3n. %t (%a)%= %y %d "				},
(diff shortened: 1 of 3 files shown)


options.c
```diff
@@ -239,7 +239,7 @@ static const struct {
 	[FMT_PLAYLIST_VA]	= { "format_playlist_va"	, " %-21%A %3n. %t (%a)%= %y %d %{?X!=0?%3X ?    }"	},
 	[FMT_TITLE_ALT]		= { "altformat_title"		, "%f"							},
 	[FMT_TITLE]		= { "format_title"		, "%a - %l - %t (%y)"					},
-	[FMT_TRACKWIN_ALBUM]	= { "format_trackwin_album"	, " %l "						},
+	[FMT_TRACKWIN_ALBUM]	= { "format_trackwin_album"	, " %l %= Total: %Z "					},
```
zandrmartin: Fixed!

ui_curses.c
```diff
@@ -302,6 +303,7 @@ enum {
 };
 
 static struct format_option track_fopts[NR_TFS + 1] = {
+	DEF_FO_TIME('Z', "albumduration", 0),
```
mahkoh: I don't think we need the one character code.

zandrmartin: Good point - to be honest I wasn't entirely clear on how format options worked when I added that char code. I've removed it and adjusted the default format string in `options.c`. I also added the key to the `Doc/cmus.txt` man page.

options.c
```diff
@@ -239,7 +239,7 @@ static const struct {
 	[FMT_PLAYLIST_VA]	= { "format_playlist_va"	, " %-21%A %3n. %t (%a)%= %y %d %{?X!=0?%3X ?    }"	},
 	[FMT_TITLE_ALT]		= { "altformat_title"		, "%f"							},
 	[FMT_TITLE]		= { "format_title"		, "%a - %l - %t (%y)"					},
-	[FMT_TRACKWIN_ALBUM]	= { "format_trackwin_album"	, " %l "						},
+	[FMT_TRACKWIN_ALBUM]	= { "format_trackwin_album"	, " %l %= %{albumduration} "				},
```
jolange: I'd like to have that variable, but do we want to show the album length here by default?

zandrmartin: I don't mind removing it if that is preferred. I had it there to test and thought it made a nice addition. I would guest most players do show album duration by default.

nefthy: @jolange what speaks against default on?

jolange: Nothing objective, just preference. I personally like the line to continue to the end, but I use my custom format anyway. The question is what most people prefer. (It's a minor point anyway, since everybody can change it)

Issue #235, opened 2015-01-10
cmus crashes when trying to clear or filter large library


iste-:
`:clear*** Error in 'cmus': double free or corruption (fasttop): 0x00007f3404025800 ***`
and then a SIGABRT signal is sent out. this happens on both cmus v2.60 and cmus v2.6.0-63-gaeb6471

flyingmutant:
Thanks a lot for the report! Can you please launch cmus under `gbd` or `lldb`, and provide a backtrace (`bt` command in debugger shell)  after the crash?

iste-:
´0x00007ffff6cca107 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: File or catalogue doesn't exist
(gdb) backtrace
#0  0x00007ffff6cca107 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6ccb4e8 in __GI_abort () at abort.c:89
#2  0x00007ffff6d08044 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7ffff6dfac60 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff6d0d81e in malloc_printerr (action=1, str=0x7ffff6dfae48 "double free or corruption (fasttop)", ptr=<optimized out>) at malloc.c:4996
#4  0x00007ffff6d0e526 in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:3840
#5  0x000000000042bf21 in artist_free (artist=artist@entry=0x7fffe0026260) at tree.c:540
#6  0x000000000042d878 in tree_remove (track=track@entry=0x7fffe00264b0) at tree.c:1214
#7  0x00000000004201ec in free_lib_track (item=0x7fffe00264b0) at lib.c:317
#8  0x0000000000414c07 in editable_remove_track (e=e@entry=0x656880 <lib_editable>, track=0x7fffe00264b0) at editable.c:103
#9  0x0000000000414e8b in editable_clear (e=0x656880 <lib_editable>) at editable.c:296
#10 0x0000000000411852 in view_clear (view=<optimized out>) at command_mode.c:74
[...]

PR #647, opened 2017-02-25
Change album_add_track to always add the track

### LINKED ISSUE #417: Unexpected behavior & segfault when adding two particular albums
There exist two very similar albums (same artist, similar album titles, similar track listing). When adding either of these albums to cmus, there is no issue. However, when adding _both_ of them, only one shows up in the library, and upon attempting to quit, cmus segfaults without writing relevant information to cmus-debug.txt.

List of files in the two albums, with slight filename differences highlighted:
list_highlighted

cmus after attempting to add both albums, showing only one in the library:
cmus

Segfault upon quitting cmus:
segfault
I have been able to reproduce this on Ubuntu (cmus v2.5.0) and Parabola (v2.7.1 and v2.7.0-67-g6704e78). An archive of the particular files which cause the problem is available [here](https://memeon.me/file/be91d45c1185d43f.tar.gz).

mahkoh:
Otherwise a track might be added to the library but not to the tree.

Fixes #417

### DIFF
--- a/tree.c
+++ b/tree.c
@@ -864,11 +864,8 @@ static void album_add_track(struct album *album, struct tree_track *track)
 		parent = *new;
 		if (result < 0)
 			new = &((*new)->rb_left);
-		else if (result > 0)
-			new = &((*new)->rb_right);
 		else
-			/* only add to tree if not there already */
-			return;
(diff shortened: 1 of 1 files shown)


iste-:
also fixes #235

Issue #640, opened 2017-02-18
Trigger the status_display_program on more event


Bew78LesellB:
It would be great if the program specified by the option `status_display_program` could be triggered on more events, and more specifically on event:
- exit

Other events like:
- next
- previous

are triggered, but from the script, there is no way to differentiate between the two, maybe add a parameter ?

jolange:
1) I also thought about implementing triggering on a (clean) exit already. Resetting to the initial state (`status stopped`) would be consistent. I guess I'll look into that soon.
2) Why do you need the "event information" apart from the new status?

Bew78LesellB:
I'm make a desktop widget, and I want it to be show the widget when cmus is running, and hide it completely when cmus is closed (but not when only stopped), will your thoughts for 1. allows me to differenciate between `stopped` and `exiting` ? 

2. I just want to make a different animation when a new track is playing (eg: slide from left when it's previous track, et slide from right when it's the next one or if the user selected a new track to play) 

---

I was thinking about adding an `event` argument passed to the script that will explain in one word/string WHY the script has been called. 

So the script arguments would be separated in 2 kinds:
- event (the WHY) 
- the state variables

I see a few possible events for now:
- start : triggered when cmus starts
- exit : triggered when cmus exits normally (or is closed by SIGHUP?) 
- previous : triggered when the user ask to play the previous track
- next : triggered when the user ask to play the next track or another track
- maybe others.. 

Thoses event does not interfere with the current script arguments, so there is no issue with compatibility I think

jolange:
> differentiate between `stopped` and `exiting`

Not with the way that I had in mind. A `status exited` or something similar might be an option, but you can of course also always just check if cmus is running.

> I was thinking about adding an event argument passed to the script that will explain in one word/string WHY the script has been called.

That might be a little harder. At the moment the status program is only called when the 1) the status 2) the file or 3) the metadata changed, and it does not know about the reasons/events.

Bew78LesellB:
> you can of course also always just check if cmus is running.

I'm not sure about that, because when the script is executed, even though it's asynchronous, cmus may be still open when the script is checking the running status of cmus.. Also I want my widget to be reactive, I don't want to check ever 1 or 2 seconds if cmus is still running! 

> > I was thinking about adding an event argument passed to the script that will explain in one word/string WHY the script has been called.
> 
[...]

PR #648, opened 2017-02-25
added "exiting" player status

jolange:
When leaving the main loop, the player status is set to "exiting" and
the status program is spawned just with the arguments "status exiting".

Partially implements #640.

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -1794,40 +1794,38 @@ static void clear_error(void)
 
 /* screen updates }}} */
 
-static void spawn_status_program(void)
+static void spawn_status_program_inner(const char *status_text, struct track_info *ti)
 {
-	enum player_status status;
-	const char *stream_title = NULL;
-	char *argv[32];
-	int i;
-
 	if (status_display_program == NULL || status_display_program[0] == 0)
 		return;
 
-	status = player_info.status;
-	if (status == PLAYER_STATUS_PLAYING && player_info.ti && is_http_url(player_info.ti->filename))
-		stream_title = get_stream_title();
+	char *argv[32];
+	int i = 0;
 
-	i = 0;
 	argv[i++] = xstrdup(status_display_program);
 
 	argv[i++] = xstrdup("status");
-	argv[i++] = xstrdup(player_status_names[status]);
-	if (player_info.ti) {
+	argv[i++] = xstrdup(status_text);
+
+	if (ti) {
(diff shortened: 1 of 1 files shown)


mahkoh:
I think I've already rejected such a PR in the past.

mahkoh:
#325

jolange:
Thanks for digging this up, I have not seen it before. I will look into the dbus/mpris API, sounds nice. By the way: this is not documented anywhere, is it?

Nevertheless, since cmus provides the interface via the status program, I think cmus' last sign of life should somehow indicate that it is terminating. Why do you think an exiting state is not a good idea?

mahkoh:
FWIW, you can simulate this by using a small wrapper

```bash
#!/bin/sh
cmus
status_program status exiting
```

mahkoh:
Something like this would probably get accepted:

```c
static void spawn_status_program_inner(const char *status_text, struct track_info ti)
{
	if (status_display_program == NULL || status_display_program[0] == 0)
		return;

	char *argv[32];
	int i = 0;

	argv[i++] = xstrdup(status_display_program);
	argv[i++] = xstrdup("status");
	argv[i++] = xstrdup(status_text);
	if (ti)
		expand_status_program_ti_args(ti, argv, &i);
	argv[i++] = NULL;

	if (spawn(argv, NULL, 0) == -1)
		error_msg("couldn't run `%s': %s", status_display_program, strerror(errno));

	for (i = 0; argv[i]; i++)
		free(argv[i]);
}
```

jolange:
@mahkoh, I guess you want to call the 'inner' function after the main loop like this?

ui_curses.c
```diff
@@ -1794,40 +1794,38 @@ static void clear_error(void)
...
 
-	i = 0;
 	argv[i++] = xstrdup(status_display_program);
 
 	argv[i++] = xstrdup("status");
-	argv[i++] = xstrdup(player_status_names[status]);
-	if (player_info.ti) {
+	argv[i++] = xstrdup(status_text);
+
+	if (ti) {
```
nefthy: please extract this to a seperate function to keep the line count per function resonable.

jolange:
Done. Let me know if you are satisfied with this, then I can rebase it.

(BTW nice to see that you try to revive the development here!)

nefthy:
LGTM, we can merge after the rebase. Thanks.

jolange:
Ok, done!

Issue #612, opened 2017-01-19
Sort by Composer


p5a0u9l:
How reasonable is it to request a feature that allows better support for music with extended tags, most importantly, Composer? One in particular would be extending `aaa_mode` to `aaac_mode` (which of course breaks the `aaa` symmetry, but you see the idea hopefully). 

If similar requests (support for classical music tags, etc.) has previously come up in forums, pardon and point me there. 

To sweeten the deal, I'd be happy to contribute said support where possible. I'm familiar with `c`, but less so with contributing to large open-source projects, but have been looking for a good opportunity to dive in somewhere.

nefthy:
> To sweeten the deal, I'd be happy to contribute said support where possible. I'm familiar with c, but less so with contributing to large open-source projects, but have been looking for a good opportunity to dive in somewhere.

cmus is quite welcoming contributions. Just clone the repository, hack away and make a pull request when your done. However keep the following in mind:

- Use and keep the formatting (linux kernel convetions) and style (c99), that is used in the source files.
- Don't break current behavior if possible. If this is not the case, discuss it with the maintainers (@flyingmutant , @mahkoh) first.

mahkoh:
Before writing any code you should explain in more detail what changes you want.

p5a0u9l:
@mahkoh agreed completely. If I can communicate the functionality I'm after sufficiently, you will be
better equipped to point me to the relevant places in the code as well as advise on the feasibility of the changes. 

The general theme is increased support for classical music tags in cmus.  Here's a simple proposed starting point, 

Add the "composer" tag to the list of Filter strings, which, from the man page, currently are limited 
to 
     filename, artist, albumartist, album, title, genre, comment, codec, codec_profile, media

this would enable creating a filter like, 
    composer="*Bach*"&artist="*Perahia"

More ambitious functionality would add Composer to view 1 in addition to Artist/Album. Basically, allow sort by Composer OR Artist OR Album. But, how's starting with the first idea, where in the code should I look? Cheers

paulrbts:
I'd definitely support/use this! I don't know C so can't help write it but if I can help test it, I'd gladly help.

jkarres:
I've just created a pull request (#666) that creates a setting enabling sorting the tree view by composer.  (If there's a way to associate it to this issue, I'm afraid I didn't see it.)  It's still got issues, but I'd appreciate it if anybody could tell me if I'm on the right path.

h4fnp:
[...]

PR #666, opened 2017-04-04
adding composer_as_artist setting

jkarres:
which uses a track's composer tag in place of artist if it exists

This allows the tree view to be organized by composer.

Is there a nicer way to do rebuild_tree?

This addresses issue #612.


flyingmutant:
This looks way too hackish. cmus' tree view is very inflexible, but making it more flexible will take a lot more refactoring; options like `composer_as_artist` are huge hacks, which make codebase worse. So I have to say "no" to this PR, sorry.

PR #667, opened 2017-04-04
Added C-n/C-p for history navigation

### LINKED ISSUE #482: FR: Control-n/p to browse command history
A more _vimish_ way to browse history will be to use `C-n`/`c-p`, is that hard to implement?

Also, question apart, is it possible to rebind keys in command mode?

tagniam:
Resolves #482. `C-p/C-n` in command mode and search mode now act like the Up/Down arrow keys, respectively.

### DIFF
--- a/command_mode.c
+++ b/command_mode.c
@@ -2909,6 +2909,12 @@ void command_mode_ch(uchar ch)
 		}
 		input_mode = NORMAL_MODE;
 		break;
+	case 0x10: // ^P
+		command_mode_key(KEY_UP);
+		return;
+	case 0xE: // ^N
+		command_mode_key(KEY_DOWN);
+		return;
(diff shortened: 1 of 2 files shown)


tagniam:
Should history navigation be separated into subroutines (putting [this](https://github.com/cmus/cmus/blob/861939d2782c814f59aec0fc428086be4cd54661/command_mode.c#L2997-L3017) and [this](https://github.com/cmus/cmus/blob/861939d2782c814f59aec0fc428086be4cd54661/search_mode.c#L262-L278) into static functions) instead of relying on `command_mode_key()` and `search_mode_key()`? or would that be overkill?

flyingmutant:
Thanks a lot! Sorry for the delayed merge.

PR #668, opened 2017-04-09
Include link to Reddit

bishisht:


flyingmutant:
I'd prefer not to. The subreddit is almost empty for now. I think wiki is the right place for this link.

PR #669, opened 2017-04-09
Expose stream info via MPRIS

### LINKED ISSUE #665: Feature request: Expose stream info via MPRIS
`cmus-remote -Q` shows stream metadata (see #23). It would be nice if this was exposed via MPRIS, too. If you consider this a reasonable feature, which naming convention would you suggest? Since there's no appropriate field in the [specs](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/), I would use something like "cmus:stream".

jolange:
... if a stream is currently playing

Resolves #665

### DIFF
--- a/mpris.c
+++ b/mpris.c
@@ -386,7 +386,7 @@ static int mpris_metadata(sd_bus *_bus, const char *_path,
 			CK(mpris_msg_append_si_dict(reply, "xesam:discNumber",
 						ti->discnumber));
 		if (is_http_url(ti->filename))
-			CK(mpris_msg_append_ss_dict(reply, "cmus:stream",
+			CK(mpris_msg_append_ss_dict(reply, "cmus:stream_title",
 						get_stream_title()));
 	}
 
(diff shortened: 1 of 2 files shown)


flyingmutant:
@mahkoh  thoughts on this one?

mpris.c
```diff
@@ -385,6 +385,9 @@ static int mpris_metadata(sd_bus *_bus, const char *_path,
 		if (ti->discnumber != -1)
```
z3t0: Sorry if this is a silly question but what does stream here refer to ?

jolange: A mp3/ogg stream, e.g. web radio.

z3t0: ahh got it

mahkoh:
Why is the property called `stream` when the function is called `stream_title`?

Such extensions also need to be documented.

jolange:
I called it `stream` for consistency with the `cmus-remote -Q` output (see #23). But I personally don't mind a different naming like `stream_title`.

I'd be happy to document this, but I did not find any documentation of the MPRIS interface at all. Where would you like to see it?

mahkoh:
stream_title sounds better. Documentation can go in cmus(1). You should probably create a new section that talks about mpris and then a subsection about extensions.

jolange:
@mahkoh I changed the name of the field and added a man page section. Let me know if you think something essential is missing.

mahkoh:
Merged

PR #675, opened 2017-04-18
Add pattern matching for add command

### LINKED ISSUE #659: Pattern matching for :add
Feature request:

When adding all files in a directory, it would be nice to be able to apply a pattern matching filter to remove unwanted file extensions. 

i.e.

```
:add ~/Downloads/*.mp3
```

tagniam:
Resolves #659 with glob patterns. Some examples:
- `:add ~/Music/*.mp3`
- `:add ~/Music/*/*.flac`
- `:add ~/Music/*[Rr]adiohead*`
- `:add ~/Music/?.wav`

Calling `:add` on a file that contains glob metacharacters `[]?*` in its filename must now be escaped:
- `:add ~/Music/[2004] Sung Tongs` &rarr;  `:add ~/Music/\[2004\] Sung Tongs`

Tab completion for `:add` has been modified to escape tab-expanded filenames.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ mpris.o: CFLAGS += $(LIBSYSTEMD_CFLAGS)
 cmus-y := \
 	ape.o browser.o buffer.o cache.o channelmap.o cmdline.o cmus.o command_mode.o \
 	comment.o convert.lo cue.o cue_utils.o debug.o discid.o editable.o expr.o \
-	filters.o format_print.o gbuf.o glob.o help.o history.o http.o id3.o input.o \
+	filters.o format_print.o gbuf.o xglob.o help.o history.o http.o id3.o input.o \
 	job.o keys.o keyval.o lib.o load_dir.o locking.o mergesort.o misc.o options.o \
 	output.o pcm.o player.o play_queue.o pl.o rbtree.o read_wrapper.o search_mode.o \
 	search.o server.o spawn.o tabexp_file.o tabexp.o track_info.o track.o tree.o \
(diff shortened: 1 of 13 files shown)


mahkoh:
I'll have to think about this a bit more.

mahkoh:
Implementation aside, I think there are several issues here:

1. If you add a path with globs, you get no feedback if the glob does not match any files. This is better:

```
~$ ls *.aou
zsh: no matches found: *.aou
```

2. Expansion of the glob blocks the UI thread. This is a big no-no if the glob matches many files.

3. You have to remember to escape certain characters when using add but not when using other commands that use paths. This is too irregular. Globbing should work for all such commands with reasonable semantics. There also needs to be a way to disable globbing ad-hoc the way shells do it with quotation marks. This also needs to work with tab expansion.

Therefore I'll close this PR for now.

tagniam:
Thanks for the feedback @mahkoh. I'll address your points:

1. Straightforward fix: return `FILE_TYPE_INVALID` when `glob()` doesn't return 0.
2. If you mean the fact that `glob()` is MT-unsafe, isn't this unavoidable? Correct me if I'm wrong, but even tab expansion calls `readdir()` which blocks the UI thread in a directory with many files.
3. Good point. I agree, the way cmus handles paths in command mode should be modified to mimic how the shell handles paths for consistency's sake.

I'll take another shot at this when I have the time.

mahkoh:
>Straightforward fix: return FILE_TYPE_INVALID when glob() doesn't return 0.

glob should not be called in the UI thread.

>If you mean the fact that glob() is MT-unsafe

No.

>Correct me if I'm wrong, but even tab expansion calls readdir() which blocks the UI thread in a directory with many files.

That is so but a completely different issue. When the user requests a tab expansion there is no choice but to wait for the data and the user gets immediate feedback. There is no reason for `add` to block the UI thread.

Issue #678, opened 2017-04-22
FLAC plugin's seek is not sample-accurrate


Wang-Yue:
Perhaps other plugins share the same problem. but I just have chance to compare the wav and flac. This may also be the main cause of https://github.com/cmus/cmus/issues/600.

When playing a flac file from specific location(for instance 0:01, or even 0:00), the seeking location is wrong. 

In my test, when asked to play from 0:01, it should output starting from location 176400(which is 44100*2*2) of the raw pcm. but it actually start from location 202752, which missed a lot of samples. to fill the length sometimes redundant samples are copied. Which leads to the issue discussed in https://github.com/cmus/cmus/issues/600. 

Opened this new bug since other plugins are affected too.

mahkoh:
Afaik most formats simply don't support this kind of operation.

Wang-Yue:
@mahkoh right, mp3 doesn't  but no one saves entire cd image as mp3s.  but libflac does support sample accurate seek.  this is of vital importance to cue support. flac should have similar support compared to wav format's.

mahkoh:
A patch implementing accurate seeking in the flac plugin would be accepted.

Wang-Yue:
please apply the following patch:
```
diff --git a/ip/flac.c b/ip/flac.c
index 5a8399b..95a6f22 100644
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -58,8 +58,6 @@ struct flac_private {
 	struct keyval *comments;
 	double duration;
 	long bitrate;
-
-	unsigned int ignore_next_write : 1;
 };
 
 static T(ReadStatus) read_cb(const Dec *dec, unsigned char *buf, size_t *size, void *data)
@@ -161,15 +159,6 @@ static FLAC__StreamDecoderWriteStatus write_cb(const Dec *dec, const FLAC__Frame
 	int ch, nch, i = 0;
 	char *dest; int32_t src;
 
-	if (ip_data->sf == 0) {
-		return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
-	}
-
-	if (priv->ignore_next_write) {
-		priv->ignore_next_write = 0;
-		return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
-	}
-
 	frames = frame->header.blocksize;
 	channels = sf_get_channels(ip_data->sf);
 	bits = sf_get_bits(ip_data->sf);
[...]

PR #679, opened 2017-04-22
Make flac do sample rate accurrate seek

Wang-Yue:
fix https://github.com/cmus/cmus/issues/678

### DIFF
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -58,8 +58,6 @@ struct flac_private {
 	struct keyval *comments;
 	double duration;
 	long bitrate;
-
-	unsigned int ignore_next_write : 1;
 };
 
 static T(ReadStatus) read_cb(const Dec *dec, unsigned char *buf, size_t *size, void *data)
@@ -160,15 +158,6 @@ static FLAC__StreamDecoderWriteStatus write_cb(const Dec *dec, const FLAC__Frame
 	int frames, bytes, size, channels, bits, depth;
 	int ch, nch, i = 0;
 	char *dest; int32_t src;
-
-	if (ip_data->sf == 0) {
-		return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
-	}
-
-	if (priv->ignore_next_write) {
-		priv->ignore_next_write = 0;
-		return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
-	}
 
 	frames = frame->header.blocksize;
 	channels = sf_get_channels(ip_data->sf);
@@ -417,8 +406,7 @@ static int flac_read(struct input_plugin_data *ip_data, char *buffer, int count)
 }
 
 /* Flush the input and seek to an absolute sample. Decoding will resume at the
- * given sample. Note that because of this, the next write callback may contain
- * a partial block.
+ * given sample.
  */
 static int flac_seek(struct input_plugin_data *ip_data, double offset)
 {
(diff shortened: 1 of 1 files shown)


ip/flac.c
```diff
@@ -161,15 +159,6 @@ static FLAC__StreamDecoderWriteStatus write_cb(const Dec *dec, const FLAC__Frame
 	int ch, nch, i = 0;
 	char *dest; int32_t src;
 
-	if (ip_data->sf == 0) {
-		return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
-	}
```
mahkoh: Why?

ip/flac.c
```diff
@@ -427,11 +415,14 @@ static int flac_seek(struct input_plugin_data *ip_data, double offset)
...
+			if (!F(flush)(priv->dec))
+				d_print("failed to flush\n");
+			priv->buf_rpos = 0;
+			priv->buf_wpos = 0;
+		}
 		return -IP_ERROR_ERRNO;
 	}
-	priv->ignore_next_write = 1;
-	priv->buf_rpos = 0;
-	priv->buf_wpos = 0;
```
mahkoh: Why?

Wang-Yue: the FLAC__stream_decoder_seek_absolute  will internally call seek_to_absolute_sample_, which process a frame. See https://github.com/xiph/flac/blob/4f47b63e9c971e6391590caf00a0f2a5ed612e67/src/libFLAC/stream_decoder.c#L3140 . 

so the wirte_cb will be called during flac_seek. 
If buf position is reset to 0, the first block is entirely missing. 

Thus, flushing and buf resetting should only be done when something is wrong.

mahkoh: Then you need to set these before calling seek.

Wang-Yue: No. These should never be reset when things are performing without error. otherwise you might get partial samples if flac_read is not finished through the end, and the stream will be broken.

mahkoh: Let's say that we're seeking from 5s to 10s. At the point of the seek we've already buffered the samples up to 5.5s. If we don't reset these values, then, after the seek, we'll continue to play the samples from 5s to 5.5s and only then start to play the samples at 10s.

Is this the intended behavior?

Wang-Yue: I think the original author is mislead by flac doc, which mentioned the stream may contain a partial block when seeking. So he tried to reset that block. But what flac doc really means is you won't get the entire block as you normally do when using process_single, becauce the sample may live in the middle of a flac block (which contains various samples). But flac ensures you that you won't get partial samples. You'll always get right data stream when doing seek_absolute. There's no need to reset at all.

mahkoh: I don't think you've answered my question. I'm talking about our buffer `flac_private::buf`. Should we not clear this buffer before seeking?

ip/flac.c
Wang-Yue: ip_data->sf should be non zero after _open. if it's zero things are terribly wrong.

ip/flac.c
Wang-Yue: You are right. Done.

ip/flac.c
Wang-Yue: see https://github.com/Wang-Yue/cmus/blob/403b63b4a77dc1068159c19f1a2f3837d5e5849d/ip/flac.c#L354

mahkoh:
```patch
diff --git a/ip/flac.c b/ip/flac.c
index 90a97b6..1bc585e 100644
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -66,6 +66,8 @@ static T(ReadStatus) read_cb(const Dec *dec, unsigned char *buf, size_t *size, v
 	struct flac_private *priv = ip_data->private;
 	int rc;
 
+	fprintf(stderr, "ip_data->sf = %d\n", ip_data->sf);
+
 	if (priv->pos == priv->len) {
 		*size = 0;
 		return E(READ_STATUS_END_OF_STREAM);
```

Tells me that it's zero sometimes.

mahkoh: No, my bad. This was the wrong callback.

mahkoh:
Merged.

Issue #600, opened 2017-01-05
[bug] hiccups between tracks in playing cue+flac files.


Wang-Yue:
Issue description:

some albums could have a few tracks that plays continuously. This is pretty common in classical music (for instance, Beethoven 5th Symphony, Rachmaninoff's Rhapsody on a Theme of Paganini ).

When having the single CD as a whole flac file and an accompanying cue file, cmus would have a tiny gap between the tracks. 

To reproduce this problem, choose any music (for instance, Eagle's Hotel California) you have as a wav or flac file. Then test using the following cue file:
```

TITLE "Test"
PERFORMER "Eagles"
FILE "hotel_california.flac" WAVE
  TRACK 01 AUDIO
    TITLE "Test"
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    TITLE "Test"
    INDEX 01 00:01:00
  TRACK 03 AUDIO
 ...
  TRACK 57 AUDIO
    TITLE "57"
    INDEX 01 00:57:00
  TRACK 58 AUDIO
    TITLE "58"
    INDEX 01 00:58:00
  TRACK 59 AUDIO
    TITLE "59"
    INDEX 01 00:59:00
```

For each second there're high chances you'll hear a small gap in the music.

karlek:
Reproduced in both `cmus v2.7.1` and `cmus v2.8.0-rc0`.

Wang-Yue:
ping

Wang-Yue:
I'm not familiar with the code. Could anyone let me know why that happens? I could help fixing it.

mahkoh:
I've tested this with both flac and wav and the issue only occurs with the flac file so the problem is most likely in the flac plugin or the flac decoder itself.

mahkoh:
Also note that the new cue parser does not care about the type of audio file. You might want to test this with mp3 or opus files.

Wang-Yue:
confirmed this is a flac plugin bug. 
wav plays fine.

I haven't tried mp3/opus yet, as I don't have those libs installed.

Wang-Yue:
is anyone interested in fixing this issue? I'm not familiar with libflac

Wang-Yue:
tried libmad with mp3 file. same issue. although not as obvious as flac. but has similar hiccups if listened carefully.

Wang-Yue:
I tried to output the everything the cue plugin writes to a single file. flac and equivalent wav would have very different result. 

sample
cat test/hotel.cue 
[...]

PR #680, opened 2017-04-22
fix cue rounding to make it sample rate accurate

Wang-Yue:
this will fix https://github.com/cmus/cmus/issues/600


mahkoh:
Merged.

PR #701, opened 2017-05-26
Added ancillary libraries to use when building cmus.

solnet-phiro:
Hi, added some information about what libraries you need to have installed to build a usable cmus instance.


mahkoh:
Let's put it in the wiki.

flyingmutant:
Yes, wiki would be the preferred place.

PR #710, opened 2017-06-24
Fix #574 (deadlocks in player.c)

### LINKED ISSUE #574: Deadlocks in player.c
We have a serious problem with deadlocks in player.c. When the player asks the UI thread for the next track, it will block until the UI thread responds and continues to hold one of the player locks while it waits. If the UI thread at the same time tries to acquire one of the player locks for some reason (e.g. to change the track manually) we have a deadlock.

I've discovered this while debugging another problem that caused `cmus_get_next` to be called in rapid succession. Luckily this doesn't seem to happen that often normally.

Right now I see two ways to fix this:
- Make `cmus_get_next` async: This is hard because the player code is very complicated and uses a completely synchronous design.
- Release the locks while waiting in `cmus_get_next`: This is easier but still quite complicated. It is unclear which invariants could be violated if we temporarily release the lock.

mahkoh:
Tl;dr: The UI thread no longer waits for the player to perform its requests (player_next, etc.) Instead it simply schedules the commands and returns immediately. This allows the UI thread to reply to `cmus_get_next` requests that occur while the player holds its locks.

Fixes #574.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ cmus-y := \
 	job.o keys.o keyval.o lib.o load_dir.o locking.o mergesort.o misc.o options.o \
 	output.o pcm.o player.o play_queue.o pl.o rbtree.o read_wrapper.o search_mode.o \
 	search.o server.o spawn.o tabexp_file.o tabexp.o track_info.o track.o tree.o \
-	uchar.o u_collate.o ui_curses.o window.o worker.o xstrjoin.o
+	uchar.o u_collate.o ui_curses.o window.o worker.o xstrjoin.o delegate.o
 
 cmus-$(CONFIG_MPRIS) += mpris.o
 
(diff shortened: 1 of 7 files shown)


mahkoh:
I'm still figuring out some places where the code assumed that the functions were synchronous.

flyingmutant:
Thanks a lot for the PR! The review can take some time though, sorry about that.

mahkoh:
Don't worry about it. It's not done yet though so don't bother with the review yet.

tagniam:
@mahkoh Will the delegate functions in 0654847 allow any command to be run asynchronously? They look like they would help me in addressing the issues of #675.

Wang-Yue:
This have been blocking the release for so long.

flyingmutant:
True.

Wang-Yue:
Please revert this commit. This breaks CoreAudio's mixer initializer.

PR #715, opened 2017-07-16
Found use after free and possible null dereference

jayay:


track_info.c
```diff
@@ -383,14 +383,17 @@ void sort_keys_to_str(const sort_key_t *keys, char *buf, size_t bufsize)
 
 	for (i = 0; keys[i] != SORT_INVALID; i++) {
 		const char *key = sort_key_to_str(keys[i]);
-		int len = strlen(key);
 
-		if ((int)bufsize - pos - len - 2 < 0)
-			break;
+		if (key != NULL) {
```
flyingmutant: `sort_key_to_str(key[i])` can not return `NULL`. We consider `keys` to consist of valid keys only (and end with `SORT_INVALID`).

input.c
```diff
@@ -468,8 +468,8 @@ static int open_file_locked(struct input_plugin *ip)
 		if (!ops)
 			break;
 
-		ip_reset(ip, 0);
 		d_print("fallback: try next plugin for `%s'\n", ip->data.filename);
+		ip_reset(ip, 0);
```
flyingmutant: `ip_reset()` does not free `ip->data.filename`.

Issue #215, opened 2014-11-11
Feature: Scroll half page up/down and H, M, L motions.


JosephFerano:
I would be very much interested in vim's equivalent to C-d and C-u, as well as H (top of page), M (middle of page), L (bottom of page). I tend to just import my music collection and browse around with C-f and C-b until I find a song I would like to play so it would be awesome to have those additional movement keys to help facilitate the browsing. I know some C, I wouldn't mind giving it a shot if someone could point me to the right place :D

flyingmutant:
Fixed by #721.

PR #721, opened 2017-08-05
Make ^U/^D default for half-page up/down motions

tagniam:
By default vim binds `^U` and `^D` to half-page up/down, so cmus should as well.


flyingmutant:
Thanks! Sorry for huge delay in merging this. Fixes #215.

Issue #221, opened 2014-12-03
Add quick fade in/out to controls (reduce clicking noise)


BorisVerk:
Any time I pause/play/fast forward etc I hear a clicking noise from my speakers. I would guess this is because cmus literally stops play back as soon as you hit pause, so the speaker is slammed down to the resting position, causing an audible noise. If I'm not mistaken this can actually damage speakers. I believe the right way to do it is add a very very short fade out to bring the membrane of a speaker back to rest in a controlled manner, the same goes for playing where you have to really quickly fade in.

For reference, I am using Arch and ALSA. I would really like to see this feature implemented, it's extremely annoying. To reproduce just go into any loud part of a song (it'll be more obvious) and tap `C` a couple times.

YellowOnion:
Playback needs to stop on a zero crossing or fade out, otherwise output will click, shifting the output value from something non-zero, to zero, [causes a full band impulse to happen.](http://www.ohdratdigital.com/tutorials-2/zero-crossing-points-what-are-they/)

Fade-out/in is a must have feature in most audio players anyway, it makes pausing/unpausing a lot less jarring.

Is there a possibility to have this feature request reopened?

PR #722, opened 2017-09-02
Fade on pause/unpause

rianmcguire:
I've had a go at adding a fade to pause/unpause (https://github.com/cmus/cmus/issues/221).

It's implemented by adjusting the output's mixer volume. That makes sense for audio interfaces where the volume is for the stream only (ie. PulseAudio, CoreAudio, Roar?). However, if the mixer is actually for the system volume it'll have the side effect of fading other applications. The default fade duration is only 100ms so it might not be a big deal, but I'm interested to get some feedback. It would be easy enough to expose a per-plugin flag that enables fading instead.

I've only tested with PulseAudio so far.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -946,6 +946,11 @@ display_artist_sort_name (false)
 	regular ones in tree view (e.g. "Artist, The" instead of "The Artist"),
 	so that artists column looks alphabetically sorted.
 
+fade_duration (0)
+	Fade output when pausing/unpausing. Duration is in milliseconds.
+
+	NOTE: has no effect if output_plugin doesn't support volume control.
+
 follow (false)
 	If enabled, always select the currently playing track on track change.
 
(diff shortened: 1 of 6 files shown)


mahkoh:
1. Documentation goes in Doc/cmus.txt
2. This implementation is blocking: `ms_sleep(FADE_STEP_MSEC)`. Usually such code would not be accepted but since this feature is optional, I will merge it as long as the other issues are addressed.
3. Because of 2, the default must be set to 0.
4. Global variables go to the top of the file and should be static if possible.
5. errno must not be set to the return value of mixer_read_volume because this function might itself set errno.
6. A block with only a single statement must not be surrounded by braces.
7. There is a lot of duplicate code between mixer_fadein_end and mixer_fadeout.
8. These functions divide by 0 if 0 < fade_duration < FADE_STEP_MSEC.

rianmcguire:
Thanks @mahkoh - should be good to re-review now. Let me know if I should squash my fixes.

nefthy:
I don't favour the approach of mangling with the mixer. Also I don't like the for (...) sleep(). It would be better to do something along the lines of what softvol does.

nefthy:

PR #731, opened 2017-10-09
handle "REM COMPILATION TRUE" in cue generated by XLD

woremacx:
XLD ( http://tmkk.undo.jp/xld/index_e.html ) creates compilation cue as following.
This pr supports such compilation.

```
REM COMPILATION TRUE
```


flyingmutant:
Thanks! Merged.

PR #735, opened 2017-10-14
added different install methods

mehandes:
Added some quick methods of cmus installation (to encourage more people to try it).


flyingmutant:
I think it is better to have just once sentence linking to a wiki or cmus.github.io page which instructs how to install a premade package (and warns that they are often outdated). `README.md` is meant more for developers, not end-users.

PR #738, opened 2017-10-18
vim-like :set command

ForTheReallys:
I'm not sure if this is a huge deal but it's something I miss from vim.  I've changed the `:set` command so it works exactly like it does in vim with the exception of accepting multiple options in a single line.  This includes commands like `:set nocontinue` and `:set continue!`

NOTE this DOES change the way commands like `:set continue` works in cmus.  Now, when the option expects a boolean value, its value is set to true.  Otherwise, `:set` behaves as it used to.

Let me know if I should change anything.


flyingmutant:
It may be good idea, but a backward-incompatible one, which I do not like. @mahkoh thoughts?

ForTheReallys:
My mistake, this doesn't include vim's features to invert an option through something like `:set invcontinue` or reset an option to its default value through something like `:set continue&`.  I'll start working on the `inv` syntax tomorrow (depending on any feedback).  I can also start working on setting options to their default values but I feel like that should be another commit.  Let me know what you think. Thanks!

nefthy:
@flyingmutant could you explaine the incompatibility?

ForTheReallys:
It changes the way 
```
:set <boolean>
```
works. Instead of printing the variable's value (like previous versions) it sets the value to true

araspik:
How about a `vim-options` option? When set, it will emulate Vim-style options. Otherwise, it will use the old setup. By default it would be disabled, and newcomers can enable it if they want. Eventually, it would be enabled by default (whenever minor version is bumped?)

phinxy:
I´ve been struggling with how to set an option to the default value.  :set lib_sort& would be a great addition, it was the first thing I tried.
As an example I wanted to sort a couple albums after date.  lib_sort did exactly that but the problem is resetting cmus to a default state.  I would have to look up the default value in the manpage and type it in manually.

pgaskin:
Has a decision been made about this?

nefthy:
I don't think that there was any decision. Breaking compatibility is an issue here however. Making the behaviour optional is, also suboptimal, as only a small subgroup of users will discover it and use it, while there is the cost of having multiple code paths that need to be maintained.

IMO we should do it when/if we ever bump the major version.

nefthy:

Issue #756, opened 2017-12-07
Allow to raise the window


kitsunyan:
A lot of terminal emulator applications place `WINDOWID` envvar with X window id. This can be used to raise the terminal window with running cmus application.

I want to suggest implementing `Raise` method of MPRIS proto using this envvar.

Using `shell` command it's possible even now: `shell xdotool windowraise $WINDOWID` (can be useful with `cmus-remote`). Probably this long command could be implemented natively as well.

flyingmutant:
Fixed in #757.

PR #757, opened 2017-12-08
Add command to raise VTE window + implement MPRIS Raise method

kitsunyan:
I've implemented what I described before in https://github.com/cmus/cmus/issues/756.

It detects X11 system, loads `libX11` using `dlopen` and raises the terminal window.

It also adds `raise-vte` command and implements `Raise` method in MPRIS protocol.


flyingmutant:
Thanks! Sorry for the delayed merge.

PR #759, opened 2017-12-11
Add color_win_cur_attr

jcs:
I prefer a monochrome color scheme and just use attributes like bold to make things stand out.  It was only possible to set the color of `win_cur` but not the attributes, so this adds `color_win_cur_attr`.

I left `attr_names`/`attrs` and the `COLOR_WIN_CUR_ATTR` enum unsorted because they weren't previously.


flyingmutant:
Merged! Sorry for the huge delay.

PR #800, opened 2018-05-12
Play history in a playlist

theojepsen:
This is a simple solution to keeping track of your recently played tracks. Every time you load a new file to play, it is added to a playlist called "play_history".

### DIFF
--- a/pl.c
+++ b/pl.c
@@ -198,19 +198,22 @@ static void pl_free(struct playlist *pl)
 	free(pl);
 }
 
-static void pl_add_track(struct playlist *pl, struct track_info *ti)
+static void pl_add_track(struct playlist *pl, struct track_info *ti, int before)
 {
 	struct shuffle_track *track = xnew(struct shuffle_track, 1);
 
 	track_info_ref(ti);
 	simple_track_init(&track->simple_track, ti);
 	shuffle_list_add(track, &pl->shuffle_root);
-	editable_add(&pl->editable, &track->simple_track);
+    if (before)
+        editable_add_before(&pl->editable, &track->simple_track);
+    else
+        editable_add(&pl->editable, &track->simple_track);
 }
 
 static void pl_add_cb(struct track_info *ti, void *opaque)
 {
-	pl_add_track(opaque, ti);
+	pl_add_track(opaque, ti, 0);
 }
 
 int pl_add_file_to_marked_pl(const char *file)
@@ -226,15 +229,15 @@ int pl_add_file_to_marked_pl(const char *file)
 
 void pl_add_track_to_marked_pl(struct track_info *ti)
 {
-	pl_add_track(pl_marked, ti);
+	pl_add_track(pl_marked, ti, 0);
 }
 
-void pl_add_track_to_named_pl(const char *name, struct track_info *ti)
+void pl_add_track_to_named_pl(const char *name, struct track_info *ti, int before)
 {
 	struct playlist *pl;
 	list_for_each_entry(pl, &pl_head, node)
 		if (strcmp(name, pl->name) == 0)
-			pl_add_track(pl, ti);
+			pl_add_track(pl, ti, before);
 }
 
 static int pl_list_compare(const struct list_head *l, const struct list_head *r)
--- a/pl.c
+++ b/pl.c
@@ -229,6 +229,14 @@ void pl_add_track_to_marked_pl(struct track_info *ti)
 	pl_add_track(pl_marked, ti);
 }
 
+void pl_add_track_to_named_pl(const char *name, struct track_info *ti)
+{
+	struct playlist *pl;
+	list_for_each_entry(pl, &pl_head, node)
+		if (strcmp(name, pl->name) == 0)
+			pl_add_track(pl, ti);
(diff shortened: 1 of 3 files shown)


flyingmutant:
This adds tracks to the end of playlist, right? Adding to the start probably makes more sense.

theojepsen:
Hi @flyingmutant 
Yes, it adds to the end of the playlist. I agree, adding to the start would make more sense. However, it doesn't seem that the playlist API in pl.h supports adding to the start. I suppose I could add that.

After a few days of using this myself, I don't think it does a good job at detecting "played" tracks. What does it mean for a track to be played? Initially, a track was added every time `file_changed()` was called. If you are just skipping through tracks, even without playing them, they would be added -- incorrectly IMO -- to the history. As of the latest commit fa7f56d, a track is only added if `PLAYER_STATUS_PLAYING`. This works well when the player is already playing. However, if the player is stopped, and you start playing a file, it won't be added to the history, because when `file_changed()` is called, the playre status is not  `PLAYER_STATUS_PLAYING` yet. Do you have any suggestions for detecting the "started playing a new track" event?

flyingmutant:
I think the real solution is to define and properly implement a heuristic like "at least N seconds of a track were played uninterrupted". Trying to hack around using existing infrastructure instead is probably not the way to go.

jayay:
The cached `play_count` [gets incremented at the reach of EOF](https://github.com/cmus/cmus/blob/8cccf702a348bd19670217f824c8e5d8e336381f/player.c#L807).

@theojepsen no matter what event you decide to choose, both incrementing `play_count` and adding to the play_history should happen together, just to keep the behavior consistent.

theojepsen:
@jayjay are you suggesting that we can detect when a track has been played just the way it is done for `play_count`? i.e. we can hook in after `play_count++` to add the current track to the play history playlist?

jayay:
That would be ok for now IMO.

The solution proposed by @flyingmutant would be more intuitive to the user, though. It should also apply to the play_count then.

theojepsen:
@jayay with the latest commit the play history is updated together with the play count.

jayay:
The playlist `play_history` doesn't get created automatically. That's probably because my `default` playlist already exists. Is that on purpose?

theojepsen:
@jayay if you start cmus for the first time `~/.config/cmus/playlists/` does not exist, so the default playlist is created (i.e. `pl_create_default()` is called).  When this happens, the `play_history` playlist is also created: https://github.com/cmus/cmus/pull/800/files#diff-44c37d81b6353408d48da66235538910R290

jayay:
I get the point but my setup was there already. I checked out your branch, recompiled but there was no sort of migration because `pl_create_default()` gets called if `(list_empty(&pl_head))` which is not the case if _only_ `play_history` is missing.

This did the job for me, though:
```diff
diff --git a/pl.c b/pl.c
index 58d3406..7a69e63 100644
--- a/pl.c
+++ b/pl.c
@@ -286,11 +286,16 @@ static void pl_create_default(void)
 {
        struct playlist *pl = pl_new("default");
        list_add_tail(&pl->node, &pl_head);
+}
 
-       pl = pl_new("play_history");
-       list_add_tail(&pl->node, &pl_head);
+
+static void pl_create_play_history(void)
+{
+        struct playlist *pl = pl_new("play_history");
+        list_add_tail(&pl->node, &pl_head);
 }
 
+
 static GENERIC_ITER_PREV(pl_list_get_prev, struct playlist, node);
 static GENERIC_ITER_NEXT(pl_list_get_next, struct playlist, node);
 
@@ -571,6 +576,9 @@ void pl_init(void)
        if (list_empty(&pl_head))
                pl_create_default();
 
+       if (!pl_name_exists("play_history"))
+               pl_create_play_history();
+
        pl_sort_all();
 
        pl_list_win = window_new(pl_list_get_prev, pl_list_get_next);
```

theojepsen:
@jayay thanks, that's an elegant way of handling the migration. The latest commit does this now.

flyingmutant:
@nefthy, sorry I don't agree: what do we gain be making this feature optional? Either it is turned on by default, and nobody will bother to turn it off (because why, it does not break or change anything), or it is turned off by default, and the vast majority will not benefit from it.

It would be nice to limit the length of `play_history`, as @jayay suggested. Generating files of unbounded length is not a good default for a console music player.

nefthy:
@flyingmutant the unlimited growth was one of my concerns. I'd prefer it if the history can be disabled. This is not hard to implement.

flyingmutant:
I think configurable history length with 0 meaning "disabled" gives us all the control we need.

nefthy:
That's fine with me.
@theojepsen do you want to incorporate the changes above with the history_length or should I take over?

theojepsen:
@nefthy yes, please take over. I think you know better than me how to add the `history_length` option.

nefthy:
Removing tracks from playlists is far from trivial. The playlist code, though functional will not win any design awards.

inspell:
Hello!
Will this get merged to master soon? I really like this feature.

nefthy:

PR #824, opened 2018-08-16
Fix #27 - Add documentation for passwd option

### LINKED ISSUE #27: 'passwd' option is undocumented
Under OPTIONS in the man page, the 'passwd' option is undocumented. It is used in conjunction with the --listen option and should probably be mentioned there too.

cheukyin699:
Fixes #27.


flyingmutant:
Thanks!

PR #836, opened 2018-11-19
Change to get_non_empty_env for CMUS_PLAYLIST_DIR.

sp1ff:
While doing some testing, I typed "export CMUS_PLAYLIST_DIR=; ./cmus". Not only did that break cmus, it produced a very unhelpful error message. Better to use get_non_empty_env.


flyingmutant:
Good catch, thanks!

PR #839, opened 2018-11-22
Add two new playlist commands: `pl-flush' & `pl-reload'.

sp1ff:
This commit does the following:

  1. Adds two new commands:

     a. `pl-flush' write all playlists to disk; this is what
         happens on exit today
     b. `pl-reload' -- walk all files in the playlists
        directory & load them; existing playlists *not* backed
        by file will not be touched

  2. Housekeeping-- added gtags tags files to .gitignore &
     *.lo to `make clean'; documented the CMUS_PLAYLISTS_DIR
     environment variable


command_mode.c
```diff
@@ -1268,6 +1268,11 @@ static void cmd_pl_export(char *arg)
 		info_msg(":pl-export only works in view 3");
 }
 
+static void cmd_pl_flush(char *arg)
```
nefthy: Why not pl_save?

command_mode.c
```diff
@@ -2552,7 +2562,9 @@ struct command commands[] = {
 	{ "right-view",            cmd_right_view,       0, 0,  NULL,                 0, 0          },
 	{ "pl-create",             cmd_pl_create,        1, -1, NULL,                 0, 0          },
 	{ "pl-export",             cmd_pl_export,        1, -1, NULL,                 0, 0          },
+	{ "pl-flush",              cmd_pl_flush,         0, 0,  NULL,                 0, 0          },
```
nefthy: pl-save?

.gitignore
```diff
@@ -22,3 +22,7 @@ cscope.out
 /cmus.base
 /cmus.def
 /cmus.exp
+GPATH
+GRTAGS
+GSYMS
+GTAGS
```
nefthy: Not related to the pl-flush/reload. Also specific to your workflow.

Doc/cmus.txt
```diff
@@ -1532,6 +1539,9 @@ CMUS_CHARSET
 CMUS_HOME
 	Override cmus config directory (default: $XDG_CONFIG_HOME/cmus).
 
+CMUS_PLAYLIST_DIR
+	Override the cmus playlist directory (default: $XDG_CONFIG_HOME/cmus/playlists).
+
```
nefthy: If taken stricktly this is also not related to the pl-flush/reload functionality. One issue per PR.

Makefile
```diff
@@ -264,7 +264,9 @@ quiet_cmd_ttman = MAN    $@
 
 data		= $(wildcard data/*)
 
-clean		+= *.o ip/*.lo op/*.lo ip/*.so op/*.so cmus libcmus.a cmus.def cmus.base cmus.exp cmus-remote Doc/*.o Doc/ttman Doc/*.1 Doc/*.7 .install.log
+clean		+= *.o *.lo ip/*.lo op/*.lo ip/*.so op/*.so cmus libcmus.a cmus.def \
+		   cmus.base cmus.exp cmus-remote Doc/*.o Doc/ttman Doc/*.1 Doc/*.7 \
+	           .install.log
```
nefthy: ok, but see above.

pl.c
```diff
@@ -271,6 +271,23 @@ static void pl_load_all(void)
 	dir_close(&dir);
 }
 
+/*
+ * Find the playlist struct corresponding to a given name. Return null if no
+ * such playlist exists. Note that `name' is stricly the playlist name, not the
+ * full path to the file.
+ *
```
nefthy: no unnecesery empty lines please.

pl.c
```diff
@@ -271,6 +271,23 @@ static void pl_load_all(void)
...
+ */
+
```
nefthy: no unnecesery empty lines please.

pl.c
```diff
@@ -396,6 +413,52 @@ static void pl_cancel_add_jobs(struct playlist *pl)
 	worker_remove_jobs_by_cb(pl_match_add_job, pl);
 }
 
+/*
+ * (Re)load all files in cmus_playlist_dir.
+ *
+ * Walk cmus_playlist_dir; for each playlist file therein, either re-load the
+ * corresponding playlist in `pl_head' (if there is one), or load it (if there
+ * is not).
+ *
```
nefthy: no unnecesery empty lines please.

pl.c
```diff
@@ -396,6 +413,52 @@ static void pl_cancel_add_jobs(struct playlist *pl)
...
+ *
+ */
+
```
nefthy: no unnecesery empty lines please.

pl.c
```diff
@@ -396,6 +413,52 @@ static void pl_cancel_add_jobs(struct playlist *pl)
...
+void pl_reload(void)
```
nefthy: If I would call pl_reload(); at start instead of pl_load(); should't the result be the same? Is pl_load still necesery?

pl.c
```diff
@@ -252,7 +252,7 @@ static void pl_load_one(const char *file)
 	free(full);
 }
 
-static void pl_load_all(void)
```
nefthy: Is this function needed outside of pl.c?

nefthy:
Actually `save -p` already does the pl-flush part

Makefile
geek1011: This should not be necessary anymore, `*.lo` has been added in #864.

pl.c
sp1ff: I don't see a function named `pl_load`?

command_mode.c
sp1ff: Renamed.

command_mode.c
sp1ff: Renamed

.gitignore
sp1ff: Removed

Doc/cmus.txt
sp1ff: Removed.

Makefile
sp1ff: Removed.

pl.c
sp1ff: removed

pl.c
sp1ff: removed

pl.c
sp1ff: removed

pl.c
sp1ff: removed

pl.c
sp1ff: Ah-- perhaps you mean `pl_load_all`? The two are similar, but different; `pl_load_all` assumes it is being called at startup, and so just runs through the files in the playlist directory, unconditionally creating new playlist structures for each one (via `pl_load_one`).  `pl_reload` has to take into account the possibility that there is already a playlist struct in-memory corresponding to each file processed.

Perhaps they could be merged, with a parameter to distinguish the two cases?

pl.c
sp1ff: Nice catch-- it is *not*. Will fix

sp1ff:
> Actually `save -p` already does the pl-flush part

Damn... how did I miss that? Or did it come in after I opened this PR? Regardless, will update.

pl.c
flyingmutant: Please leave `pl_load_all` static.

pl.c
flyingmutant: Based on your description, shouldn't `pl_reload` just make `pl_load_all` obsolete?

command_mode.c
```diff
@@ -2585,7 +2595,9 @@ struct command commands[] = {
 	{ "right-view",            cmd_right_view,       0, 0,  NULL,                 0, 0          },
 	{ "pl-create",             cmd_pl_create,        1, -1, NULL,                 0, 0          },
 	{ "pl-export",             cmd_pl_export,        1, -1, NULL,                 0, 0          },
+	{ "pl-save",               cmd_pl_save,          0, 0,  NULL,                 0, 0          },
```
flyingmutant: I think it is better to have the commands (as well as the docs) keep the alphabetic order inside their sections.

Issue #844, opened 2018-12-11
elogind support


shibotto:
Hi all! Could you please add elogind support for MPRIS? It supports sd-bus and I'm using it right now. The only thing I needed to change was https://github.com/cmus/cmus/blob/f5a442a041ee30b8a6bf8a171fa295b02d07481e/configure#L292 to
`pkg_config LIBSYSTEMD "libelogind"`

geek1011:
@nefthy, shouldn't this issue be closed?

nefthy:
Yes, it has been merged, but did not have the "Fixes #XXX" in the description, that does the closing automagic.

PR #846, opened 2018-12-19
Add support for elogind

shibotto:
Support for elogind, as already reported here: https://github.com/cmus/cmus/issues/844

### DIFF
--- a/configure
+++ b/configure
@@ -287,9 +287,9 @@ check_vorbis()
 	fi
 }
 
-check_libsystemd()
+check_logind()
 {
-	pkg_config LIBSYSTEMD "libsystemd"
+	pkg_config LOGIND "libsystemd" || pkg_config LOGIND "libelogind >= 239.3"
 	return $?
 }
 
@@ -581,7 +581,7 @@ check check_bass       CONFIG_BASS
 check check_mpc        CONFIG_MPC
 check check_vorbis     CONFIG_VORBIS
 check check_opus       CONFIG_OPUS
-check check_libsystemd CONFIG_MPRIS
+check check_logind     CONFIG_MPRIS
 check check_wavpack    CONFIG_WAVPACK
 check check_mp4        CONFIG_MP4
 check check_aac        CONFIG_AAC
--- a/configure
+++ b/configure
@@ -287,9 +287,9 @@ check_vorbis()
 	fi
 }
 
-check_logind()
+check_libsystemd()
 {
-	pkg_config LOGIND "libsystemd" || pkg_config LOGIND "libelogind >= 239.3"
+	pkg_config LIBSYSTEMD "libsystemd" || pkg_config LIBSYSTEMD "libelogind >= 239.3"
 	return $?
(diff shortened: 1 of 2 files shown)


flyingmutant:
Sorry for such a delayed reply. Aren't elogind headers in `<elogind/...>` instead of `<systemd/...>`?

shibotto:
It's OK :D
Actually elogind is meant to be a drop-in replacement of systemd's logind. For that purpose all the headers are stored in `/usr/include/elogind/systemd/*` and [pkg-config](https://github.com/elogind/elogind/blob/ec32681ee46e16f1dad6848f5a83583bc550805d/src/libelogind/libelogind.pc.in) takes care of importing the right directory, so the source code doesn't need to be altered.

flyingmutant:
I have a feeling that it is better to limit this patch to just this line: https://github.com/cmus/cmus/pull/846/files#diff-e2d5a00791bce9a01f99bc6fd613a39dR292, because AFAIK there is nothing really logind-specific in the APIs cmus uses, they are just a part of systemd's sd-bus. @mahkoh any thoughts?

nefthy:
Patch LGTM

geek1011:
@flyingmutant Can this be merged either way?

configure
```diff
@@ -287,9 +287,9 @@ check_vorbis()
 	fi
 }
 
-check_libsystemd()
+check_logind()
 {
-	pkg_config LIBSYSTEMD "libsystemd"
+	pkg_config LOGIND "libsystemd" || pkg_config LOGIND "libelogind >= 239.3"
```
flyingmutant: Can we only leave this bit in the patch? So that `LIBSYSTEMD_` stays etc.? I don't like having `elogind` names around.

geek1011: Yes, that should work.

shibotto: I just wanted to let you discuss this "internally" first, but I actually don't mind 😀

I'm not quite fond myself of the `LOGIND` change, but `LIBSYSTEMD` seemed even more generic, as only sd-bus is required. Anyway, pushing this!

nefthy: Why libelogind >= 239.3? Are earlier versions not compatible?

shibotto: @nefthy if I remember correctly version previous to 239.3 and 238.3 didn't have complete sd-bus support. I'm not sure how to correctly address this scenario.

geek1011: Yes, that's correct, but AFAIK, it should work fine quite a few versions back too.

nefthy: I'm OK with the check. Just wanted to make sure it is not arbitrary.

nefthy:
There is something wrong here. There are changes in the diff, that stem from other PRs. It seems that upstream was merged into this branch instead of rebasing the branch on top of master.

please rebase.

shibotto:
Sorry, I wanted to finish this before going out this morning, but doing things in a hurry is never a good a idea :/

nefthy:
Looks good. Thanks for your work.

PR #854, opened 2019-01-17
Fix build error on CI

jgoldfar:
Also, add OSX build (though that can be removed if you prefer.)


flyingmutant:
Thanks! Although the script looks a bit hacky, this is definitely an improvement over broken status quo.

PR #857, opened 2019-02-01
Make contrib/cmus-update*.py scripts use env python and fix pidgin with py3

knixeur:
Changed shebang to use /usr/bin/env to get the python interpreter
updatepidgin: We simple remove the useless print without parenthesis.


flyingmutant:
Thanks!

PR #858, opened 2019-02-07
Fixes hanging when terminal is closed while promting

### LINKED ISSUE #765: cmus hangs if the window is killed
---

1) Linux arch 4.14.11-1-ARCH #1 SMP PREEMPT Wed Jan 3 07:02:42 UTC 2018 x86_64 GNU/Linux

```
(paste output here)
```

2) `cmus --version`:
```
cmus v2.7.1
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  modplug:
    Priority: 50
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  aac:
    Priority: 50
    File Types: aac

  pulse
  alsa
  oss
  ao

```

open gnome-terminal
start cmus
press q
kill window

-> 100% cpu usage -> SIGKILL

nefthy:
Fixes #765

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -1599,7 +1599,7 @@ void error_msg(const char *format, ...)
 	}
 }
 
-int yes_no_query(const char *format, ...)
+enum ui_query_answer yes_no_query(const char *format, ...)
 {
 	char buffer[512];
 	va_list ap;
@@ -1623,12 +1623,21 @@ int yes_no_query(const char *format, ...)
 
 	while (1) {
 		int ch = getch();
-
-		if (ch == ERR || ch == 0)
+		if (ch == ERR || ch == 0) {
+			if (!cmus_running) {
+				ret = UI_QUERY_ANSWER_ERROR;
+				break;
+			}
 			continue;
-		if (ch == 'y')
-			ret = 1;
-		break;
+		}
+
+		if (ch == 'y') {
+			ret = UI_QUERY_ANSWER_YES;
+			break;
+		} else {
+			ret = UI_QUERY_ANSWER_NO;
+			break;
+		}
 	}
 	update_commandline();
 	return ret;
@@ -1885,6 +1894,7 @@ static void sig_int(int sig)
 
 static void sig_shutdown(int sig)
 {
+	d_print("sig_shutdown %d\n", sig);
 	cmus_running = 0;
 }
 
--- a/browser.c
+++ b/browser.c
@@ -449,7 +449,7 @@ void browser_delete(void)
 		char *name;
 
 		name = fullname(browser_dir, e->name);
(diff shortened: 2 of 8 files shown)


ui_curses.c
```diff
@@ -1624,11 +1624,11 @@ int yes_no_query(const char *format, ...)
 	while (1) {
 		int ch = getch();
 
-		if (ch == ERR || ch == 0)
-			continue;
-		if (ch == 'y')
+		if (ch == 'y' || !cmus_running) {
```
flyingmutant: why do we treat `!cmus_running` as "yes"?

nefthy: Well !cmus_running means that  cmus should exit, because it received a SIGTERM for example or something along the lines. If we treat it as 'n' cmus gets zombified and continues playing in the background.

nefthy: if we ignore cmus_running altogether it, cmus falls in an endless loop with 100% cpu utilisation under the conditions described in #765

flyingmutant: `yes_no_query` can be called in different places, and "no" is a better default answer (in general). Imagine if `rm -rf /` when killed while prompting "are you sure" would treat that as "yes" :-)

This fix looks more like a workaround than a solution.

nefthy: yes you are right in part. But returning "no" is also not an option. As is cmus zombifies if we return "no".

To fix this issue we will have to refactor the API to:

```bool yes_no_query(const char const * msg, char * res)```

And then change all invocations to check if the function for success before using the result. I could do that, but I'd prefer if we could work this out ahead of time, so we don't have that much back and forth.

nefthy: Sorry the function signature above is total crap.
Point is we need three return values: yes, no, error

geek1011: Maybe return an enum and check that?

nefthy: Sure, I have already started making the necessary changes.

nefthy:
Please review.

PR #867, opened 2019-02-12
Fixes incorect bit rate declaration for 24bit flac

### LINKED ISSUE #644: ip_read: BUG: rc % (((((((ip->data.sf))) & 0x00000038 ) >> 0) >> 3) * ((((ip->data.sf)) & 0xff000000 ) >> 24)) != 0
Hi there! Seems like i got myself a little bug.  In a track "Corrupt system"(By VHS Glitch, album is Moral Decay), i'm getting this after some time of playing and cmus crashes

```
ip_read: BUG: rc % (((((((ip->data.sf))) & 0x00000038 ) >> 0) >> 3) * ((((ip->data.sf)) & 0xff000000 ) >> 24)) != 0
```
It seems that some of the tracks in this album is causing this bug too. The other ones is playing absolutely great. I'm actually listening CD-rip in flac with ~1500 kbps, so it's not caused by some badly converted song.

1) `uname -a`:
```
Linux laptop.scvh 4.9.5-200.fc25.x86_64 #1 SMP Fri Jan 20 12:24:16 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.7.1
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib64/cmus/ip
  ffmpeg:
    Priority: 300
    MIME Types:
  mad:
    Priority: 55
    File Types: mp3 mp2
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  aac:
    Priority: 50
    File Types: aac
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    MIME Types:
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack

Output Plugins: /usr/lib64/cmus/op
  pulse
  alsa
  ao
```

nefthy:
flac inflates the bit rate to 32bit for 24bit files, but declares 24bit in
the sf structure.

fixes #644

### DIFF
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -217,8 +217,6 @@ static void metadata_cb(const Dec *dec, const FLAC__StreamMetadata *metadata, vo
 				break;
 			case 20:
 			case 24:
-				bits = 24;
-				break;
 			case 32:
 				bits = 32;
 				break;


tomty89:
It doesn't seem to add up to me.

> flac inflates the bit rate to 32bit for 24bit files
@nefthy any actual reference for this? So it is 16-bit aligned? And why was it (apparently) only problematic when it is 6-ch?

geek1011:
@tomty89 I'm not sure about the 6ch, but it doesn't happen in 2ch. I'm not familiar enough with the FLAC format to understand the reason why this happens, but here are a few things I can find after a quick search:

- https://trac.videolan.org/vlc/ticket/20120
- https://forum.videolan.org/viewtopic.php?t=145087
- https://bugzilla.mozilla.org/show_bug.cgi?id=523837
- https://trac.videolan.org/vlc/ticket/10053

Also, see:
```
patrick@dpc01:~/tmp$ sox -n -r 96k -b 24 -c 6 buggy-synth.flac synth 3 sine 500
patrick@dpc01:~/tmp$ ffprobe buggy-synth.flac 
ffprobe version 4.1-1 Copyright (c) 2007-2018 the FFmpeg developers
  built with gcc 8 (Debian 8.2.0-12)
  configuration: --prefix=/usr --extra-version=1 --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --arch=amd64 --enable-gpl --disable-stripping --enable-avresample --disable-filter=resample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librsvg --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-lv2 --enable-omx --enable-openal --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-chromaprint --enable-frei0r --enable-libx264 --enable-shared
  WARNING: library configuration mismatch
  avcodec     configuration: --prefix=/usr --extra-version=1 --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --arch=amd64 --enable-gpl --disable-stripping --enable-avresample --disable-filter=resample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librsvg --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-lv2 --enable-omx --enable-openal --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-chromaprint --enable-frei0r --enable-libx264 --enable-shared --enable-version3 --disable-doc --disable-programs --enable-liblensfun --enable-libopencore_amrnb --enable-libopencore_amrwb --enable-libtesseract --enable-libvo_amrwbenc
  avfilter    configuration: --prefix=/usr --extra-version=1 --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --arch=amd64 --enable-gpl --disable-stripping --enable-avresample --disable-filter=resample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librsvg --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-lv2 --enable-omx --enable-openal --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-chromaprint --enable-frei0r --enable-libx264 --enable-shared --enable-version3 --disable-doc --disable-programs --enable-liblensfun --enable-libopencore_amrnb --enable-libopencore_amrwb --enable-libtesseract --enable-libvo_amrwbenc
  libavutil      56. 22.100 / 56. 22.100
  libavcodec     58. 35.100 / 58. 35.100
  libavformat    58. 20.100 / 58. 20.100
  libavdevice    58.  5.100 / 58.  5.100
  libavfilter     7. 40.101 /  7. 40.101
  libavresample   4.  0.  0 /  4.  0.  0
  libswscale      5.  3.100 /  5.  3.100
  libswresample   3.  3.100 /  3.  3.100
  libpostproc    55.  3.100 / 55.  3.100
Input #0, flac, from 'buggy-synth.flac':
  Metadata:
    COMMENT         : Processed by SoX
  Duration: 00:00:03.00, start: 0.000000, bitrate: 2731 kb/s
    Stream #0:0: Audio: flac, 96000 Hz, 5.1(side), s32 (24 bit)
patrick@dpc01:~/tmp$ soxi buggy-synth.flac 

Input File     : 'buggy-synth.flac'
Channels       : 6
Sample Rate    : 96000
Precision      : 24-bit
Duration       : 00:00:03.00 = 288000 samples ~ 225 CDDA sectors
File Size      : 1.02M
Bit Rate       : 2.73M
Sample Encoding: 24-bit FLAC
Comment        : 'Comment=Processed by SoX'

patrick@dpc01:~/tmp$
```

And in gstreamer:

https://github.com/freedesktop/gstreamer-sdk-flac/blob/7ef6fde752fc4eb68f4e4549952eb8c3c27552c3/src/flac/decode.c#L1123-L1133

I think this is due the fact that there is no 24-bit integer/float.

tomty89:
The thing is I don't see cmus using different native C types for different bits-per-sample like ffmpeg or so does. Instead it uses `char` for everything.

tomty89:
Not sure if it's relevant, but this line looks really weird to me:
https://github.com/cmus/cmus/blob/65d36d5ef4289bca26089f88e9a4aa6c2f915f37/ip/flac.c#L183
I wonder if it should be `<< (sizeof(int32_t) - depth)` instead, as:
https://github.com/cmus/cmus/blob/65d36d5ef4289bca26089f88e9a4aa6c2f915f37/ip/flac.c#L153-L154
No idea what `% nch` is for either.

geek1011:
> Not sure if it's relevant, but this line looks really weird to me:
> I wonder if it should be `<< (sizeof(int32_t) - depth)` instead, as:

That seems to be a good point. @nefthy, @Wang-Yue (you seem to have done most of the FLAC stuff), @flyingmutant what do you think? This would explain a number of unusual things, especially since the size of a decoded sample may not necessarily be the bitrate of the flac (which was the cause of this whole issue, in an indirect way). This seems to have been introduced here: https://github.com/cmus/cmus/commit/edc7974c013907b6693e598092e9a03ddfd8b9e8. Also, before that commit, a comment seems to be indicating a 24-bit flac is shifted to 32.

> No idea what `% nch` is for either.

Each frame consists of a sample for each channel, and that code would handle having more than one frame in buf by making wrapping back around after nch (number of channels) is reached. I'm not totally familiar with that code, so I'm not sure if it's necessary. The only thing which makes me skeptical about that code is the fact that nch will never be exceeded unless frame->header.channels is greater than sf_get_channels (which is set to the number of channels in the streaminfo). EDIT: going back through git blame, it [seems](https://github.com/cmus/cmus/commit/6d306e56332579c716d52b1ca372861b77d9c637) that this exact situation is being handled to prevent segfaults for invalid flacs.

tomty89:
Actually if the decoded samples are always stored as * signed* 32-bit integers, can/should we even bit shift and memcpy it to char buffer like that anyway?

nefthy:
@tomty89 the char buffer is just used as a generic chunk of memory. It's quite usual to it this way in C. Not very elegant, but I do not see that changing any time soon.

nefthy:
@geek1011 that line stuck me as odd when I was looking at the issue, so you are probably right. I will have a closer look at this and ammend the PR if needed.

Btw. my coder heart gets a warm fuzzy feeling, when so many eyes review my changes. Thanks :)

nefthy:
@geek1011 the assertion is not triggered with 2-channels, as the check is only "is the length of the read

The assertion is not triggered, as it only checks that the read chunk is a multiple of the frame size. I have not checked the math, but my guess is, that it incidentally works out for 2 channels.

Wang-Yue:
I'm not sure if it works. Basically if user's DAC supports 24 bits, it should use 24bits for output.
I'd say please test this on DAC that supports 24bits but not 32bits, and see if alsa and coreaudio plays ok.

tomty89:
Never mind I guess there's no need for special treatment as apparently the samples are "right-justified", which I assume means the "sign bit" is bit 23 but not bit 31 (bit 24 - 31 are always 0), so we can indeed "simply" copy it to three `char`s.

I still don't understand what the bit shift is for though. With the original code the shift would normally be a no-op, as `bit` and `depth` should normally be the same value. But with this PR, 24-bit samples would be "left-justified" I assume. I am not seeing any case that the shift can be appropriate/necessary, do any of you?

Note that there are (mainly) two ways to do 24-bit playback. USB audio needs the samples to be 24-bit "packed" (a.k.a. 3 bytes), while others mostly do it unpacked (right-justfied 24-bit in 32-bit). The ALSA output plugin seems to support both as desired, but it seem to determines the which to use base on the input but not the device. (It should not be a problem when it comes to the pulse plugin though as pulse does conversion as necessary.)

TL;DR, I still don't know what exactly is the cause of the OP. If doing 24-bit "unpacked" solve the problem, maybe we should just do it. It's just we need to figure out whether the bit shift is wrong or really necessary (and confirm that it is done in the right way so it works for 12/20-bit files as well). Also we need to confirm that it works for USB audio with the ALSA output plugin. (I wonder if currently it works for non-USB audio.)

nefthy:
> Not sure if it's relevant, but this line looks really weird to me:
> [cmus/ip/flac.c](https://github.com/cmus/cmus/blob/65d36d5ef4289bca26089f88e9a4aa6c2f915f37/ip/flac.c#L183)
> 
> Line 183 in [65d36d5](/cmus/cmus/commit/65d36d5ef4289bca26089f88e9a4aa6c2f915f37)
> 
>  src = LE32(buf[ch % nch][i] << (bits -depth)); 
> 
> I wonder if it should be `<< (sizeof(int32_t) * 8 - depth)` instead, as:

the code is correct as is. 

- depth is the bit depth of the encoded stream. 
- bits is the bit depth of the decoded stream, which is either 8, 16 or 32 and is chosen so that the samples fit. 

So lets say we have an 24bit flac (`depth = 24`), it will be decoded to 32 bit (`bits = 32`). So we need to shift the sample by 8 bits to the left. That is exactly what this code does, `(bits - depth) = (32 - 24) = 8`. 

Same goes for 12 bit which gets decoded to 16. Again `(bit - depth) = (16 - 12) = 4`, as it should be.

nefthy:
@flyingmutant I think we can merge this, as multiple people confirmed that this patch fixes the issue.

tomty89:
> * bits is the bit depth of the decoded stream, which is either 8, 16 or 32 and is chosen so that the samples fit.

This is only true if cmus/flac use different c types to store decoded samples of different bps (as there are no 24-bit type). However it doesn't appear to be the case at all. From what we have looked into, flac always use `int32_t` to store right-justified samples, and cmus always use `char`s.

> So we need to shift the sample by 8 bits to the left.

There's nothing telling us that is true.

I suspect that the files tested so far all have frames without bps specified (so depth = 0 then = bits = 32), so after all still no shift was performed, which make the PR "works" as "passing-through" right-justified samples in 32-bit width is expected to be fine to the output plugins (at least if we ignore whether it truncate every 4 `char`s to 3 `char`s when its target needs 24-bit "packed").

tomty89:
I wonder if the right way is to always declare and copy 32-bit to our buffer (and never perform any bit shift), unless the extra `char`(s) causes a noticeable difference in `memcpy` or so. Can't really hurt otherwise as the output plugins (well, at least the ALSA one) should be capable of truncating samples when necessary anyway.

nefthy:
That is not correct @tomty89.
```	for (i = 0; i < frames; i++) {
		for (ch = 0; ch < channels; ch++) {
			src = LE32(buf[ch % nch][i] << (bits -depth));
			memcpy(dest, &src, bits / 8);
			dest += bits / 8;
		}
	}```
memcpy copies 1,2 or 4 bytes to dest, depending on the value of bits.

**Comment by tomty89** [comment]
Never mind, I thought the usual s32le (in ALSA/pulse sense) are right-justified as the flac decoded but they are actually left-justified. That's why the bit shifting is right and necessary (otherwise the volume gets really small).

Anyway, I just found the real cause of the problem:
https://github.com/cmus/cmus/blob/65d36d5ef4289bca26089f88e9a4aa6c2f915f37/buffer.h#L22-L23

60 has only one 3, while 18 have two, hence the samples aren't aligned to the buffer chunk. Bumping to 90 (or lowering it to 45) fix the problem.

That said, the alignment requirement probably indicates that our approach is primitive and would have problem dealing with 7-ch files anyway. If we for example ditch the the 5 and make it 36 or 72, we would have problems with 5-ch files as well. If we want to cover both of them it probably needs to be 105 (3 * 5 * 7), or 315 (3 * 5 * 7 * 3) if we want to cover 24-bit at the same time.

**Comment by tomty89** [comment]
Perhaps it can be `(315 * 256)`?

**Comment by Wang-Yue** [comment]
I believe my code is correct and it is just simple bit manipulation. Don’t know if the patch works for 24bit dacs. I used to have one but I sold it:(

**Comment by tomty89** [comment]
@Wang-Yue You only missed one bit I think, which is when the frames do not have bps specfied. To cover such case we have to make sure the samples are always left-justified/shifted with referrence to the true depth:
```
diff --git a/ip/flac.c b/ip/flac.c
index 0153d56..12cbdee 100644
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -58,6 +58,7 @@ struct flac_private {
        struct keyval *comments;
        double duration;
        long bitrate;
+       int bps;
 };

 static T(ReadStatus) read_cb(const Dec *dec, unsigned char *buf, size_t *size, void *data)
@@ -175,12 +176,12 @@ static FLAC__StreamDecoderWriteStatus write_cb(const Dec *dec, const FLAC__Frame

        depth = frame->header.bits_per_sample;
        if (!depth)
-               depth = bits;
+               depth = priv->bps;
        nch = frame->header.channels;
        dest = priv->buf + priv->buf_wpos;
        for (i = 0; i < frames; i++) {
                for (ch = 0; ch < channels; ch++) {
-                       src = LE32(buf[ch % nch][i] << (bits -depth));
+                       src = LE32(buf[ch % nch][i] << (bits - depth));
                        memcpy(dest, &src, bits / 8);
                        dest += bits / 8;
                }
@@ -205,6 +206,7 @@ static void metadata_cb(const Dec *dec, const FLAC__StreamMetadata *metadata, vo
        case FLAC__METADATA_TYPE_STREAMINFO:
                {
                        const FLAC__StreamMetadata_StreamInfo *si = &metadata->data.stream_info;
+                       priv->bps = si->bits_per_sample;
                        int bits = 0;

                        switch (si->bits_per_sample) {
@@ -306,7 +308,8 @@ static int flac_open(struct input_plugin_data *ip_data)
        const struct flac_private priv_init = {
                .dec      = dec,
                .duration = -1,
-               .bitrate  = -1
+               .bitrate  = -1,
+               .bps      = 0
        };

        if (!dec)
```

nefthy:
Why?

flyingmutant:
As the patch is simple and people report that it improves the things for them, I'm going to merge it, even though I didn't dig into the code and determine if it is the "correct" fix or not.

In case a better fix will be discovered, please open a new PR.

tomty89:
@nefthy Same reason why you need to bit shift 8 bit when the samples are 24-bit right-justified in the 32-bit (int32_t / 4 chars). You don't want to lose loudness when you express the 12/20-bit samples in 16/24/32-bit, so you need to left shift/justify it to the width we determined for them.

Think of the case with your PR, if the frames don't have bps specified in their header (i.e. 0, which means we should refer to the bps in the metadata), depth will be set to the value of bits which is 32. No bit shift would then be performed to retain the loudness (and pad the missing "details" with zeros).

tomty89:
@flyingmutant Will open PR to introduce the two patch I proposed (and revert this :P)

nefthy:
Yeah, after looking at the specs I do understand what you mean. However, incorrect padding when missing per frame bits per sample marking, is a separate issue. It would help to keep things focused on the topic at hand.

nefthy:
@tomty89 you patch does not fix #644

tomty89:
@nefthy You mean changing the `CHUNK_SIZE` "patch" or the actual patch I pasted? Coz they are addressing different problems and the former is for #644.

nefthy:
5 channels work fine with master. As do 6. Test file created with `sox -n -r 96k -b 24 -c 5 644_5_ch.flac synth 3 sine 500`.

Provide a test case if you still think that something is broken.

tomty89:
I haven't said the current value would not work for 5-ch. You can read more carefully or wait for my PR (I would explain the details, as in, how my proposed value is formed, as comment above the definition.)

nefthy:
Do what you must.

Issue #847, opened 2018-12-21
Cmus freezes when playing Allegaeon's album "Elements Of The Infinite"


ghost:
1) `uname -a`:
```
Linux anarchy 4.19.8-300.fc29.x86_64 #1 SMP Mon Dec 10 15:23:11 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

Distro is Fedora 29, since Fedora don't put that in uname.

2) `cmus --version`:
```
cmus v2.8.0-rc0-93-gf5a442a
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/local/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  opus:
    Priority: 50
    File Types: opus
    MIME Types:

Output Plugins: /usr/local/lib/cmus/op
  pulse
  alsa
  oss
  jack
  ao
```

I know the title is oddly specific, but it's the only album I've noticed the issue with. Sometime through the album, usually on "Through Ages Of Ice - Otzi's Curse" (displayed as Through Ages Of Ice - Otzi<92>s Curse), the interface freezes. once the album finishes, it doesn't repeat despite Cmus being set to repeat the album. No input is processed, and I have to close the terminal window and reopen it to get Cmus working again.

Ashymad:
[...]

PR #868, opened 2019-02-15
Ensures all data sent on dbus is in utf8 (#847)

amangrewal:
Partly addresses #847. All bytes that are invalid utf8 are converted
to \<xx\>, just like other areas in cmus.

### DIFF
--- a/configure
+++ b/configure
@@ -289,7 +289,7 @@ check_vorbis()
 
 check_libsystemd()
 {
-	pkg_config LIBSYSTEMD "libsystemd"
+	pkg_config LIBSYSTEMD "libsystemd" || pkg_config LIBSYSTEMD "libelogind >= 239.3"
 	return $?
 }
 


flyingmutant:
This looks like treating the symptom, not the cause. Shouldn't we use valid UTF-8 internally for everything?

geek1011:
This patch is simple and will fix the bug (which is about cmus freezing, not utf-8 not working) for now without causing any side-effects. IMO, this should be merged until a full solution is decided on, and another issue should be opened for full utf-8 support (and another about the select issue mentioned).

amangrewal:
I agree that this is more of a bandage than a full solution. But as geek1011 said, this fixes a bug many people have complained about.

A better solution would be to escape all invalid bytes when they are read in. This solution would require significantly more time and testing.

nefthy:
I'm not sure on this. It's quite a severe bug as it causes cmus to break down and required `kill -9` and reset of the terminal. So it would be good to mitigate this. 

However, I think @flyingmutant is right, we should fix this error at the source. Our database should contain only valid utf-8 and nothing else. I can not say on the top of my head, how big a change that would be.

In case we keep this patch, the if/else looks a bit smelly. May you could extract a function `static void ensure_valid_utf8(const char const* source, char* corrected)` or something along the lines

flyingmutant:
Yeah, @amangrewal can you please get rid of conditionals and also add a comment that this is a temporary solution to the code? Then we can merge it to have a bandage in place for the time being.

PR #869, opened 2019-02-15
Fixed issue with `xstrndup`

araspik:
Previously, `ape` and a few other modules refused to link, citing a
missing `xstrndup`. This seemed to be due to some issue with
`HAVE_STRNDUP` checks not quite working. Here's a fix. The only issue is
that it may slow down `cmus` a bit since the inline definition (which
was in `xmalloc.h`) has been removed.

### DIFF
--- a/xmalloc.c
+++ b/xmalloc.c
@@ -31,9 +31,14 @@ void malloc_fail(void)
 	exit(42);
 }
 
-#ifndef HAVE_STRNDUP
 char *xstrndup(const char *str, size_t n)
 {
+#ifdef HAVE_STRNDUP
+	char *s = strndup(str, n);
+	if (unlikely(s == NULL))
+		malloc_fail();
+	return s;
+#else
 	size_t len;
 	char *s;
 
@@ -43,5 +48,5 @@ char *xstrndup(const char *str, size_t n)
 	memcpy(s, str, len);
 	s[len] = 0;
 	return s;
-}
 #endif
+}
--- a/xmalloc.h
+++ b/xmalloc.h
@@ -73,17 +73,7 @@ static inline char * CMUS_MALLOC xstrdup(const char *str)
 #endif
 }
 
-#ifdef HAVE_STRNDUP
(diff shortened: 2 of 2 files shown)


nefthy:
- Why exactly is the HAVE_STRNDUP check failing? 
- Why is it necessary to remove the inline definition?

araspik:
@nefthy On my machine `HAVE_STRDUP` is not set. So in `xmalloc.h` it uses the declaration (not inline definition) of `xstrndup`, expecting `xmalloc.c` to provide it. But then, when I `objdump -d xmalloc.o`, only `malloc_fail` is declared. So when linking together `cmus`, `ape.o` reports a missing `xstrndup` function.

nefthy:
I think that something else is wrong here. If HAVE_STRNDUP is not defined, then the function should be compiled from xmalloc.c.

- Do you use any parameters to `./configure`?
- What does `./configure | grep 'strn\?dup'` output?
- What is the content of the file `config/xmalloc.h` after running ./config?
- What system are you on?

nefthy:
-What compiler are you using?

araspik:
@nefthy Really sorry, it works without issues now. I have no idea why. I'm just closing this. Sorry about time wasted. No idea what just happened. I'm looking into it, but it's a ghost error.

nefthy:
No worries

sawaYch:
I have the same problem.

Building environment: 
KDE Neon (Ubuntu 18.04 lts)
GNU Make 4.1
gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0

./configure | grep 'strn\?dup' 
return:
```bash
checking for function strdup... yes
checking for function strndup... yes
```

After ./configure, config/xmalloc.h :
```cpp
#ifndef CONFIG_XMALLOC_H
#define CONFIG_XMALLOC_H

#define HAVE_STRDUP 1
#define HAVE_STRNDUP 1

#endif
```

I tried to use  @araspik araspik patch, it works and now I can build cmus.

PR #871, opened 2019-02-16
Ensures all data sent on dbus is in utf8 (#847)

amangrewal:
Partly addresses 847. All bytes that are invalid utf8 are converted
to <xx>, just like other areas in cmus. This is a temporary fix.

### DIFF
--- a/mpris.c
+++ b/mpris.c
@@ -25,6 +25,7 @@
 #include "output.h"
 #include "track_info.h"
 #include "utils.h"
+#include "uchar.h"
 
 #define CK(v) \
 do { \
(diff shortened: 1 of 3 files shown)


amangrewal:
Sorry for creating a new pull request. I couldn't figure out how to get github to use the old request after squashing the commits.

I wasn't really sure if this was the approach you had in mind. Once the conditional is abstracted into a function, the buffer has to be allocated in either case. I didn't see any value in adding something like
```
static void ensure_utf8( dst, src)
{
    if (u_is_valid(src))
        strcpy(src, dst);
    else
        u_to_utf8(dst, src);
}
```

geek1011:
@amangrewal you would need to force-push to the branch from the command line.
@nefthy, can you review this?

nefthy:
@flyingmutant LGTM

Issue #742, opened 2017-11-13
rc rules don't override autosave


rlue:
From the manpage:

```
$XDG_CONFIG_HOME/cmus/rc
      Static config file. This file is read immediately after the autosave file, and is
      never modified by cmus.  You can override auto-saved settings in this file.  This
      file is not limited to options, it can contain other commands too.
```

However, if cmus finds duplicate or conflicting settings in `rc` and `autosave` files (_e.g.,_ `bind common q quit` defined in both), it raises the following error on launch:

```
cmus: Error: key q already bound in context common
cmus: Press <enter> to continue.
```

---

1) `uname -a`:
```
Darwin liberte 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64
```

2) `cmus --version`:
```
cmus v2.7.1
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2013 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/local/Cellar/cmus/2.7.1_2/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  aac:
    Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  wav:
    Priority: 50
    File Types: wav
    MIME Types:

Output Plugins: /usr/local/Cellar/cmus/2.7.1_2/lib/cmus/op
```

ForTheReallys:
Use `bind -f`

rlue:
Great! What do the maintainers think about either mentioning this in the manpage or applying `-f` by default to all `bind` and `unbind` directives found in the `rc` file?

geek1011:
@nefthy @flyingmutant would you rather have this added to the documentation under the config file, added to the documentation under `bind`, or default to `-f` when loading from `rc`?

nefthy:
[...]

Issue #732, opened 2017-10-11
Optional Autosave?


ForTheReallys:
I ran into a small probblem the other day with the rc file and keybindings.  When I'd bind a key in my rc file and then remove the line entirely, the binding remained.  I found the problem was the autosave file.  I like the autosave file for the most part but is it necessary to save everything?  If so, can it be optional?  Should there be another option to autosave on exit?There are tons of different ways to do this and I can implement any of these but would like some input on how it should work and if it's even a big enough deal.

Thanks!

PR #874, opened 2019-02-19
Add an option to disable autosaving

Fabian-G:
I (and probably some other people) prefer to edit the configuration file by hand. Especially when you try out some theming settings you usually don't want this to be immediately rewritten to your config file.

This PR adds an option to disable cmus autosave feature. This also solves the need for `bind -f` in your
cmus rc file (#742 ) as one could simply put `set autosave=false` in the there and create an empty autosave file to prevent reading of `/usr/share/cmus/rc`

There was once an issue opened for this feature (#732 ), but it was closed by OP for some reason.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -809,6 +809,10 @@ aaa_mode (all) [all, artist, album]
 	like there were only the files of the currently playing artist in the
 	library.
 
+autosave (true)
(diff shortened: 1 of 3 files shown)


Fabian-G:
Nevermind. I just noticed that overwriting every option in the cmus rc file in conjunction with `bind -f` does the trick in my use case.

Issue #605, opened 2017-01-12
Cmus becomes unresponsive when playing tracks in cue files with certain character (0x85)


karlek:
Sorry for the ridiculously long issue name.

The cue file contains a Next Line character\[1\] and when the track with that name starts playing, cmus becomes unresponsive. Track name appears normal except a <85> in it's name.

[1]: http://www.fileformat.info/info/unicode/char/85/index.htm

Cue file: http://karlek.io/a.cue

1) `uname -a`:
```
Linux elysium 4.5.4-1-ARCH #1 SMP PREEMPT Wed May 11 22:21:28 CEST 2016 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0-rc0
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  aac:
    Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  ffmpeg:
    Priority: 30
    File Types: aa aac ac3 aif aifc aiff ape au fla flac m4a m4b mka mkv mp+ mp2 mp3 mp4 mpc mpp ogg shn tak tta wav webm wma wv
    MIME Types:

Output Plugins: /usr/lib/cmus/op
  pulse
  alsa
  oss
  jack
  ao
```

mahkoh:
cmus only supports cue files stored as utf8.

[...]

PR #876, opened 2019-02-22
Store metadata internally in utf8 [WIP]

amangrewal:
Here's a first draft. It works for me, but I don't have an easy way of inserting invalid utf8 characters into metadata yet.

TODO

- More testing

Here's a list of open issues that this pull request *should* close:
#894, #885, #847, #605.

### DIFF
--- a/uchar.c
+++ b/uchar.c
@@ -520,16 +520,6 @@ int u_to_ascii(char *dst, const char *src, int len)
 	return i;
 }
 
-void u_to_utf8(char *dst, const char *src)
-{
-	int s = 0, d = 0;
-	uchar u;
-	do {
-		u = u_get_char(src, &s);
-		u_set_char(dst, &d, u);
-	} while (u!=0);
-}
-
 int u_skip_chars(const char *str, int *width)
 {
 	int w = *width;
--- a/uchar.c
+++ b/uchar.c
@@ -135,6 +135,70 @@ int u_is_valid(const char *str)
 	return 1;
 }
 
+const char *ensure_utf8(const char *src)
+{
+	const unsigned char *s = (const unsigned char *)src;
+	char *buffer;
+	int i = 0, is_valid = 1;
+	int strlen = 1; //Start at 1 to account for \0
+
+	if (src == NULL)
+		return NULL;
+
+	while (s[i]) {
+		unsigned char ch = s[i++];
+		int len = len_tab[ch];
+
+		if (len <= 0) {
+			is_valid = 0;
+			//If it's invalid unicode, it will be store as <xx>.
+			strlen += 4;
+			continue;
+		}
+
+		if (len > 1) {
+			/* len - 1 10xxxxxx bytes */
+			uchar u;
+			int c;
+
+			len--;
+			u = ch & first_byte_mask[len];
+			c = len;
+			do {
+				ch = s[i];
+				if (len_tab[ch] != 0) {
+					is_valid = 0;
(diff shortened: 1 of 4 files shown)


nefthy:
You could do something like this to produce invalid utf-8:
```
t=`echo öäü | iconv -f utf8 -t latin1`
id3tag -a$t broken_artist.mp3
```

uchar.c
```diff
@@ -520,14 +584,16 @@ int u_to_ascii(char *dst, const char *src, int len)
 	return i;
 }
 
-void u_to_utf8(char *dst, const char *src)
+const char *u_to_utf8(char *dst, const char *src)
```
nefthy: would you mind explaining this signature change? It is not obvious to me why it is needed.

geek1011: Looks like it is used to allow for chaining, which makes the code look a lot nicer.

amangrewal: Yeah, it's just for chaining. There's no need for it beyond aesthetics.

nefthy: I don't like this change, as it is purely aesthetic and chaining is not a widely used idiom in the cmus code base.

nefthy:
Hm, I just stumbled across this in the documentation:
```
id3_default_charset (ISO-8859-1)

Default character set to use for ID3v1 and broken ID3v2 tags.
NOTE: This is used only if the tag is not valid UTF-8.
```
How does this option interact with this PR?

amangrewal:
Good catch, I have to look into it.

From a cursory glance, the behavior depends on which input plugin is used. I'll provide an update when I have something more definite.

amangrewal:
I've updated the above script. I was having issues with id3tag and certain file types.

I used ```grep id3 *.c *.h``` in ip to get a list of all file types/plugins that should support id3. The list of plugins is aac, mad, mp4, nomad, and wavpack. Using ```cmus --plugins``` to get the list of file types gives .aac, .m4a, .m4b, .mp2, .mp3, .mp4, and .wv.

For the file types that id3tag works with (aac, mp2, mp3, .wv), any invalid utf8 is treated as ISO-8859-1. When tagging with id3tag, .aac, .mp2, and .mp3 treat valid utf8 as ISO-8859-1; .wv treats it as utf8. When tagging with ffmpeg, .mp3 and .wv treat it as utf8; .aac and .mp2 don't seem to be tagged by ffmpeg.

In other words, embedded metadata is a complete mess.

Things I still need to test:
flac/cue
http streams
building without iconv

nefthy:
I file with this id3 tag failes to parse. The file failed to parse before this patch, so I guess it's not a show stopper.

[meta.txt](https://github.com/cmus/cmus/files/3086419/meta.txt)

nefthy:
I tested today by I adding about every mp3 file I have on my system. Everithing seems fine. I'll do some more deliberate tests these days. 

We should get some more testers on board before we merge this, since it is a major change. Do you think the patch is ready for testing/merging or are there open todos?

amangrewal:
I tested the meta.txt file. It didn't cause a crash or a freeze, but perhaps I added the metadata differently.

I would like to have a few testers. There are no more open TODOs.

I couldn't manage to configure icecast properly so that cmus could play from it, so I wasn't able to test streams. Other than that, I've tested all the file types and configurations mentioned that I said I would test.

nefthy:
@amangrewal it does not crush, it's just not added correctly. The filename is displayed instead of the meta data.

nefthy:

PR #904, opened 2019-05-09
Fix :mark on playlist view so filter is used

torbiak:
The manpage says filter expressions can be used in views 2-4, and I don't see the utility in `:mark` inverting the selection in the playlist view.

### DIFF
--- a/pl.c
+++ b/pl.c
@@ -849,7 +849,7 @@ void pl_invert_marks(void)
 void pl_mark(char *arg)
 {
 	pl_tw_only("mark")
-		editable_invert_marks(&pl_visible->editable);
+		editable_mark(&pl_visible->editable, arg);
 }
 
 void pl_unmark(void)


nefthy:

hunkyburrito:
Would love to see this re-opened and merged. Still works completely fine when I rebased and tested it locally.

torbiak:
Reading my description of the problem again, I think I may have been too terse.

In the manpage it says that `:mark` can be given a filter expression on views 2-4. But currently in view 3 (ie. playlist), when you run `:mark` it instead inverts the selection and ignores any given filter expression. I don't get why that'd be useful, since there's already the `:invert` command. My change just calls the function to select using the given filter expression (`editable_mark()`), making `:mark` behaviour in the Playlist view consistent with the Sorted Library and Play Queue views.

pgaskin:
Makes sense at a glance, original implementation looks like a copy-paste error. I'll test it properly later.

pgaskin (approved):
Code looks good, matches the other views, works as expected.

This PR actually fixes a regression introduced in 8e03cf645a1a6756b9ee9963f348bf11b6cee573, which refactored the pl commands to use wrappers in pl.c, and made an error while copy-pasting the mark implementation.

PR #906, opened 2019-05-11
fixed playlist name searching

jwparkau:


pl.c
```diff
@@ -132,8 +132,12 @@ static int pl_search_matches(void *data, struct iter *iter, const char *text)
...
 	for (size_t i = 0; words[i]; i++) {
-		if (u_strcasestr_base(pl->name, words[i])) {
-			matched = 1;
+
+		/* set in the loop to deal with empty search string */
+		matched = 1;
+
+		if (!u_strcasestr_base(pl->name, words[i])) {
+			matched = 0;
 			break;
```
nefthy: I'm not sure about this. On the left the break terminates the for loop, when a match is found. On the right version the oposite is the case. Would you please explaine why that is?

jwparkau: Sorry, I should have been more clearer.

The issue is that when searching for a playlist, it should be ensured that all the search terms match, rather than at least one.

One of the implications of the above issue is that when you have a playlist called `default` (or any other name with `a` in it), a query such as `a thousand miles` would match the playlist, thus matching all the songs in it, which is useless.

nefthy:
LGTM

flyingmutant:
Thanks!

PR #916, opened 2019-06-23
Fix a crash when moving all tracks in a view

### LINKED ISSUE #599: window_set_sel: BUG: iter_is_empty(&win->top)
---

Reproduce -

1.  select playlist view (3)
2.  select default playlist
3.  tab into tracks
4.  press space and select each track (including the last one in the list)
5.  press "p"

result: window_set_sel: BUG: iter_is_empty(&win->top)

1) `uname -a`:
```
Linux 4.8.15-300.fc25.x86_64 #1 SMP Thu Dec 15 23:10:23 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0-rc0

```

3) `cmus --plugins`:
```
cmus --plugins
Input Plugins: /usr/local/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    MIME Types:
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  aac:
    Priority: 50
    File Types: aac
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  ffmpeg:
    Priority: 30
    MIME Types:

Output Plugins: /usr/local/lib/cmus/op
  pulse
  alsa
  oss
  jack
  ao

```

jetomit:
Marking all tracks in library/playlist/queue view and moving them using p/P would crash cmus. Fix this by updating window state *before* setting selected track after the move.

This fixes #599; note that moving all tracks is a no-op, but it’s still better not to crash.

### DIFF
--- a/editable.c
+++ b/editable.c
@@ -247,8 +247,8 @@ static void move_sel(struct editable *e, struct list_head *after)
 	editable_track_to_iter(e, to_simple_track(after->next), &iter);
 
 	if (editable_owns_shared(e)) {
-		window_set_sel(e->shared->win, &iter);
 		window_changed(e->shared->win);
+		window_set_sel(e->shared->win, &iter);
 	}
 }
 


editable.c
```diff
@@ -247,8 +247,8 @@ static void move_sel(struct editable *e, struct list_head *after)
 	editable_track_to_iter(e, to_simple_track(after->next), &iter);
 
 	if (editable_owns_shared(e)) {
-		window_set_sel(e->shared->win, &iter);
```
nefthy: Could you explain what exactly causes the crash when window_set_sel is called first?

jetomit: Immediate cause of the crash is line 350 in window.c, in function `window_set_sel` (used to set the selected row):

    BUG_ON(iter_is_empty(&win->top));

This asserts that the window must not be showing an empty list  (`win->top` points to the item in the window’s topmost row, and is empty if there are no items in the list). This variable was not updated correctly when all tracks in a list were moved, so the window appeared to contain an empty list when trying to set the selected row.

Moving tracks to another position in a list works by first removing all marked items and then reinserting them at the new position. For each removed item, `window_row_vanishes` is (correctly) called to update `win->top`. After adding all the items back to the list, `window_changed` should be called for the same reason.

Existing code, however, called `window_changed` only after `window_set_sel`. This means that `win->top` was empty (if all tracks were moved) during the call to `window_set_sel`, and the above assertion failed.

Calling `window_changed` first ensures that `top` points to an actual (first) row, and the check for non-empty list succeeds when setting selection. Compare this with functions `editable_sort` and `editable_rand`, where `window_changed` is also called before setting the selected row.

nefthy:
@flyingmutant LGTM

flyingmutant:
@jetomit thanks for the PR!

Issue #895, opened 2019-04-09
WMA freezes cmus


amangrewal:
cmus freezes when playing wma files. SIGKILL is required to stop cmus (SIGTERM doesn't work). ffplay can play the files correctly.

My best guess is that the issue is with the ffmpeg plugin. I used to be able to play wma without issue, and I recently updated ffmpeg.

Here's the backtrace from gdb
```
#0  0x00007ffff7bb8b48 in __memmove_sse2_unaligned_erms () from /usr/lib/libc.so.6 
#1  0x00007ffff494147b in avpriv_copy_bits () from /usr/lib/libavcodec.so.58   
#2  0x00007ffff4fcad59 in ?? () from /usr/lib/libavcodec.so.58 
#3  0x00007ffff4fcfdbd in ?? () from /usr/lib/libavcodec.so.58   
#4  0x00007ffff49daa60 in ?? () from /usr/lib/libavcodec.so.58 
#5  0x00007ffff49db768 in avcodec_send_packet () from /usr/lib/libavcodec.so.58   
#6  0x00007ffff6823ab5 in ffmpeg_fill_buffer (ic=0x5555558cb4c0, cc=0x5555558961c0, input=0x5555558c8520,  
    output=0x5555558c7090, swr=0x5555559fdac0) at ip/ffmpeg.c:338   
#7  0x00007ffff6823c85 in ffmpeg_read (ip_data=0x555555888538,     
    buffer=0x55555567e6b0 "Q(\035\032a'C\031\220&m\030\342&\214\030\017'{\030J&9\027e%\336\025\023%\254\025]%M\026\064%D\026\005$\357\024\027#v\023\320\"\250\022&\"", count=60480) at ip/ffmpeg.c:394 
#8  0x000055555558112b in ip_read (ip=0x555555888530,
    buffer=0x55555567e6b0 "Q(\035\032a'C\031\220&m\030\342&\214\030\017'{\030J&9\027e%\336\025\023%\254\025]%M\026\064%D\026\005$\357\024\027#v\023\320\"\250\022&\"", count=60480) at input.c:695 
#9  0x000055555558b4d0 in _prebuffer () at player.c:599
#10 0x000055555558cc05 in player_set_file (ti=0x7fffcc006180) at player.c:1171    
#11 0x0000555555574136 in run_command (buf=<optimized out>) at command_mode.c:2883
#12 0x000055555556a7a1 in u_getch () at ui_curses.c:2163   
[...]

Issue #812, opened 2018-07-25
APE playback freezes cmus


shoober420:
---

1) `uname -a`:
```
Linux _QzWxEc_-717 4.17.9-1-zen #1 ZEN SMP PREEMPT Sun Jul 22 20:23:46 UTC 2018 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0-rc0-90-g17bf542
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  aac:
    Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  ffmpeg:
    Priority: 30
    File Types: aa aac ac3 aif aifc aiff ape au fla flac m4a m4b mka mkv mp+ mp2 mp3 mp4 mpc mpp ogg shn tak tta wav webm wma wv
    MIME Types:

Output Plugins: /usr/lib/cmus/op
  pulse
  alsa
  oss
  jack
  ao
```

After the first track finishes from an APE file in the playlist, cmus locks up. FLAC playback works perfectly fine. This only happens with APE files. The APE cuesheet is being used to create the tracklist and play via the "Library view (1)".

ffmpeg 1:4.0.2-1 on Arch Linux

shoober420:
[...]

PR #931, opened 2019-08-27
ip/ffmpeg: decode multiple frames from one packet

Ashymad:
See here https://github.com/cmus/cmus/issues/812#issuecomment-495201322 for a breakdown of the problem.
This is mostly a hack, but fixes the problem as far as i know. APE files play without a problem (fixes https://github.com/cmus/cmus/issues/812). It should also fix https://github.com/cmus/cmus/issues/895, but I don't have any WMA files to test it.

### DIFF
--- a/ip/ffmpeg.c
+++ b/ip/ffmpeg.c
@@ -37,7 +37,7 @@
 #endif
 
 #ifndef AVCODEC_MAX_AUDIO_FRAME_SIZE
-#define AVCODEC_MAX_AUDIO_FRAME_SIZE 192000
+#define AVCODEC_MAX_AUDIO_FRAME_SIZE 1920000
 #endif
 
 struct ffmpeg_input {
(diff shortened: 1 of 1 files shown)


ip/ffmpeg.c
```diff
@@ -37,7 +37,7 @@
 #endif
 
 #ifndef AVCODEC_MAX_AUDIO_FRAME_SIZE
-#define AVCODEC_MAX_AUDIO_FRAME_SIZE 192000
+#define AVCODEC_MAX_AUDIO_FRAME_SIZE 1920000
```
Ashymad: To fit the additional frames I had to increase the size of this constant. I've increased it tenfold, but this obviously makes little sense. I don't have an idea how to approach this without changing too much code.

nefthy:
Thanks for your effort, but I don't think its wise to merge this PR as it is obviously a hacky workaround. A proper fix should adress the issue as described in the documentation linked in the mentioned comment.

Ashymad:
A proper fix would probably entail rewriting quite a bit of the plugin. Although this is hacky, it mostly adheres to documentation (the only real problems are the constant buffer size and lack of proper error handling) and only touches the code, that would have to be rewritten anyway (there is no error handling in the original code as well). I think merging this for now doesn't significantly decrease the code quality, whilst giving us a bit more time with a working plugin until someone willing to really fix it steps up.

nefthy:

Issue #930, opened 2019-08-23
Enable push command without arguments


dirdi:
I am submitting a feature request:

I would like to bind `;` (additional to `:`) to enter command mode. Therefore I tried to bind `;` to the push command without applying any arguments to it: `:bind common ; push`. But instead of entering command mode, an error was raised: `Error: not enough arguments`. I suspect this happens, because push expects at least one argument.

It would be convenient to allow calling `push` without arguments.

dirdi:
Thank you @mrfft for implementing this! Since your patch has been merged, I am closing this.

PR #937, opened 2019-11-03
Allow push to accept no arguments

mrffr:
Implements #930

Allows the push command to be used with no arguments. This results in an empty command line.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -518,6 +518,7 @@ push <text>
 		bind common w push filter artist=
 
 	Text can contain spaces and even trailing spaces will be honored.
+	Text can also be left blank.
 	This command can only be bound to a key but not used in the command
 	line directly.
 
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -512,13 +512,13 @@ left-view
 right-view
 	Go to view "to the right" of current one, e.g. view 3 -> view 4.
 
-push <text>
+push [text]
 	Enter command mode with the command line pre-set to text. Example:
 
 		bind common w push filter artist=
 
 	Text can contain spaces and even trailing spaces will be honored.
-	Text can also be left blank.
+	If no text is given it defaults to a blank command line.
 	This command can only be bound to a key but not used in the command
 	line directly.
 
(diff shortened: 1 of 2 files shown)


command_mode.c
```diff
@@ -1353,7 +1353,9 @@ static void cmd_view(char *arg)
 
 static void cmd_push(char *arg)
 {
-	cmdline_set_text(arg);
+	if (arg) {
+		cmdline_set_text(arg);
+	}
```
geek1011: To match the rest of the code, don't put braces around an if block with a single statement in it.

command_mode.c
```diff
@@ -2586,7 +2588,7 @@ struct command commands[] = {
 	{ "pl-export",             cmd_pl_export,        1, -1, NULL,                 0, 0          },
 	{ "pl-import",             cmd_pl_import,        0, -1, NULL,                 0, 0          },
 	{ "pl-rename",             cmd_pl_rename,        1, -1, NULL,                 0, 0          },
-	{ "push",                  cmd_push,             1, -1, expand_commands,      0, 0          },
+	{ "push",                  cmd_push,             0, -1, expand_commands,      0, 0          },
```
geek1011: You'll also want to update the [documentation](https://github.com/cmus/cmus/blob/b2921a24357299b3771da5eecfd68465e1280e97/Doc/cmus.txt#L515).

mrffr: Thanks will do.

Doc/cmus.txt
```diff
@@ -518,6 +518,7 @@ push <text>
 		bind common w push filter artist=
 
 	Text can contain spaces and even trailing spaces will be honored.
+	Text can also be left blank.
```
geek1011: You'll want to change `push <text>` to `push [text]` too.

mrffr: There's probably no need for that sentence then.

geek1011: In that case, you'll also want to update the description to "... command line pre-set to text **if provided**".

nefthy:
LGTM

flyingmutant:
Thanks! Merged.

PR #941, opened 2019-11-18
Fix segfault on ppc64 caused by narrowing cast

grahn:
The narrowing cast happens with the second argument in the call to
`u_set_char()` in `format_parse()` in format_print.c. The argument is
`str->len`.

Before this commit, `&str->len` was cast to `(int *)`, despite `str` being
a `struct gbuf` whose `len` member has type `size_t`.

This does not cause an issue on 64-bit little-endian systems because a
cast to `int *`, though narrowing, points to the bottom four bytes of
`str->len`, and so the incrementing of the local copy of the `idx`
parameter in `u_set_char()` and then the assignment back to `*idx` doesn't
have any undesired behavior.

64-bit little-endian
```
|0|1|2|3|4|5|6|7| (str->len, numbers indicate byte significance)
 ^
 |
 |_(size_t *)

|0|1|2|3|
 ^
 |_(int *)
```
An increment here on 64-bit little-endian increases the value held in
byte 0, which is the expected result, so narrowing has no detrimental
effect.

However the narrowing cast causes a bug on big-endian 64-bit
systems (such as ppc64) because a cast to `int *` points to the top
four bytes of `str->len`.

64-bit big-endian
```
|7|6|5|4|3|2|1|0| (str->len)
 ^
 |
 |_(size_t *)

|7|6|5|4|
 ^
 |_(int *)
```
Besides `u_set_char()` being passed the top 4 bytes when the bottom 4
bytes are desired, an increment of the local copy of `idx` in
`u_set_char()` and the assignment back to `*idx` increases the value held
in byte 4, when byte 0 should be incremented. So the assignment back
to `*idx` increases the value of `str->len` by 2^32, and thus the next
time `str->len` is used for array access a segfault occurs.

This commit fixes the bug by changing the `idx` parameter of
`u_set_char()` from type `int *` to type `size_t *`, and removing the
cast to `int *` in `format_parse()`.

This commit also updates all functions that use `u_set_char()` to pass
a `size_t *` argument in order to eliminate any warnings about
incompatible pointer types.

### DIFF
--- a/format_print.c
+++ b/format_print.c
@@ -216,7 +216,8 @@ static void print_str(const char *src)
 
 		}
 	} else {
-		int s = 0, d = 0;
+		int s = 0;
+		size_t d = 0;
 		uchar u;
 
 		while (1) {
@@ -471,7 +472,7 @@ static void format_parse(int str_width, const char *format, const struct format_
 		u = u_get_char(format, &s);
 		if (u != '%') {
 			gbuf_grow(str, 4);
-			u_set_char(str->buffer, (int *)&str->len, u);
+			u_set_char(str->buffer, &str->len, u);
 			(*len) += u_char_width(u);
 			continue;
 		}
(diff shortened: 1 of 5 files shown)


flyingmutant:
Thanks a lot, merged! Sorry for the huge delay, cmus is not actively maintained right now (see #856).

Traace:
Thanks :)

PR #972, opened 2020-04-30
Fix type of MPRIS2 fields title and album

mathieu-lemay:
The `title` and `album` mpris2 fields were sent as list of strings,
which caused issues with the GNOME integrated notification system.

The mpris2 specification states that these two fields should be strings,
not lists of strings.


pgaskin (approved):
Looks good; confirmed with the [spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata); tested with GNOME, playerctl, and py3status.

flyingmutant:
Thanks a lot!

PR #982, opened 2020-06-04
Add support for 'DISCNUMBER' REM-field in CUE-sheets

lucius-martius:
Tools like Exact Audio Copy use the "REM DISCNUMBER" comment as a convention to store the disc-number of a multi-disc album within the cue-sheet.

Other players that implement this convention are for example Foobar2000 or Clementine.

Currently in cmus track sorting wrong for such albums, alternating tracks with the same numbers from disc 1 and disc 2 etc.


nefthy (approved):

flyingmutant:
Thanks!

PR #986, opened 2020-06-10
Ignore false positive format truncation warnings

mokulus:
Ignore `-Wformat-truncation` where buffer would be truncated later to fit the screen anyway. This is gcc specific warning, so I check for gcc before, so it works with clang and other compilers. Now there are no warnings/errors when compiling with gcc.


nefthy (changes requested):
The way to get rid of the warnings is to use %.<maxlen>s

nefthy:

PR #992, opened 2020-08-11
[RFC] avoid redundant layer of buffer in coreaudio ao

tomty89:
NOT TESTED AT ALL.

I DO NOT have any Mac and was only doing it for fun. Not sure if it will work / make sense at all. Regardless, everyone is welcome to give it a try. NO WARRANTY ;-)

### DIFF
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -497,7 +497,8 @@ static int coreaudio_open(sample_format_t sf, const channel_position_t *channel_
 
 static void coreaudio_flush_buffer() {
 	pthread_mutex_lock(&mutex); // always succeed; wait until cond_wait unlock or right away
-	stopping = true; // deals with later callback(s)
+	if (coreaudio_buffer != NULL) // drop-ready; be careful
+		stopping = true; // deals with later callback(s)
 	// TODO: nullify untouched buffer
 	if (coreaudio_buffer != NULL && coreaudio_buffer_size > 0) {
 		memset(coreaudio_buffer, 0, coreaudio_buffer_size);
@@ -524,8 +525,6 @@ static int coreaudio_drop(void)
 {
 	/* coreaudio_buffer = NULL; */
 	/* coreaudio_flush_buffer(); */
-	/* coreaudio_buffer_size = 0; // always 0? */
-	/* stopping = false; */
 	return OP_ERROR_SUCCESS;
 }
 
@@ -534,10 +533,13 @@ static int coreaudio_write(const char *buf, int cnt)
 	memcpy(coreaudio_buffer, buf, cnt);
 	d_print("written to coreaudio: %d\n", cnt);
 	coreaudio_buffer_size -= cnt;
-	if (coreaudio_buffer_size == 0)
+	if (coreaudio_buffer_size == 0) {
+		pthread_mutex_lock(&mutex);
 		pthread_cond_signal(&cond);
-	else
+		pthread_mutex_unlock(&mutex);
+	} else {
 		coreaudio_buffer += cnt;
+	}
 	return cnt;
 }
 
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -45,6 +45,8 @@ static AudioStreamBasicDescription coreaudio_format_description;
 static AudioUnit coreaudio_audio_unit = NULL;
 static UInt32 coreaudio_buffer_size = 0;
 static char *coreaudio_buffer = NULL;
+static int write_cnt = 0;
+static char *write_buf = NULL;
 static UInt32 coreaudio_stereo_channels[2];
 static int coreaudio_mixer_pipe_in = 0;
 static int coreaudio_mixer_pipe_out = 0;
@@ -75,20 +77,20 @@ static OSStatus coreaudio_play_callback(void *user_data,
 	while (coreaudio_buffer_size > 0) {
 		if (stopping)
 			break;
-		if (coreaudio_write_size > 0) {
-			memcpy(buflist->mBuffers[0].mData, coreaudio_write, coreaudio_write_size);
-			coreaudio_buffer_size -= coreaudio_write_size;
-			while (coreaudio_write_size != 0)
+		if (write_cnt > 0) {
+			memcpy(buflist->mBuffers[0].mData, write_buf, write_cnt);
+			coreaudio_buffer_size -= write_cnt;
+			while (write_cnt != 0)
 				pthread_cond_signal(&cond);
 		}
 	}
-	d_print("coreaudio_buffer_size: %d\n");
+	d_print("coreaudio_buffer_size: %d\n", coreaudio_buffer_size);
 	if (//drop_buffer ||
 	    coreaudio_buffer_size == buflist->mBuffers[0].mDataByteSize) {
 		//drop_buffer = 0;
 		return kAudioUnitErr_NoConnection;
 	} else if (coreaudio_buffer_size > 0) {
-		memset(buflist->mBuffers[0].mData, 0, coreaudio_buffer,size);
+		memset(buflist->mBuffers[0].mData, 0, coreaudio_buffer_size);
 	}
 	return noErr;
 }
@@ -524,12 +526,14 @@ static int coreaudio_drop(void)
 
 static int coreaudio_write(const char *buf, int cnt)
 {
-	coreaudio_write = buf;
-	coreaudio_write_size = cnt;
+	d_print("cnt: %d\n", cnt);
+	write_buf = buf;
+	write_cnt = cnt;
+	// while(write_cnt != 0);
 	pthread_mutex_lock(&mutex);
-	pthread_cond_wait(&cond);
+	pthread_cond_wait(&cond, &mutex);
 	pthread_mutex_unlock(&mutex);
-	coreaudio_write_size = 0;
+	write_cnt = 0;
 	return cnt;
 }
 
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -52,6 +52,7 @@ static pthread_mutex_t mutex;
 static pthread_cond_t cond;
 static bool stopping = false;
 static bool partial = false;
+static bool hmm = false;
 
 static OSStatus coreaudio_device_volume_change_listener(AudioObjectID inObjectID,
 							UInt32 inNumberAddresses,
@@ -71,10 +72,11 @@ static OSStatus coreaudio_play_callback(void *user_data,
 {
 	pthread_mutex_lock(&mutex);
 	d_print("mDataByteSize: %d\n", buflist->mBuffers[0].mDataByteSize);
-	if (!stopping) {
+	if (hmm = !stopping) {
 		coreaudio_buffer = buflist->mBuffers[0].mData;
 		coreaudio_buffer_size = buflist->mBuffers[0].mDataByteSize;
 		pthread_cond_wait(&cond, &mutex);
+		hmm = false;
 	} else {
 		coreaudio_buffer = NULL; // post-flush stopping
 	}
@@ -510,7 +512,8 @@ static void coreaudio_flush_buffer() {
 	}
 	stopping = true; // debug
 	coreaudio_buffer_size = 0; // always before signal (i.e. next non-stopping callback)
-	pthread_cond_signal(&cond); // deals with current callback (if any)
+	while (hmm)
+		pthread_cond_signal(&cond); // deals with current callback (if any)
 	pthread_mutex_unlock(&mutex);
 }
 
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -82,7 +82,7 @@ static OSStatus coreaudio_play_callback(void *user_data,
 			stopping = false; // let flush() go
 			while (!stopping); // wait until it's ready for the next callback
 		}
-		d_print("stopping: %d\n");
+		d_print("stopping: %d\n", stopping);
 		d_print("unblocked\n");
 		pthread_mutex_unlock(&mutex);
 	} else {
@@ -506,7 +506,7 @@ static int coreaudio_open(sample_format_t sf, const channel_position_t *channel_
 	return OP_ERROR_SUCCESS;
 }
 
-static void coreaudio_flush_buffer() {
+static int coreaudio_flush_buffer() {
 	int ret;
 	stopping = true; // signifies stopping
 
@@ -518,13 +518,16 @@ static void coreaudio_flush_buffer() {
 	} else {
 		pthread_mutex_unlock(&mutex);
 	}
-	d_print("flushed: %d\n", ret);
+	return ret;
 }
 
 static int coreaudio_close(void)
 {
+	int flushed;
+
 	AudioOutputUnitStop(coreaudio_audio_unit);
-	coreaudio_flush_buffer();
+	flushed = coreaudio_flush_buffer();
+	d_print("flushed: %d\n", flushed);
 
 	AudioUnitUninitialize(coreaudio_audio_unit);
 	pthread_cond_destroy(&cond);
@@ -705,8 +708,11 @@ static int coreaudio_mixer_get_fds(int *fds)
 
 static int coreaudio_pause(void)
 {
+	int flushed;
+
 	OSStatus err = AudioOutputUnitStop(coreaudio_audio_unit);
-	coreaudio_flush_buffer();
+	flushed = coreaudio_flush_buffer();
+	d_print("flushed: %d\n", flushed);
 	if (err != noErr) {
 		errno = ENODEV;
 		return -OP_ERROR_ERRNO;
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -495,6 +495,7 @@ static int coreaudio_close(void)
 {
 	AudioOutputUnitStop(coreaudio_audio_unit);
 
+	coreaudio_buffer_size = 0;
 	coreaudio_buffer = NULL;
 	pthread_cond_signal(&cond);
 
@@ -515,7 +516,7 @@ static int coreaudio_write(const char *buf, int cnt)
 {
 	if (coreaudio_buffer == NULL) { // this should never happen?
 		d_print("unexpected; race?\n");
-		cnt = coreaudio_buffer_size = 0;
+		cnt = 0;
 	} else {
 		memcpy(coreaudio_buffer, buf, cnt);
 		coreaudio_buffer_size -= cnt;
@@ -680,6 +681,7 @@ static int coreaudio_pause(void)
 {
 	OSStatus err = AudioOutputUnitStop(coreaudio_audio_unit);
 
+	coreaudio_buffer_size = 0;
 	coreaudio_buffer = NULL;
 	pthread_cond_signal(&cond);
 
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -51,7 +51,6 @@ static int coreaudio_mixer_pipe_out = 0;
 static pthread_mutex_t mutex;
 static pthread_cond_t cond;
 static bool stopping = false;
-static bool locked = false;
 
 static OSStatus coreaudio_device_volume_change_listener(AudioObjectID inObjectID,
 							UInt32 inNumberAddresses,
@@ -513,15 +512,15 @@ static int coreaudio_open(sample_format_t sf, const channel_position_t *channel_
 }
 
 static void coreaudio_flush_buffer() {
+	stopping = true; // signifies stopping
 	if (pthread_mutex_trylock(&mutex)) { // callback locked
-		stopping = true; // signifies stopping
 		while (stopping) // wait for unblocked signal
 			pthread_cond_signal(&cond);
 		// callback unlocked
 		d_print("toggling\n");
 		stopping = true; // singal received; toggle back
 	} else {
-		locked = true; // main locked
+		pthread_mutex_unlock(&mutex);
 	}
 }
 
@@ -532,10 +531,6 @@ static int coreaudio_close(void)
 
 	AudioUnitUninitialize(coreaudio_audio_unit);
 	pthread_cond_destroy(&cond);
-	if (locked) {
-		locked = false;
-		pthread_mutex_unlock(&mutex);
-	}
 	pthread_mutex_destroy(&mutex);
 	stopping = false;
 
@@ -730,10 +725,6 @@ static int coreaudio_pause(void)
 
 static int coreaudio_unpause(void)
 {
-	if (locked) {
-		locked = false;
-		pthread_mutex_unlock(&mutex);
-	}
 	stopping = false;
 	d_print("unpausing\n");
 	OSStatus err = AudioOutputUnitStart(coreaudio_audio_unit);
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -88,9 +88,9 @@ static OSStatus coreaudio_play_callback(void *user_data,
 	if (locked) {
 		coreaudio_buffer = buflist->mBuffers[0].mData;
 		coreaudio_buffer_size = buflist->mBuffers[0].mDataByteSize;
-		coreaudio_buffer_size_delay = coreaudio_format_description.mSampleRate /
+		coreaudio_buffer_size_delay = coreaudio_format_description.mSampleRate / 2 /
 		        nframes; // inverse proportion
-			// nframes - nframes / 20;
+			// (nframes - nframes / 20); // inverse proportion
 		/* blocking = true; */
 		pthread_cond_wait(&cond, &mutex);
 		/* blocking = false; */
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -493,11 +493,11 @@ static int coreaudio_open(sample_format_t sf, const channel_position_t *channel_
 }
 
 static void coreaudio_flush_buffer() {
-	if (coreaudio_buffer_size > 0 && coreaudio_partial_buffer) {
+	if (coreaudio_partial_buffer) {
 		memset(coreaudio_buffer, 0, coreaudio_buffer_size);
 		d_print("drain buffer with %d bytes of zeroes padded\n", coreaudio_buffer_size);
 	} else {
-		d_print("buffer dropped: %d\n", coreaudio_partial_buffer);
+		d_print("buffer dropped: %d\n", -1 * coreaudio_buffer_size);
 		coreaudio_buffer = NULL;
 	}
 	coreaudio_buffer_size = 0;
@@ -519,7 +519,7 @@ static int coreaudio_close(void)
 
 static int coreaudio_drop(void)
 {
-	coreaudio_buffer_size = 0;
+	coreaudio_partial_buffer = false;
 	coreaudio_flush_buffer();
 	return OP_ERROR_SUCCESS;
 }
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -76,16 +76,16 @@ static OSStatus coreaudio_play_callback(void *user_data,
 		pthread_cond_wait(&cond, &mutex);
 		d_print("unblocked\n");
 		pthread_mutex_unlock(&mutex);
-	} else {
-		if (stopping)
-			d_print("stopping\n");
-		else
-			d_print("main locked\n");
-		return kAudioUnitErr_NoConnection;
-	}
+	}/*  else { */
+	/* 	if (stopping) */
+	/* 		d_print("stopping\n"); */
+	/* 	else */
+	/* 		d_print("main locked\n"); */
+	/* 	return kAudioUnitErr_NoConnection; */
+	/* } */
 
-	if (coreaudio_buffer_size == buflist->mBuffers[0].mDataByteSize)
-		coreaudio_buffer = NULL;
+	/* if (coreaudio_buffer_size == buflist->mBuffers[0].mDataByteSize) */
+	/* 	coreaudio_buffer = NULL; */
 	if (coreaudio_buffer != NULL && coreaudio_buffer_size > 0)
 		memset(coreaudio_buffer, 0, coreaudio_buffer_size);
 	coreaudio_buffer_size = 0; // this must be ensured before we let flush() go
@@ -97,9 +97,9 @@ static OSStatus coreaudio_play_callback(void *user_data,
 		d_print("toggled\n");
 	}
 
-	if (coreaudio_buffer == NULL)
-		return kAudioUnitErr_NoConnection;
-	else
+	/* if (coreaudio_buffer == NULL) */
+	/* 	return kAudioUnitErr_NoConnection; */
+	/* else */
 		return noErr;
 }
 
--- a/op/coreaudio.c
+++ b/op/coreaudio.c
@@ -74,33 +74,35 @@ static OSStatus coreaudio_play_callback(void *user_data,
 	//stopping set
 	//signal until (!callback_locked): exit right away but we have stopping set
 	callback_locked = !pthread_mutex_lock(&mutex);
-	d_print("callback_locked: %d\n", callback_locked);
-	if (callback_locked) {
-		//stopping set
-		//signal until (!callback_locked): stops when callback unlocked
-		if (stopping) {
-			callback_locked = !!pthread_mutex_unlock(&mutex);
-		} else {
-			// stopping set
-			// signal until (!callback_locked): stops when callback unlocked
-			coreaudio_buffer = buflist->mBuffers[0].mData;
-			coreaudio_buffer_size = buflist->mBuffers[0].mDataByteSize;
-			pthread_cond_wait(&cond, &mutex);
-			// stopping set
-			// signal until (!callback_locked): stops when callback unlocked
-			if (dropping) {
-				dropping = false;
-			} else if (coreaudio_buffer_size == 0) {
-				ret = false;
-			} else if (coreaudio_buffer_size < buflist->mBuffers[0].mDataByteSize) {
-				memset(coreaudio_buffer, 0, coreaudio_buffer_size);
-				coreaudio_buffer_size = 0;
-				ret = false;
-			}
-			callback_locked = !!pthread_mutex_unlock(&mutex);
+	//stopping set
+	//signal until (!callback_locked): wait until unlocked if callback locked
+	d_print("pre-wait: %d\n", callback_locked);
+	if (callback_locked && !stopping) {
+		// stopping set
+		// signal until (!callback_locked): stops when callback unlocked
+		coreaudio_buffer = buflist->mBuffers[0].mData;
+		coreaudio_buffer_size = buflist->mBuffers[0].mDataByteSize;
+		pthread_cond_wait(&cond, &mutex);
+		// stopping set
+		// signal until (!callback_locked): stops when callback unlocked
+		if (dropping) {
+			dropping = false;
+		} else if (coreaudio_buffer_size == 0) {
+			ret = false;
+		} else if (coreaudio_buffer_size < buflist->mBuffers[0].mDataByteSize) {
+			memset(coreaudio_buffer, 0, coreaudio_buffer_size);
+			coreaudio_buffer_size = 0;
+			ret = false;
 		}
+		// do not unlock here; could be stopping
 	}
-	d_print("callback_locked: %d\n", callback_locked);
+	// stopping set
+	// signal until (!callback_locked): either lock failed or stops when callback unlocked
+
+	d_print("pre-unlock: %d\n", callback_locked);
+	if (callback_locked) // includes stopping
+		callback_locked = !!pthread_mutex_unlock(&mutex);
+	d_print("post-unlock: %d\n", callback_locked);
 
 	d_print("stopping: %d\n", stopping);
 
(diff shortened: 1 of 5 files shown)


op/coreaudio.c
```diff
@@ -699,27 +495,32 @@ static int coreaudio_open(sample_format_t sf, const channel_position_t *channel_
...
-	coreaudio_ring_buffer_flush(&coreaudio_ring_buffer);
+	coreaudio_buffer_rpos = NULL;
 	return OP_ERROR_SUCCESS;
 }
 
 static int coreaudio_write(const char *buf, int cnt)
 {
-	return coreaudio_ring_buffer_write(&coreaudio_ring_buffer, buf, cnt);
+	while (coreaudio_buffer_rpos != NULL) {
+		;
```
tomty89, line 517: Perhaps it could simply return 0 in this case.

op/coreaudio.c
```diff
@@ -267,10 +65,14 @@ static OSStatus coreaudio_play_callback(void *user_data,
 					AudioBufferList *buflist)
 {
 	int count = nframes * coreaudio_format_description.mBytesPerFrame;
```
tomty89, line 67: Would this always be equal to `buflist->mBuffers[0].mDataByteSize`? Maybe we can use that directly when `memcpy()`?

op/coreaudio.c
```diff
@@ -267,10 +65,14 @@ static OSStatus coreaudio_play_callback(void *user_data,
...
-	buflist->mBuffers[0].mDataByteSize = coreaudio_ring_buffer_read(&coreaudio_ring_buffer,
```
tomty89, line 270: Are we supposed to set/change it?

lochlainn:
Cheers, I'll put this through a round of testing today!

lochlainn:
Just some simple testing reveal the same issues I was having before. I'll generate some example FLAC files this week to test more granularly.

tomty89:
> Just some simple testing reveal the same issues I was having before. I'll generate some example FLAC files this week to test more granularly.

With or without this PR? (I'd be glad even if it only works with audio of certain specs though, as that probably means what I removed was indeed junk.)

Would be great if you can test the same set of files using both cmus with and without this PR.

tomty89:
@lochlainn Btw, **have you been testing** with git master instead of 2.8.0 (if you weren't using my fork)? See https://github.com/cmus/cmus/pull/870/commits

britisch:
Hi, I have the same issue with flac files at 192khz (tested with 3 different files) on a mac. 96khz works just fine. hardware is a topping d10s, works fine with other software (e.g., ncmpcpp). compiled from the master branch. anything I can be of help with?

pgaskin:
@britisch, try compiling this branch and see if it solves your issue.

britisch:
done so. cmus does start, however no sound at all. playing any song stalls the program (it shows the duration on the bottom, e.g., -5:14), but then it does not play the song, no sound. when trying to quit, after q I need to wait for some seconds before I can confirm with y. then the screen closes, but the prompt does not appear in terminal. need to manually kill cmus.

I have tried all commits (first compilable is f82a220), all the same problem.

How can I help next? BTW, master compiles and runs without problems (well, beside the noise problem with high sample rate files ...)

tomty89:
@britisch Can you try commit 91bbcb7?

tomty89:
@britisch Can you try commit 6e51734?

britisch:
Thanks! I compiled latest commit. However, still the same result, cmus does not play anything. During compilation I receive the following warning message:

`➜  cmus git:(pr/tomty89/992) make install
   CC     op/coreaudio.lo
op/coreaudio.c:502:9: warning: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        memcpy(coreaudio_buffer, buf, cnt);
               ^~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/secure/_string.h:63:27: note: expanded from macro 'memcpy'
                __builtin___memcpy_chk (dest, __VA_ARGS__, __darwin_obsz0 (dest))
                                        ^~~~
1 warning generated.
   LD     op/coreaudio.so
`

tomty89:
@britisch Not exactly sure why it didn't work. Tried one more thing, you may give it a spin if you wish. (I squashed back to cf28145 btw)

britisch:
@tomty89, thanks for your efforts to make this work! unfortunately the latest version (rolled back and applied one commit) gives a buffer error (regardless of the file played):
`[1]    2203 bus error`

tomty89:
@britisch Please try a7db856. Thanks.

britisch:
Thanks! tested @28b6ef6, does not crash, but does not play either (same as before yesterday). so sorry to say - and thanks for your continued tries!

tomty89:
@britisch I added some logging to the code. Can you try to get a sequence of logs? Might help me to see why it doesn't play.

britisch:
Thanks. were do I find the log output? maybe I just miss the point. is there a parameter I need to provide?

tomty89:
@britisch you need to run `configure` with `DEBUG=2`, then there will be a `cmus-debug.txt` in your home directory

britisch:
Unfortunately the file remains empty ...

tomty89:
Hmm, you mean the file is produced but it's empty? Or is it not produced at all?

britisch:
yes, like that. the file is created but empty.

tomty89:
@britisch No idea then. There should be logs in it even without my commit. Maybe yet another macOS-specific issue.

britisch:
no clue either. too bad. maybe at a later point in time somebody with a mac at hand can try. but thanks a ton for all your efforts!!!

Allavaz:
I'm a noob but I just compiled from this PR I think (I ran `gh pr checkout 992`) with `DEBUG=2` and sadly I'm still getting noises, but I did get a `cmus-debug.txt` file, and it's not empty:

```
main: charset = 'UTF-8'
player_init: using real-time scheduling
player_init: setting priority to 47
coreaudio_hog_device: Cannot get hog information: 560947818
player_set_op: setting op to 'coreaudio'
init_plugin: initialized mixer for coreaudio
init_plugin: initialized pcm for coreaudio
init_curses: Number of supported colors: 256
init_curses: ts: 0 fs: 0
TIMER: worker job:           0.003516
TIMER: worker job:           0.000028
TIMER: worker job:           0.012432
file_changed: file: /Volumes/HDD/Music/S. Kiyotaka & Omega Tribe/River's Island/1-01. River's Island.flac
update_rg_scale: gain = -4.440000, peak = 0.989839, db = -4.440000, scale = 0.599791, limit = 1.010265, replaygain_scale = 0.599791
metadata_cb: something else
metadata_cb: VORBISCOMMENT
metadata_cb: something else
tell_cb: 
flac_open: sr: 96000, ch: 2, bits: 24
ip_setup: pcm convert: scale=1 convert=0 convert_in_place=0
coreaudio_fill_format_description: Bits:24
```

I'm running macOS 10.15.7 if it's any help.

tomty89:
@Allavaz if it plays for you (even if it's noises) then perhaps you are not really compiling this PR, as the problem was playback does not start at all after certain changes according to @britisch. Also I'm not seeing the extra logs I should see.

tomty89:
@Allavaz Can you try to build from my fork directly? I just pushed the changes to its master branch as well.

Allavaz:
Alright, I just cloned your fork, compiled and installed it. When I try to play a song it doesn't play anything, it stays at 0:00 and the program becomes unresponsive. Here's the logs:

```
main: charset = 'UTF-8'
player_init: using real-time scheduling
player_init: setting priority to 47
player_set_op: setting op to ''
player_error: ERROR: 'selecting output plugin '': no such plugin'
init_curses: Number of supported colors: 256
init_curses: ts: 0 fs: 0
error_msg: Error: selecting output plugin '': no such plugin
TIMER: worker job:           0.003942
TIMER: worker job:           0.000028
TIMER: worker job:           0.012952
player_set_op: setting op to 'coreaudio'
init_plugin: initialized mixer for coreaudio
init_plugin: initialized pcm for coreaudio
file_changed: file: /Volumes/HDD/Music/S. Kiyotaka & Omega Tribe/River's Island/1-01. River's Island.flac
update_rg_scale: gain = -4.440000, peak = 0.989839, db = -4.440000, scale = 0.599791, limit = 1.010265, replaygain_scale = 0.599791
metadata_cb: something else
metadata_cb: VORBISCOMMENT
metadata_cb: something else
tell_cb: 
flac_open: sr: 96000, ch: 2, bits: 24
ip_setup: pcm convert: scale=1 convert=0 convert_in_place=0
coreaudio_fill_format_description: Bits:24
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_play_callback: nframes: 4096; buffer size: 24576
coreaudio_buffer_space: get buffer: 24576
coreaudio_write: written to coreaudio: 24576
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
(the same line repeats a hundred times)
coreaudio_buffer_space: get buffer: 0
coreaudio_buffer_space: get buffer: 0
sig_shutdown: sig_shutdown 1
```

Allavaz:
I just realized it becomes unresponsive after hitting pause `c`, but still doesn't play no matter what.

tomty89:
@Allavaz Hmm, the result is kind of expected to me, but I'm unsure why the callback thread seems to have stalled/stopped. Can you test again with the latest commit in addition?

Allavaz:
Just woke up and tested it, and when I tried to play a song it segfaulted:

`zsh: segmentation fault  ./cmus`

Here's [cmus-debug.txt](https://github.com/cmus/cmus/files/5836700/cmus-debug.txt)

tomty89:
@Allavaz just added some sleep to the loop. can you test again? thanks a lot!

Allavaz:
It segfaulted again. Just curious, does the log file reset every time you run the program? Now it has much fewer lines:

[cmus-debug.txt](https://github.com/cmus/cmus/files/5836894/cmus-debug.txt)

Allavaz:
Still segfaulting :/

[cmus-debug.txt](https://github.com/cmus/cmus/files/5837095/cmus-debug.txt)

Allavaz:
This one is with commit 14629b6

[cmus-debug.txt](https://github.com/cmus/cmus/files/5837107/cmus-debug.txt)

tomty89:
@Allavaz Looks like I have to fill the whole buffer. How about now?

Allavaz:
With these changes it now plays!!! Although it struggles to play continuously, like there's micropauses as if it couldn't keep up with the track or something. I tried setting softvol to false to see if it improved, and it didn't. Then I hit pause and it became unresponsive so I had to force close it.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5837204/cmus-debug.txt)

tomty89:
@Allavaz I switched to another approach (pthread mutex/cond). Would be great if you can give it a go. Thanks!

Allavaz:
We're getting closer! Now it plays perfectly, some hiccups here and there but very very rarely. Changing songs works, with some minimal hiccups here and there. Pausing worked the first time, but the second time it became unresponsive.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5837759/cmus-debug.txt)

tomty89:
Is the pausing issue fixed with 6dc473d?

Allavaz:
Just tried and nope. This time it froze on the first try.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5837789/cmus-debug.txt)

Allavaz:
Sorry, I thought I was on 6dc473d but I forgot to fetch & pull. But still, this time pausing worked only the first time.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5837846/cmus-debug.txt)

tomty89:
@Allavaz Added some more fixes and improvements. Please see how it works now. Thanks!

Allavaz:
Just tried, sadly it still hangs when pausing.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5839206/cmus-debug.txt)

tomty89:
@Allavaz I'm unsure what *could* happen when is AudioOutputUnitStop() is called, so I'll need your help to investigate further. Just added two commits that expects a waiting callback to be "killed"

Allavaz:
Sure! Here are the logs:

[cmus-debug.txt](https://github.com/cmus/cmus/files/5844131/cmus-debug.txt)

tomty89:
@Allavaz New fix is up. Please test. Thanks!

Allavaz:
I got all excited because I pressed pause and it worked!! But then I paused again and it froze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5846830/cmus-debug.txt)

tomty89:
Can you test again with the extra logging? Thank you so much!

tomty89:
I hope that I'm not missing any more race. Seems to have gotten a bit nasty. Wondering if I'm doing it right.

Thanks!

Allavaz:
Paused one time... it worked, paused again, it worked, paused a third time, it worked!!!

Then I hit next (`b`) and it segfaulted. But we're getting closer!

[cmus-debug.txt](https://github.com/cmus/cmus/files/5850042/cmus-debug.txt)

tomty89:
@Allavaz Please test. Thank you for your patience and help!

Allavaz:
It's freezing again when pausing :/

Allavaz:
Just tested eff4c22, pausing worked once.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853354/cmus-debug.txt)

tomty89:
@Allavaz The problem is strangely enough, only some of debug printing seem work work, and there is no obvious distinction between those that do and those that don't, so I'm still pretty much in blind.

I'm kind of out of idea now. Would you test and see if it works if no op_drop() is implemented?

Also, do you unpause very quickly after pausing? Does that seem to make it freeze more easily?

Allavaz:
I just ran it like 3 times. I waited like a minute before pausing, it worked, then I waited like 5 seconds before unpausing, waited 30 more seconds, paused again and it froze. Then I ran it again, paused 2 seconds into a song, it worked, tried again and it froze. The third time I ran it, I did the exact same thing (paused 2 seconds into the song) and it froze. I used to think that the more you let the song run, the more chances it won't freeze when you pause, but it really seems pretty random. 

It's weird because there was that one time when it let me pause 3 times (which is a record!), but then it segfaulted when I tried to change songs. When I was testing the commit before the last one, I seeked through a song and that also caused a segfault. Now it's not segfaulting, but sometimes I can pause once, sometimes I can't at all.

I'd also like to add that this pausing bug happens in any type of track, MP3, FLAC 16bit and so on.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853469/cmus-debug.txt)

tomty89:
@Allavaz Just noticed that I forgot to push the 1800d35 to the master branch. Were you compiling the coreaudio branch?

Well the thing is I can't even be sure the pausing "really" worked with 9856bc7. If you can (make sure you are compiling that commit), would be great if you can test it for some more time.

Allavaz:
Yeah I noticed that the master branch didn't have 1800d35, but the coreaudio one did so I switched to that one. 

Hopefully I did this right: `git checkout 9856bc7`

- Waited like 3 seconds, paused and it froze. Didn't even let me do it once.
- Ran it again, did the exact same thing, waited 3 seconds and hit pause, it worked. Waited like 2 seconds to unpause, then paused again and it froze.
- Third run, waited 3 seconds and hit pause, it froze.
- Fourth run, waited 30 seconds this time and hit pause, it froze.
- Fifth run, waited 30 seconds again, hit pause and it worked. Then I waited 15 seconds before unpausing. Paused again when the song was at 0:55 and it froze.
- Sixth run, waited 2 seconds, paused and it worked, waited 2 seconds, unpaused, paused again and it froze.

I'm extremely confused. I swear it let me pause 3 times, and I didn't even try more because I was confident it was fixed for good, then the segfault happened.

Don't know if logs are still useful at this point but here's the one from the last run.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853566/cmus-debug.txt)

tomty89:
Hmm just noticed that I did introduced a regression in 73bf69d. (Would have noticed it right away if all the loggings work...) Hopefully if your e.g. build script somehow always ignore/revert your checkout, 73bf69d did worked.

Please test the latest fix, thanks!

Allavaz:
Ran it 3 times, the 3 of them it froze on the first pause attempt. Then I ran it a couple more times and noticed that playing a 16bit flac song it let me pause like >6 times (and they were quick ones), it did eventually freeze though.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853729/cmus-debug.txt)

Allavaz:
On 563e5fb, I ran it and waited till ~1:00 and paused, it worked. Waited till ~2:00 and paused again, it worked. Then I paused at 2:20 and it froze. Every unpause was after ~1 sec of pausing. In earlier runs I waited like 5 seconds and it always died on the first pause attempt. I'm starting to believe that letting the track run for around a minute actually does help it somehow not to freeze when you pause, but I don't understand anything about buffers and stuff so I might be wrong.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853827/cmus-debug.txt)

tomty89:
@Allavaz Could you help test whether d0c4902 can give us more logs as well? Thanks!

Allavaz:
Paused at 0:45 and it froze, but here's the logs:

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853869/cmus-debug.txt)

Allavaz:
Logs for a551f87:

[cmus-debug.txt](https://github.com/cmus/cmus/files/5853931/cmus-debug.txt)

tomty89:
@Allavaz Hunted down more possible race. Please test, thanks!

Allavaz:
Just tested 1a0df61 - Pausing worked 2 times (on 0:10 and 0:20), then I tried at 0:30 and it froze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5856922/cmus-debug.txt)

tomty89:
@Allavaz Hmm, could you test c97a63 as well? Thanks!

Allavaz:
Paused at 0:10 and it froze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5857337/cmus-debug.txt)

Allavaz:
Stuttering returned in 216ec9a, but it happened just a couple times. It also froze when I paused at 0:45.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5857624/cmus-debug.txt)

Allavaz:
I was testing c9a7e63 again and it happened a couple times there too, so I guess it doesn't have to do with 216ec9a.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5857678/cmus-debug.txt)

tomty89:
@Allavaz Hi, could you help test 8bbd4df? Thanks!

tomty89:
@Allavaz Hi, could you help test 30f829c? Thanks!

tomty89:
@Allavaz Hi, could you help test 18055ba? Thanks!

Allavaz:
Paused at 0:30, it worked, then I waited around 10 seconds and paused again, and it froze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5860449/cmus-debug.txt)

Tested again, paused at 0:10 and it froze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5860454/cmus-debug.txt)

tomty89:
@Allavaz Allow me to propose yet another approach. Thanks a lot!

nefthy:
changed to draft. Please use the "Ready for review" button when you got this working.

tokyovigilante:
I'm not using macOS, however I have considerable experience programming using the CoreAudio API and am working on a Pipewire output plugin for Linux currently, so was interested in your approach. However after reading your code changes, it becomes clear that you are essentially locking around the CoreAudio device buffer in the realtime output thread, then writing to that buffer from cmus' buffer thread.

If that is correct, then it is probably not a good approach, as you are now blocking a realtime priority thread for an indeterminate length of time, which is explicitly recommended against by CoreAudio engineers, and in general a bad idea for realtime audio processing. [1-4]. It is perhaps not surprising that you are then having the indeterminate issues described above, which do sound like lock contention and priority inversion.

Given CoreAudio's scheduler model (and Pipewire's) is pull, rather than Pulseaudio's push model (cmus being set up for push) a ring buffer is the only real thread-safe option to not block the realtime thread. The buffering here is perhaps redundant but ultimately necessary.

The better change would be to have cmus decode straight to a ring buffer which could then read directly by CoreAudio, or passed to Pulseaudio and other push APIs in chunks, but that is likely a breaking output API change.

It's also unclear why this change would be expected to solve issues with high sample rate/bit depth files, which sound like buffer size or format negotiation issues.

1. https://lists.apple.com/archives/coreaudio-api/2009/Jun/msg00150.html
2. http://www.rossbencina.com/code/real-time-audio-programming-101-time-waits-for-nothing
3. https://www.mikeash.com/pyblog/why-coreaudio-is-hard.html
4. http://atastypixel.com/blog/four-common-mistakes-in-audio-development/

Allavaz:
@tomty89 sorry for the delay! I just tested d52ea70:

- Paused at 0:20, it worked.
- Waited 3 seconds to unpause, then I paused again at 0:30, and it froze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5863165/cmus-debug.txt)

tomty89:
@tokyovigilante I am aware of the model difference. Basically "to see how well/badly it could work by blocking/waiting instead" is exactly one of my intentions. This is just for my personal interest, I'm fine with it not merged if it is found to be inferior ultimately.

@Allavaz oops I forgot to remove some lines that contradicts with the latest approach. New fix is up. I hope you won't feel your time wasted btw.

Allavaz:
@tomty89 Don't worry :). Just tested ecde58a:

- Playback doesn't work now. I try to play a song and it stays at 0:00. 
- I can still navigate the menus, but only for ~5 seconds, then it completely freezes. 
- If I try to quit cmus (`q`), the UI disappears but it doesn't take me back to the shell, as if it was still running. I have then to force quit it like when it freezes.

Allavaz:
Just noticed that the UI freezing doesn't happen sometimes, it's weird. It still doesn't take me back to the shell when I quit though.

tomty89:
@Allavaz Looks like I do need to unlock mutex on both sides. Could you test 77a1883? Thanks!

Allavaz:
Now it plays. Small hiccups here and there though. 
- Paused at 0:30, it worked.
- Paused at 0:55, it worked.
- Paused at 1:00, it worked (!!!)
- Paused at 1:18, still works!!
- Paused at 1:20, no freezes so far!
- At this point I just started spamming `c`, like 6 times and it didn't freeze at all!
- Tried changing songs just to see if it segfaults like the other time, it didn't!
- Finally quit and it worked just fine.
So yeah, everything is working now (I think) but there's some micropauses going on, especially at the start of the songs, but it's minor, like a couple ones.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5869119/cmus-debug.txt)

tomty89:
@Allavaz Hmm, seems like the "real" reason that it froze was I have been stopping the audio unit before unblocking a waiting callback (if any). I just pushed one more commit to confirm it. Please try to test it as thoroughly as you did for the last commit. If it also works we can then try something to see if we can solve micropauses problem.

Btw, you mentioned that you had "noise" problem without this PR, was it also micropauses or do you mean like white noise / completely messed up audio? (In other words, have this PR prevented the noise problem? Assuming you were testing with the same file.)

Allavaz:
> Btw, you mentioned that you had "noise" problem without this PR, was it also micropauses or do you mean like white noise / completely messed up audio? (In other words, has this PR prevented the noise problem? Assuming you were testing with the same file.)

This PR definitely made 24 bit flacs playable (with micropauses, but still). Before these changes all I heard was extremely loud noises which the internal volume of cmus didn't seem to affect, I could set the volume to 0% and I'd still hear them. It was pretty dangerous to use. Thank god I wasn't using headphones when I first discovered this.

> I just pushed one more commit to confirm it. Please try to test it as thoroughly as you did for the last commit. If it also works we can then try something to see if we can solve micropauses problem.

Alright:

- Heard micropauses like before, just a couple, then paused at 0:30, no freezes.
- Paused at 0:45, no freezes.
- Paused at 0:50, no freezes.
- Paused at 0:51, no freezes.
- Started spamming `c` like the last time, no issues at all.
- Changed songs twice (just to spice it up a little bit), no freezes.
- Tried pausing again, still works.
- Finally quit and no problems at all.

Pretty much every time a song starts, at around 10 seconds it has these micropauses but again, this is huge progress. We're closer than ever I think.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5869308/cmus-debug.txt)

tomty89:
@Allavaz Thanks! Please provide log of b68a1e5. This time you only need to play one song until micropause happens for a couple times.

Allavaz:
- Happened at 0:59
- Happened at 0:29 (twice), then I quit

[cmus-debug.txt](https://github.com/cmus/cmus/files/5869518/cmus-debug.txt)

tomty89:
@Allavaz Hi, can you try if 37ce3b9 helps? Either way please provide logs. If possible, please also test with a file of 24/192 (if you don't have such file and don't know how to produce one, test additionally with file with lower sample rate can help as well). Thanks a lot!

Allavaz:
I rendered something I was working on at 24bit/192KHz and tried to play it. Didn't go well, lots of long pauses, like it was really struggling to play it. It was like trying to watch a 4k video on youtube on a dial up connection. 

[cmus-debug.txt](https://github.com/cmus/cmus/files/5870287/cmus-debug.txt)

This one's a 24bit/48KHz file, it went super smooth, no micropauses at all. Also I played this same song earlier today and heard 2 or 3 micropauses, so probably these changes helped.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5870321/cmus-debug.txt)

24bit/96KHz file here, same as before, no micropauses.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5870357/cmus-debug.txt)

Lastly, a 16bit/44.1KHz file. Same as before, no micropauses.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5870382/cmus-debug.txt)

So it definitely improved I'd say.

tokyovigilante:
> I am aware of the model difference. Basically "to see how well/badly it could work by blocking/waiting instead" is exactly one of my intentions. This is just for my personal interest, I'm fine with it not merged if it is found to be inferior ultimately.

That's fine, I'm just pointing out that most of the issues encountered here with your code are symptoms of locking in a realtime thread, and that the approach is generally not recommended in realtime audio given these exact issues are the result. 

Given that altering a buffering mechanism allows playback (however buggy) of previously unplayable high-bitrate media, it seems likely that the root cause is a buffer size or offset issue somewhere, rather than the use of an intermediary ring buffer being an issue in itself.

tomty89:
@tokyovigilante well, the hiccups/micropauses are actually due to the fact that, coreaudio will "pull" at varying rate (which is normal and expected), while the consumer_loop() of cmus checks buffer size at static rate (25ms sleep), so it either waits for too long or too short. (The rate can vary from <1ms to ~100ms, apparently; it might be partly because we set the "buffer frame size" to the largest possible reported; also regardless of the sample rate).

I end up decided to just bypass the sleep in the consumer loop and do cond wait in the ao's op_buffer_space().

> the root cause is a buffer size or offset issue somewhere
That somewhere is likely in the ring buffer implementation. I have no interest to investigate or fix anyway.

@Allavaz Can you test whether aef1343 works well? Thanks!

tokyovigilante:
> well, the hiccups/micropauses are actually due to the fact that, coreaudio will "pull" at varying rate (which is normal and expected), while the consumer_loop() of cmus checks buffer size at static rate (25ms sleep), so it either waits for too long or too short. (The wait span varies from <1ms to ~100ms, apparently; it might be partly because we set the "buffer frame size" to the largest possible reported; also regardless of the sample rate).

This is not really correct. 

CoreAudio will ultimately pull audio samples to match the output hardware rate. The number of samples requested each time the render callback is called is set in `coreaudio_start_audio_unit`. If you set a breakpoint in `coreaudio_play_callback` you will discover that `nframes` is the same each time. This is also very regularly timed, as it is being driven by a crystal oscillator in the sound hardware at exactly your audio sample rate, and the hardware must have exactly that many samples a second for glitch-free playback.

If you debug or log this somehow you will observe that for example a 512 frame buffer and 44.1kHz audio, you will see the callback called 44100/512 or just over 86 times a second. Typically the buffer can go up to 4096 frames (from memory) which would be ~10 callbacks/sec etc, so you are correct to say it varies depending on the selected frame size, but the time interval is then very regular.

The flip side of this is worth thinking about. CoreAudio is designed to minimise latency. If it requests x frames of audio, and it takes your callback longer than x/sample rate (512/44100 = 0.011 sec or 11msec) to provide (because it is locked, allocating memory, or some other indeterminate length operation), Core Audio will have nothing to play and you will have an audible glitch. The only safe thing to do in this callback is a memory copy, which is why the ring buffer is used.

cmus' input buffer function is ideally called every 25ms, but because `sleep` is implemented by `nanosleep` in BSD/Linux which can be interrupted by signals (and is subject to clock slew in recent macOS versions for battery life improvements), the length of time is indeterminate and can be longer than 25ms. This thread does not have realtime priority either. This is why your testers are experiencing these race conditions. It is also important to note they may be hardware and OS version specific, ie even if this approach works for you and for others, it may not work across all macOS users. 

In any event, any audible breaks in the audio, are only ever caused by your callback being late. This is why you should not be locking in your output callback, and this is why your users are experiencing bugs.

Allavaz:
@tomty89 just tested aef1343:

The 24bit/192KHz song plays fine, but the UI was completely unresponsive, the audio would still carry on just fine, but I couldn't do anything. After around 2 minutes of the song, it started working again, and all the commands/keys I pressed took effect.
At around 3:30 I tried pausing, when I pressed `c` the audio still continued but the UI froze, and after 2 seconds it finally paused and the UI became responsive again. I then unpaused, and paused again at ~3:45 I think, and that time it just didn't respond at all. The song still played till it was over, and the UI completely locked up, but this time it was forever. As I'm typing this, the song ended like 5 minutes ago and the UI still doesn't work, so I'll have to force close it.
But the good part is that there were no micropauses at all. Before these changes it was literally unplayable, you would hear like 50ms of the song, a 2 seconds pause, then some more 50ms and so on.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5875129/cmus-debug.txt)

Now I'm testing a 24bit/96KHz song and it's the same problem. The UI doesn't respond, the clock that shows the time of the song changes at a ~10 seconds rate, the keybindings respond with a delay of like 10 seconds too. I'm still gonna let the whole song run to see if there's any micropauses. Something I think is important to mention is that this time the UI never fully "unlocked", I'm on the last 10 seconds of the song and it's still lagging a lot. The song just ended and I think the UI doesn't respond at all now, same thing happened with the 192KHz song.

I lost the log file because I reopened cmus before it finished uploading so I'm playing the same song, and this time I noticed that if I don't try to change the volume, the UI works fine. Even `:` works fine, it's the volume (and pause) that makes it freeze.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5875192/cmus-debug.txt)

Finally, a 16bit/44.1KHz song. Tried pausing right away and the song carried on playing but the UI froze, then after 3 seconds it finally paused and the UI unfroze. It seems like the more times you try to change the volume, the longer it takes to respond. This time the song finished and the UI was working, maybe because I didn't try to change the volume a few seconds before it ended. So this time I didn't have to force close it.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5875217/cmus-debug.txt)

lochlainn:
I've been following this issue since I initially reported it a while ago. I'm back on my Mac, so I'll be able to test new patches as well.

Curious, how do musikcube/mpv/etc implement their coreaudio buffers? Do they use a ring buffer?

tokyovigilante:
mpv: [ringbuffer](https://github.com/MusicPlayerDaemon/MPD/blob/c72006dbccd70312670760e246f8ad7d72f11147/src/output/plugins/OSXOutputPlugin.cxx#L623)

musikcube: [Audio Queue Services API](https://github.com/clangen/musikcube/blob/edce1d7c8829906d573f2a9fb85c9e2db0285055/src/plugins/coreaudioout/CoreAudioOut.cpp#L89) - pool of AudioQueue buffers

mpv: [buffer, unknown implementation](https://github.com/mpv-player/mpv/blob/dcec6d9556c6cfca11b865442174479096f55111/audio/out/buffer.c#L188). Interestingly although not recommended as above, mpv does lock around it's buffer read call. Worth noting mpv has to synchronise with video and configurably does this by altering audio sample rate on the fly, so not a pure audio player. This might be the best example to review if you are insistent on locking around a buffer access.

libsoundio: [user callback](http://libsound.io/doc/structSoundIoOutStream.html#a9b6634417ccaec9b5c4376728efb186d), worth noting documentation also recommends against using potentially blocking functions.

tomty89:
@tokyovigilante 
> If you set a breakpoint in coreaudio_play_callback you will discover that nframes is the same each time.
That's false, regardless of whether you set "buffer frame size" (well, I haven't actually tested without setting the property, so at least "when it is set"). You can try it yourself. Even aaudio on Android have similar behavior. Also by rate I'm not (only) referring to nframes, but the rate that the callback is called (like the time span between each call can vary a lot).
It is probably a "self adjusting" mechanism to avoid buffer overflow or so.

Btw, the cmus core (the "producer loop") already uses a ring buffer. I'm only trying to avoid a second ring buffer in the ao when reading from it.

@Allavaz I'll see if I have missed anything, but it might not be any time soon. There's probably a reason I didn't want to wait in op_buffer_space()...

tomty89:
@Allavaz Just notice that I forgot to take the signaling in flush out for the op_buffer_space change. Please see if 995f574 helps. Thanks!
P.S. This design will probably make user interaction a little bit (not exactly sure how obvious it could be) less responsive anyway, as long as there could be big time gap between two callback calls. It would probably be more obvious when playing files of lower sample rate like 44.1/48KHz. In that case we could derive buffer frame size to use from sample rate (e.g. 40ms), instead of always using the probably hard-coded maximum in coreaudio (8192 frames).

tokyovigilante:
> @tomty89 

> That's false, regardless of whether you set "buffer frame size" (well, I haven't actually tested without setting the property, so at least "when it is set to the maximum allowed"). 
> NOT TESTED AT ALL.
> I DO NOT have any Mac

We may have to agree to disagree. The main point here is the output callback is called from a real time thread inside the Coreaudio daemon (which is why the thread is called at a constant rate with constant sample size) and locking in a real time thread is not deterministic behaviour. 

Note that macOS is not really a realtime OS so it is an approximation, but the hardware must be fed exactly the same number of frames per second as the sample rate otherwise you will have audio glitches. Your code cannot do this so it your testers have glitches. The lockups they report are probably due to priority inversion.

> Btw, the cmus core (the "producer loop") already uses a ring buffer.

This is why the correct approach would be to read directly from that ring buffer as I mentioned before. However that would require an output API change.

Anyway good luck with your patch.

tomty89:
> but the hardware must be fed exactly the same number of frames per second as the sample rate

Let alone whether this is ever true for any kind of audio hardware on any OS (you are talking about "second alignment"), the amount we feed coreaudio with (i.e. the amount it requests us for) at a time is pretty much never the same amount of frames it feeds the hardware with, because coreaudio has its own (ring) buffer as well (which is what we write to). That's why it would/could request at varying rate for varying amount according to the momentary fullness of its own buffer. (I'm not gonna describe the pattern, but if you are interested, it can be seen in some of the latter logs above.)

Allavaz:
@tomty89 Hi! Sorry for the long delay, I had to reinstall macOS, trying to update to Big Sur was a bad idea lol.

Just tested 995f574 on the 24bit/192KHz track. Paused at 1:00, it took like 1 or 2 seconds to actually pause though. Then at 2:00 I paused again and this time it took longer. Then after a few secs I tried changing the volume and again, a lot of delay. Seems like the more you interact with the program, the slower it becomes. This issue persisted when I played a 16bit/44.1KHz song.

[cmus-debug.txt](https://github.com/cmus/cmus/files/5927284/cmus-debug.txt)

lochlainn:
I got the chance to test this today. The 24-bit flac files I tested all played fine, pausing and resuming didn't cause the issues mentioned above, but I did notice that adjusting the volume (cmus softvol, not on macOS) caused the output to become distorted with lots of static.

I've attached a debug log from a run where I play a 24-bit flac song for 5 seconds, reduce the softvol to 90 (causing the static), then let that play for 5 seconds.

[cmus-debug.txt.gz](https://github.com/cmus/cmus/files/6157739/cmus-debug.txt.gz)

lxcode:
Tested today, playback works fine with softvol off. I did see the issues noted by @Allavaz: pausing takes a bit and the problem persists after the high bitrate files are done. That said, in its current state it's better than the status quo — which is risk of hearing damage — so I think it should probably be committed as is.

pgaskin:
@tomty89 are you still working on this?

pgaskin:
I just read through this PR and the code, and it seems that this PR has mixed results. Regardless, locking is probably not appropriate for this, and refactoring the output plugin API to be callback-based (#1328) is the most correct solution overall.

On a side note, I've started encountering buffer underruns with pipewire on Linux, which should also be fixed by refactoring the output plugin API to give the output direct control over buffering.

I'm going to close this for now; feel free to re-open it if you disagree.

PR #1009, opened 2020-12-04
Add MPRIS album art

mpostaire:
Extracts the album image of a song from its file to access it with MPRIS.
This extracts the images only once per album if an album is present in the tags.
Uses the cache to store the images.
If there is no image in the song file and there is a `cover.jpg` file in the same folder as the song file, it will use it as the song album art.

This also includes the fix introduced by this PR: #972


ip/mp4.c
```diff
@@ -26,6 +26,9 @@
 #endif
 #include "../comment.h"
 #include "aac.h"
+#include "misc.h"
```
Fernsicles, line 29: Shouldn't this be:
```
#include "../xstrjoin.h"
#include "../misc.h"
#include "../file.h"
```
?
I get include errors when compiling as is.

mpostaire: Yes you're right, I'll fix this.

pgaskin (commented):
Thanks for working on this, and sorry about the late review!

This seems to work in general, but I'm not sure I like the semantics of how this is implemented:

- Removals aren't handled.
- The cache size is unbounded (but isn't a severe issue).
- Directory/file handing doesn't feel right (see my comments below).
- The caching feels tacked-on.
- Albums/tracks with different art are not handled.
- I'm not completely sure if this belongs with track import.
- There should probably be an option to disable it.

One possible solution to these issues might be to store a hash of the image data in the library cache, and to extract the image on playback to a file named after that hash in the cache if it does not exist.

gavtroy:
> One possible solution to these issues might be to store a hash of the image data in the library cache, and to extract the image on playback to a file named after that hash in the cache if it does not exist.

Following that direction of thought maybe we could store the APIC / picture block offset durring metadata build and do the actual image extraction on-demand. Then we might be able to use a single file and avoid a cache altogether (which is 300 MB for me, with wildly varying cover sizes).

We need to be careful about the small amount of externally sourced utility code.
In the case of https://github.com/dnmfarrell/URI-Encode-C we could contain all the code in a separate directory and copy the original LICENSE text into our COPYING while clearly stating which directory it applies to. If that's too ugly we should write it from scratch.
For `b64_decode` from https://nachtimwald.com/2017/11/18/base64-encode-and-decode-in-c/ we need to find an explicitly licensed alternative and do something similar to record the license (or write from scratch).

pgaskin:
> Following that direction of thought maybe we could store the APIC / picture block offset during metadata build

That has a risk of an invalid image being used if the file is updated but not re-added. Also, I don't think all libraries used by the plugins support getting the specific offset for a metadata block.

But, as a variation of that, we could cache the last-known picture hash for a track instead, and also implement some sort of reference counting and/or cache-eviction algorithm on top of that. This would probably go in a separate art cache file rather than `lib.pl` for simplicity, stability, and compatibility.

> In the case of https://github.com/dnmfarrell/URI-Encode-C we could contain all the code in a separate directory and copy the original LICENSE text into our COPYING while clearly stating which directory it applies to. If that's too ugly we should write it from scratch.

I have an implementation of that I wrote a while ago; I'll just need to dig that up. IIRC, it's also quite a bit shorter than this one, and requires fewer allocations.

> For b64_decode from https://nachtimwald.com/2017/11/18/base64-encode-and-decode-in-c/ we need to find an explicitly licensed alternative and do something similar to record the license (or write from scratch).

That should probably be written from scratch (or found from a public-domain repository) since I don't think it's worth bringing in an additional license just for that.

mpostaire:
Thanks for the feedback! It was something I did quickly during summer break and I just focused on making a working code. I didn't actually think about the cache size and song removals but you're right it can be a big problem in large libraries :)

I decided to extract the art during the track import because it's the part of the code I felt comfortable working on and I worried of the cost of extracting an image and writing it to the disk. I thought it could introduce weird cuts between 2 songs if the art wasn't already present in the cache.

I initially thought about the different art per album but I saw that VLC uses only one album art per album even if some tracks have different art so I decided to do the same. Your idea of using a hash for the image data is a nice solution.

 I'm a student and I got a lot of work to do right now but as soon as I have some free time I'll look into it.

TorchedSammy:
hi, any updates on this? currently using this forked version instead so i can have album art in my notifications and other things

pgaskin:
@mpostaire, are you still working on this?

pgaskin:
I'm going to close this for now due to inactivity and since major changes are required:

- The helper code included for base64 and urlencode.
- It assumes that files are local in many places and is overly dependent on filenames.
- The caching needs to be redesigned to be usable.
- It ties the input plugins with core cmus logic too much in the current form (e.g., album art should not be written separately by each input plugin).

The core album art and mpris logic is still valuable as a base for future PRs, but the way the extraction and caching is done needs a major rework. Feel free to re-open this PR or start a new one based on this (I may also work on this if I have time/motivation later).

I agree that album art should be processed during import/update, but it should be optional and probably be cached in a file named after a hash of the image.

PR #1018, opened 2020-12-12
Add tree width options

### LINKED ISSUE #772: Is there a way to make playlist list smaller?
So, here is what it looks like right now:

As you can see the left part takes too much space. Is it configurable?  
If it's not then it would be a very nice addition.

### LINKED ISSUE #990: How to change treeview width
---
I don't find the option to change treeview width anywhere? How can i change treeview width (the list that is always show on the left)

1) `uname -a`:
```
5.7.10-arch1-1 #1 SMP PREEMPT x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0
```

pgaskin:
closes #990
closes #772

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1137,6 +1137,9 @@ status_display_program () [command]
 time_show_leading_zero (true)
 	Pad durations of less than 10 minutes with a leading 0
 
+tree_width_percent (33) [1-99]
+	Percentage of window to use for left pane.
+
 wrap_search (true)
 	Controls whether the search wraps around the end.
 
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1137,8 +1137,39 @@ status_display_program () [command]
 time_show_leading_zero (true)
 	Pad durations of less than 10 minutes with a leading 0
 
-tree_width_percent (33) [1-99]
+tree_width_percent (33) [1-100]
 	Percentage of window to use for left pane.
+
+tree_width_max (0) [0-9999]
+	Restrict the size calculated from tree_width_percent. 0 disables this
+	option.
+
+	This affects the tree in the library view and the playlist list in the
+	playlist view.
+
+	Note that this value will be further constrained by the window width
+	and/or the minimum tree size.
+
+	For example, to set a fixed size of 60 columns:
+
+		@pre
+		set tree_width_percent=100
+		set tree_width_max=60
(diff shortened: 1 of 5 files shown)


pgaskin:
A few notes:
- I'm not completely familiar with the UI code, so I'm not 100% sure this doesn't cause issues elsewhere, but it seems to work fine right now.
- Since I'm using an int between 1 and 100, we lose precision (it used to divide by 3), but this is mostly irrelevant since terminal windows aren't wide enough to make this have a significant effect. If this is an issue, I could parse it as a double or a fraction with a numerator/denominator instead.

pgaskin (commented):
Tested tree/library/playlist views, mouse events, resizing window, setting window size from rc/autosave/command-line/remote.

nefthy:
> * I'm not completely familiar with the UI code, so I'm not 100% sure this doesn't cause issues elsewhere, but it seems to work fine right now.
IMO its better to handle the width in columns since that is the basic underlying unit.

pgaskin:
What do you think of renaming it to `tree_width` and making it accept either a fixed value (later constrained to the max width of the window) or a percentage (if there's a percent sign at the end).

nefthy:
sounds good

pgaskin:
Actually, I think I have a better idea: adding an optional `tree_width_max` option to limit the maximum width, and increasing the maximum `tree_width_percent` to 100%. This will allow both: setting a fixed size and still making it work properly at smaller window sizes, limiting the size only on large window sizes, and keeping the old behaviour.

nefthy:
Also a good option.

pgaskin (commented):
Tested tree/library/playlist views, mouse events, resizing window, setting window size from rc/autosave/command-line/remote with:
- `tree_width_percent=100` `tree_width_max=60`
- `tree_width_percent=33` `tree_width_max=60`
- `tree_width_percent=33` `tree_width_max=0`
- `tree_width_percent=1` `tree_width_max=9999`

pgaskin:
@nefthy, this is ready for review now.

pgaskin:
I took a closer look at the code, and I'm now fairly certain this won't cause issues elsewhere.

nefthy (changes requested):
Looks good, but please rebase, as there is a conflict with another merge

nefthy (approved):

nefthy:
thanks

PR #1030, opened 2020-12-27
ip/cdio.c: only declare track_comments_found if HAVE_CDDB was defined

BouncyMaster:
Also change it's data type: int -> bool,

this variable is only needed when `HAVE_CDDB` is defined.


nefthy:
Getting rid of track_comments_found does not have any real advantage, while the #ifdefs complicate the code. I'm not in favor of this PR.

pgaskin:
Also, why char rather than bool?

BouncyMaster:
> Getting rid of track_comments_found does not have any real advantage, while the #ifdefs complicate the code. I'm not in favor of this PR.

Well the reason for this PR is to get rid of the warning when building without cddb, plus I don't think it adds **that** much complexity.


> Also, why char rather than bool?

You're right, will update the PR with the change.

nefthy (approved):

nefthy:
I think this can be merged.

pgaskin (approved):

PR #1014, opened 2020-12-12
Add gruvbox-warm.theme

pgaskin:
A warmer version of gruvbox.theme.

PR #1038, opened 2021-01-12
Add comments about fg/bg color to gruvbox-warm.theme

pgaskin:


pgaskin:
@flyingmutant, this PR only adds two comments to extend what I did in #1014.

PR #1025, opened 2020-12-15
Implement mouse events for title bar

pgaskin:
closes #1015

PR #1039, opened 2021-01-12
Increase help view keybinding width

pgaskin:
#1025 increased the length of the largest keybinding name.


pgaskin:
@flyingmutant, this is a trivial fix for what could be considered a regression (the misalignment of the help columns).

PR #1040, opened 2021-01-13
Fixes truncation warnings by replacing %s with %.s

nefthy:
Fixes gcc warnings about format trunctations

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -924,7 +924,7 @@ static void print_filter(struct window *win, int row, struct iter *iter)
 		e_filter = conv_buffer;
 	}
 
-	snprintf(buf, sizeof(buf), "%c%c%c%-15s  %s", ch1, ch2, ch3, e->name, e_filter);
+	snprintf(buf, sizeof(buf), "%c%c%c%-15s  %.235s", ch1, ch2, ch3, e->name, e_filter);
 	pos = format_str(print_buffer, buf, COLS - 1);
 	print_buffer[pos++] = ' ';
 	print_buffer[pos] = 0;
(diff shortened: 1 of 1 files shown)


nefthy:
@pgaskin can you review?

pgaskin (approved):
LGTM (build: [before](https://github.com/pgaskin/cmus/runs/1693832535?check_suite_focus=true#step:5:148), [after](https://github.com/pgaskin/cmus/runs/1697912602?check_suite_focus=true#step:5:148)), works ok.

nefthy:
thanks. any ideas on the comment above concerning #pragma?

pgaskin:
> thanks. any ideas on the comment above concerning #pragma?

That's due to the second parameter since GCC doesn't realize the max length 256 since `pretty` is defined in another file. To fix the warning, you can either:

- Change it to `%.256s`.
- Keep the current pragma.
- Increase the buffer size > 535.

Usually, you'd use the `alloc_size` attribute for this kind of thing, but it only works when the size is passed as a function argument (I guess one could be added and checked to ensure it's < 256, but that's a larger refactor).

ui_curses.c
```diff
@@ -1105,8 +1105,10 @@ static void update_editable_window(struct editable *e, const char *title, const
 			utf8_encode_to_buf(filename);
 			filename = conv_buffer;
 		}
-		snprintf(buf, sizeof(buf), "%s %s - %d tracks", title,
-				pretty(filename), e->nr_tracks);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-truncation"
+		snprintf(buf, sizeof(buf), "%s %s - %d tracks", title, pretty(filename), e->nr_tracks);
+#pragma GCC diagnostic pop
```
pgaskin, line 1111:
```suggestion
		snprintf(buf, sizeof(buf), "%s %.256s - %d tracks", title, pretty(filename), e->nr_tracks);
```

PR #1045, opened 2021-01-20
Add album shuffle mode

### LINKED ISSUE #641: Add option to shuffle album order.
Cmus currently has a shuffle feature that shuffles every track in the library. I'd like an option to shuffle the order of the albums, so the tracks belonging to each album are played in order but once the album ends it plays another album in the library.
The android music player 'Vanilla music' Has a feature identical to this, if I haven't described the request properly, then that is an example.

gavtroy:
...for the library. Playlists aren't flexible enough. If we wanted playlists to support album shuffle, we'd probably need to consider reimplementing them as secondary libraries.

Closes #641.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -654,8 +654,8 @@ source <filename>
 	Reads and executes commands from <filename>.
 
 toggle <option>
-	Toggles the value of a toggle-able option (all booleans and tristate
-	*aaa_mode*).
+	Toggles the value of a toggle-able option (all booleans and the options
+	*shuffle*, *aaa_mode*, and *replaygain*).
 
 tqueue [NUM]
 	Queues NUM (default 1) random tracks from the library. See also
@@ -1149,9 +1149,18 @@ show_playback_position (true)
 show_remaining_time (false)
 	Display remaining time instead of elapsed time.
 
-shuffle (false)
-	Play in shuffled order. Shuffle works in the library views (1-2) and
-	playlist view (3).
+shuffle (off) [off, tracks, albums]
+	@li *off*
+	Play all tracks in order. See also *lib_sort* and *pl_sort*.
+	@li *tracks*
+	Play all tracks in the library or playlist in a shuffled order.
+	@li *albums*
+	Play each library album to completion before shuffling to the first track
+	of another library album. In playlists this option falls back to *tracks*
+	behaviour.
+
+	Filters and *aaa_mode* can be used to limit the items available for play.
+	Note that *shuffle*=`albums` will have no effect with *aaa_mode*=`album`.
 
 skip_track_info (false)
 	Don't load metadata when adding tracks. Useful when using network file
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -579,6 +579,9 @@ rand
 refresh (*^L*)
 	Redraws the terminal window.
 
+reshuffle
+	Reshuffles the shuffle lists for both library and playlist views.
+
 run <command>
 	Runs a command for the marked tracks OR the selected one if none marked.
 
@@ -657,9 +660,6 @@ shell <command>
 
 showbind <context> <key>
 	Shows a key binding.
-
-shuffle
-	Reshuffles the shuffle lists for both library and playlist views.
 
 source <filename>
 	Reads and executes commands from <filename>.
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -201,6 +201,9 @@ x                       player-play
 z                       player-prev
 v                       player-stop
 
+B                       player-next-album
+Z                       player-prev-album
+
 ]                       vol +0 +1%
 [                       vol +1% +0
 +                       vol +10%
@@ -513,6 +516,10 @@ pl-rename <name>
 player-next (*b*)
 	Skips to the next track.
 
+player-next-album (*B*)
+	Skips to the next album. If *shuffle*=`tracks` or a playlist is active,
+	skips to the next track.
+
 player-pause (*c*)
 	Toggles pause.
 
@@ -525,6 +532,10 @@ player-play [filename] (*x*)
 
 player-prev (*z*)
 	Skips to the previous track.
+
+player-prev-album (*Z*)
+	Skips to the previous album. If *shuffle*=`tracks` or a playlist is active,
+	skips to the previous track.
 
 player-stop (*v*)
 	Stops playback.
--- a/cmus.c
+++ b/cmus.c
@@ -29,6 +29,7 @@
 #include "utils.h"
 #include "path.h"
 #include "options.h"
+#include "command_mode.h"
 #include "xmalloc.h"
 #include "debug.h"
 #include "load_dir.h"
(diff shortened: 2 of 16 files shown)


cmus.c
```diff
@@ -104,6 +105,22 @@ void cmus_prev(void)
 		player_set_file(info);
 }
 
+void cmus_rand(void)
+{
+	struct track_info *info;
+
+	view_clear(QUEUE_VIEW);
```
gavtroy, line 112: Is it okay to clear the queue this way? `cmus_next()` has thread synchronization when removing items from the queue, but I didn't dig into why.

pgaskin: I think so, as long as `cmus_rand` is only being called from the UI thread.

options.c
```diff
@@ -1002,25 +1000,38 @@ static void toggle_set_term_title(void *data)
...
+	strscpy(buf, shuffle_names[shuffle], size);
 }
 
 static void set_shuffle(void *data, const char *buf)
 {
-	int old = shuffle;
-	if (!parse_bool(buf, &shuffle))
+	int tmp;
+
+	if (!parse_enum(buf, 0, 2, shuffle_names, &tmp))
```
pgaskin, line 1016: While this will work correctly and maintain backwards-compatibility if `1`/`0` are used, it won't work with `true`. If the intent is to maintain backwards compatibility, it should probably be set to `tracks` if it's `true` and `parse_enum` fails.

Doc/cmus.txt
```diff
@@ -1141,7 +1145,7 @@ show_playback_position (true)
 show_remaining_time (false)
 	Display remaining time instead of elapsed time.
 
-shuffle (false)
+shuffle (off) [off, tracks, albums]
 	Play in shuffled order. Shuffle works in the library views (1-2) and
 	playlist view (3).
```
pgaskin, line 1150: To make this less ambiguous, it should say something about `tracks` being any track within the entire lib rather than just the current album (speaking of this, an `album-tracks` option might be useful). It should also mention that `albums` only chooses the first track of a random album after the current one has ended. To prevent confusion, it might be a good idea to mention that the `albums` option does nothing if the current `aaa_mode` is `album`.

gavtroy:
Memo to self to fix `shuffle_insert()` in `cmd_win_activate()`.

Doc/cmus.txt
gavtroy: `album-tracks` sounds like `aaa_mode=album` + `shuffle=tracks`, though I doubt many people really know what `aaa_mode` is (until it's tripped them up enough, at least).

gavtroy:
Sorry I keep changing this. After rolling it around my head for a while I decided `player-next-album` / `player-prev-album` would be more intuitive than `player-rand`.

pgaskin:
I've been running e6d6a284445b7bcdeae7ab64a97b42817c8eb38b for a while (and actively using it), and it works well. I'm going to test 350215609aefa3f9d63ba50d2fd71297b0dd0c17 and do a proper review later this week.

cmus.c
```diff
@@ -104,6 +105,38 @@ void cmus_prev(void)
...
+
+	if (info)
+		player_set_file(info);
+}
+
+void cmus_prev_album(void)
+{
+	struct track_info *info;
+
+	view_clear(QUEUE_VIEW);
```
pgaskin, line 128: I'm on the fence about this one (and the one in `cmus_next_album`).

Currently, switching tracks manually while there are tracks in the queue will replace the current track and the overall playback position, while preserving the playlist. Going to the next track will pop an item from the queue, and going to the previous track is the same as manually selecting that track. After the playlist ends, playback will resume at the previous location (i.e. the last track selected by something other than the playlist).

While this change makes sense in some ways, it is not completely consistent with the existing commands. If this is kept, a note should probably be added to the documentation.

@gavtroy, can you explain your reasoning behind doing this?

pgaskin (commented):
Works well and hasn't caused any regressions so far. I especially like how you preserved the shuffle previous/next list preservation behavior to keep it consistent until a reshuffle.

Doc/cmus.txt
```diff
@@ -513,6 +516,10 @@ pl-rename <name>
 player-next (*b*)
 	Skips to the next track.
 
+player-next-album (*B*)
```
pgaskin, line 519: I wonder if `player-next-shuffled` might be a better name for this.

cmus.c
gavtroy: To me "next album" doesn't even make sense while tracks are queued, but obviously the user can do it so we have to think about it. The most intuitive behavior would be to skip through the queue until finding a track with a different album, but that would be painful to implement for such a rare scenario, especially considering the queue contains "simple" tracks (i.e. not album aware). So I considered what easier options I had:
1. Ignore the command.
2. Go to next queue item even if it's the same album.
3. Go to 1st track of album outside the queue, but (surprisingly?) continue with the queue immediately afterwards.
4. Clear queue and go to 1st track of album outside the queue.

These options all looked bad so I just went with what I personally thought would be least surprising to the user (4).

For `prev_album` it's even less clear what the user wants to happen. Since I had decided to clear the queue in `next_album` I supposed the same reasoning would apply here, although you're right that it is then inconsistent with `cmus_prev`.

I don't have strong feelings about this either way.

Doc/cmus.txt
gavtroy: It does also work outside shuffle mode where it'll just advance to the next album in library or sorted order.

What you suggest sounds like the `player-next-rand` I implemented originally, but after playing around with it I thought this would be more intuitive and more discoverable.

cmus.c
pgaskin: > The most intuitive behavior would be to skip through the queue until finding a track with a different album

Personally, I'd be against that because of the complexity (as you mentioned) and also since that's different than how track shuffle works (it doesn't pick a random track from the queue).

> Ignore the command.

I agree that that's unacceptable.

> Go to next queue item even if it's the same album.

That's what the existing next/prev commands are for, so I agree with not going with that either.

> Go to 1st track of album outside the queue, but (surprisingly?) continue with the queue immediately afterwards

That's more of what I'd have expected based on the existing behavior, which I admit is surprising at first (but I've gotten quite used to it and I like it).

> Clear queue and go to 1st track of album outside the queue.
>
> These options all looked bad so I just went with what I personally thought would be least surprising to the user (4).

I agree with the logic of that compared to other players, but I still think it's different from the existing behavior. OTOH, it's a new feature, so I guess it doesn't *have* to match the other commands.

> For prev_album it's even less clear what the user wants to happen

Personally, I believe, it should go to the first track of the previous album or whatever the shuffling is done by (like it does now), but keep the queue (like the current prev).

> I don't have strong feelings about this either way.

Neither do I, and I generally like it the way it is other than the somewhat unexpected queue behavior. Personally, I'd prefer to preserve the queue, but I'd be fine with either adding a note about it to the documentation. The way I see it, the queue should only be affected by commands which manipulate it or pop from it, and should otherwise be ignored by the other commands.

It'd be nice to have another opinion from someone else, though.

Doc/cmus.txt
pgaskin: OTOH, without reading the code, I'd expect `player-next-album` to go to the next album regardless of the shuffle mode. I'd read `player-next-shuffled` as `Skips to the next shuffled entry. If not playing from the library, skips to the next track.`.

> would be more intuitive and more discoverable

I agree.

pgaskin: In that case, after thinking about it, I agree that it's better than the alternatives.

cmus.c
gavtroy: I've decided to just remove the clear for now. Other opinions are still welcome if anyone feels strongly about it...

gavtroy:
I rebased to be able to fixup the "smart replaygain" code from `if ... shuffle` to `if ... shuffle == SHUFFLE_TRACKS`. I haven't made any other alterations (other than removing the play queue wiping).

I haven't had any issues in the last 4 months of running this.

pgaskin:
> I haven't had any issues in the last 4 months of running this.

Neither have I. I think this is ready now.

pgaskin (approved):

pgaskin:
@nefthy?

nefthy (approved):

flyingmutant:
Merged, thanks!

PR #1049, opened 2021-01-22
Implement smart ReplayGain setting

### LINKED ISSUE #275: Use intelligent setting for replaygain
I have set my `replaygain` config to `album-preferred` because I often play albums a a whole and I don't want the volume to change then. But when I play random or with a `Play Queue` (with different albums) I would rather have it set to `track-preferred`.

So I'd love to see another option called e.g. `intelligent` which changes between `album-` and `track-preferred` depending on the `shuffle` state, the `aaa_mode` state and the `Play Queue`.

Allavaz:
It works by using the album gain if all of these conditions are met simultaneously:

- cmus is playing from the library (instead of a playlist).
- `aaa_mode` is set to `album from library`.
- Shuffle is off.
- Queue isn't active.

If any of these conditions isn't met, it uses the track gain. 

If the conditions are met but the song doesn't have album gain tag, it uses the track gain (like when using `album-preferred`), and viceversa.

Closes #275.


options.c
```diff
@@ -751,7 +751,7 @@ static void toggle_repeat(void *data)
 }
 
 static const char * const replaygain_names[] = {
-	"disabled", "track", "album", "track-preferred", "album-preferred", NULL
+	"disabled", "track", "album", "track-preferred", "album-preferred", "smart", NULL
 };
```
gavtroy, line 756: Need to increment the "4" in `set_replaygain()` and the "5" in `toggle_replaygain()`

player.c
```diff
@@ -354,7 +355,16 @@ static void update_rg_scale(void)
...
+	if (replaygain == RG_SMART) {
+		if (play_library && aaa_mode == AAA_MODE_ALBUM 
+			&& !shuffle && !cmus_queue_active()) {
+			gain = player_info_priv.ti->rg_album_gain;
+			peak = player_info_priv.ti->rg_album_peak;
+		} else {
+			gain = player_info_priv.ti->rg_track_gain;
+			peak = player_info_priv.ti->rg_track_peak;
+		}
+	} else if (replaygain == RG_TRACK || replaygain == RG_TRACK_PREFERRED) {
```
gavtroy, line 367: It would be neater to extend this condition instead.
```
if (RG_TRACK or RG_TRACK_PREFERRED or avoid_album_gain)
```
where avoid album would be set at the start of the function as
```
aviod_album_gain = RG_SMART and (shuffle or queue_active)
```

player.c
```diff
@@ -369,6 +379,15 @@ static void update_rg_scale(void)
...
 			peak = player_info_priv.ti->rg_track_peak;
+		} else if (replaygain == RG_SMART) {
+			if (play_library && aaa_mode == AAA_MODE_ALBUM 
+				&& !shuffle && !cmus_queue_active()) {
+				gain = player_info_priv.ti->rg_track_gain;
+				peak = player_info_priv.ti->rg_track_peak;
+			} else {
+				gain = player_info_priv.ti->rg_album_gain;
+				peak = player_info_priv.ti->rg_album_peak;
+			}
```
gavtroy, line 390: Likewise, it would be neater to just extend the first condition
```
if (replaygain == RG_TRACK_PREFERRED || avoid_album_gain)
```

gavtroy (changes requested):
* Don't forget to update `Doc/cmus.txt`.
* I don't think aaa_mode should be checked. Albums are still albums regardless.
* Checking the queue sounds good, although I wonder if it might cause volume jumps when audio is pre-buffered. I don't really know how it works, just thinking aloud :slightly_smiling_face:.
* I'll suggest a name like "shuffle_aware" might be better, but I'm also fine with the current name.

cmus.c
```diff
@@ -509,3 +509,7 @@ void cmus_raise_vte(void)
 		}
 	}
 }
+
+bool cmus_queue_active(void) {
+	return play_queue_active;
+}
```
pgaskin, line 515: There should be a newline at the end of the file.

options.c
nefthy: good catch. IMO it would be better do something like:

<pre lang="C">static const size_t replaygain_names_len = sizeof(replaygain_names) / sizeof(replaygain_names[0]) - 1;</pre>

and use that instead of hardcoding the length in the other functions.

nefthy (changes requested):
Please updated doc/cmus.txt to reflect your changes

nefthy:
IMO this should be the default option for replaygain, it's more sensible than disabled.

gavtroy:
Disabled is a good default. Unless every track in your library is tagged it will be confusing and make volume jumps even worse. Anyone tagging their own library will know to turn it on, and the (few) players I've checked have it off by default.

Anyway, the One True Setting is `album-preferred`.[¹](https://productionadvice.co.uk/tidal-normalization-upgrade/) :slightly_smiling_face:

Allavaz:
* Updated doc/cmus.txt.
* Added a line at end of file to `cmus.c`
* Followed @nefthy suggestion to use a variable for the length of `replaygain_names`.
* Followed @gavtroy suggestion to use a bool to check that any of the conditions isn't met, to use track gain instead (I almost forgot to check if a playlist is being played!). Also stopped checking for `aaa_mode` since it's not actually needed.

cmus.c
Allavaz: Done

player.c
Allavaz: Looks good, but we also need to check if we're playing from a playlist instead of the library.

gavtroy (approved):

nefthy:
> Anyway, the One True Setting is `album-preferred`.[¹](https://productionadvice.co.uk/tidal-normalization-upgrade/)

true :)

nefthy (approved):

Doc/cmus.txt
```diff
@@ -1101,7 +1101,7 @@ repeat (false)
 repeat_current (false)
 	Repeat current track forever.
 
-replaygain (disabled) [track, album, track-preferred, album-preferred]
+replaygain (disabled) [track, album, track-preferred, album-preferred, smart]
 	Enable Replay Gain.
```
pgaskin, line 1105: It'd probably be a good idea to describe what `smart` does here since it's not as obvious as the others.

pgaskin (approved):
LGTM

Doc/cmus.txt
flyingmutant: @Allavaz can you please add the documentation for the new option?

Allavaz: Just pushed a commit trying to explain what it does but I'm not sure if it's clear enough. Let me know if it's confusing or anything.

flyingmutant:
Thanks!

P.S. Github inserts `Co-authored-by: Gregory Petrosyan <gregory.petrosyan@gmail.com>` into the squash commits automatically now just because I press the green button? Thats emberassing :-(

PR #1052, opened 2021-01-30
Ignore duplicates when adding tracks to the library

### LINKED ISSUE #423: Duplicate files while scanning library

gavtroy:
Closes #423.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1085,6 +1085,11 @@ icecast_default_charset (ISO-8859-1)
 
 	Note: This is used only if the metadata is not valid UTF-8.
 
+ignore_duplicates (false)
+	Ignore duplicates when adding tracks to the library. A track is a duplicate
+	if it has a matching artist, album, disc number, track number, and track
+	name.
+
 lib_add_filter [`Filter`]
 	Apply filter when adding files to the library. See *FILTERS*.
 
--- a/lib.c
+++ b/lib.c
@@ -24,6 +24,7 @@
 #include "rbtree.h"
 #include "debug.h"
 #include "utils.h"
+#include "u_collate.h"
 #include "ui_curses.h" /* cur_view */
 
 #include <pthread.h>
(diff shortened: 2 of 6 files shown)


Doc/cmus.txt
```diff
@@ -1064,6 +1064,11 @@ icecast_default_charset (ISO-8859-1)
 
 	Note: This is used only if the metadata is not valid UTF-8.
 
+ignore_duplicates (true)
+	Ignore duplicates when adding tracks to the library. A track is a duplicate
+	if it has a matching artist, album, disc number, track number, and track
```
pgaskin, line 1069: I wonder if it might be a good idea to check the length too.

gavtroy: On the one hand I'd worry the same track could be +/- 1 second depending on encoder, album edition, etc. At the same time, in my quite tidy library there was just one duplicate: an album with two "0. Bonus Track" that were in fact different bonus tracks. A length check would have caught that.

Overall I don't know which works better for all the real world messy libraries out there. If I go by feelings, the second "0. Bonus Track" _deserved_ to get removed. :grin:

pgaskin: > A length check would have caught that.

That's the kind of situation I had in mind.

> real world messy libraries out there

For simplicity, it might be easier to keep it without the length check, and to add an option to force-add tracks.

lib.c
```diff
@@ -159,13 +159,52 @@ static int is_filtered(struct track_info *ti)
...
+	rb_for_each_entry(artist, node, &lib_artist_root, tree_node) {
+		if (strcmp(artist->name, ti_artist_name) == 0)
+			break;
+	}
+	if (!artist)
+		return false;
+
+	const char *ti_album_name = tree_album_name(ti);
+	rb_for_each_entry(album, node, &artist->album_root, tree_node) {
+		if (strcmp(album->name, ti_album_name) == 0)
```
pgaskin, line 182: For consistency, this should probably use `special_album_cmp` instead. This function will also check the date and will use the sort name instead if present.

options.c
```diff
@@ -76,6 +76,7 @@ int smart_artist_sort = 1;
 int scroll_offset = 2;
 int rewind_offset = 5;
 int skip_track_info = 0;
+int ignore_duplicates = 1;
```
pgaskin, line 79: @flyingmutant, @nefthy, do you think this should this be enabled by default? Personally, the only potential issue I see is confusion if an added track doesn't appear without an explanation (especially if the user tries to add a specific track).

Also, depending on how strict we are about breaking changes in releases, this might be considered to be a breaking change since it changes the behavior when adding tracks. However, this won't affect existing libraries, so it could also be argued the other way around.

gavtroy: Actually the same code is used when lib.pl is being imported each startup, so it will remove dups from existing libraries if left 
 enabled by default.

pgaskin: Looks like I missed that since I was testing with an existing rc with this disable previously. In that case, we should probably leave this disabled until maybe the next major version or after we have some more feedback on how this is used.

lib.c
gavtroy: Thanks for the pointer. It seems to directly use those functions I'd have to construct temporary album / artist objects - doable, although perhaps just slightly ugly / inefficient.

- The overall album date is defined by the newest track, so adding it here is tricky. I could add it below as a `ti` check, but I'm anyway not convinced it's needed. The most likely cause of a different date (but equal everything else) is that the date is blank or filled in wrong, rather than it not being a duplicate.
- I wouldn't care if the sort name is _different_, but I suppose theoretically some tracks could set a sort name and leave "name" blank so it is worth accounting for.
- Looking at this function, we could very easily be using the "collkey_" versions of everything to be case insensitive, and I probably should.

gavtroy: > some tracks could set a sort name and leave "name" blank

I'm gonna disagree with yesterday me. Nobody should be doing this.

gavtroy:
Just a simple rebase to fix the conflict. I still think we compare enough already that comparing the track durations isn't going to be helpful, but let me know if there's strong objection and I can look at it.

pgaskin (approved):
Tested Works as expected.

https://copr.fedorainfracloud.org/coprs/pgaskin/cmus-testing/build/6167850/
https://github.com/pgaskin/cmus/commits/a52df51b476c7aba0b08ebe59d622a198ba95612
https://nightly.link/pgaskin/rpms/actions/runs/5538232778

pgaskin:
@flyingmutant, this then a release?

flyingmutant:
> @flyingmutant, this then a release?

What about your draft PRs?

pgaskin:
> What about your draft PRs?

I personally use packages built from my dev fork of cmus including them, so I (and the few hundred other users of those packages) am not affected too much if they doesn't go out in the next release.

The pl-env one works fine, but there's a few minor things I need to change based on the review comments. I'll do them soon, probably by the end of today, just in case you want to release it.

PR #1065, opened 2021-03-25
Limit max draw width to 1023 columns

gavtroy:
Previously we allowed for 511 columns and would crash if the window was
larger than this. The browser tab had an even lower limit when the path
contained multibyte characters.

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -107,24 +107,24 @@ int using_utf8 = 0;
 static char *lib_autosave_filename;
 static char *play_queue_autosave_filename;
 
+#define PRINT_BUFFER_SZ 4096
+/* one character can take up to 4 bytes in UTF-8 */
+#define PRINT_BUFFER_MAX_WIDTH (PRINT_BUFFER_SZ / 4 - 1)
+
+static char print_buffer[PRINT_BUFFER_SZ];
+/* destination buffer for utf8_encode_to_buf and utf8_decode */
+static char conv_buffer[PRINT_BUFFER_SZ];
+
 /* shown error message and time stamp
  * error is cleared if it is older than 3s and key was pressed
  */
-static char error_buf[512];
+static char error_buf[PRINT_BUFFER_SZ];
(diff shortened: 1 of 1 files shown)


ui_curses.c
```diff
@@ -921,13 +922,17 @@ static void update_window(struct window *win, int x, int y, int w, const char *t
...
+	int c, i;
 
 	win->changed = 0;
 
 	bkgdset(pairs[CURSED_WIN_TITLE]);
-	sprintf(print_buffer, " %-*s ", w - 2, title);
+	c = snprintf(print_buffer, sizeof(print_buffer), " %s", title);
+	int pad_sz = w - 1 - u_str_width(title);
+	memset(print_buffer + c, ' ', pad_sz);
+	print_buffer[c + pad_sz] = '\0';
```
gavtroy, line 933: This small fix is not strictly related to the buffer size changes, but I realized sprintf is not multibyte aware (duh) so it won't always add enough padding in the title of the browser tab.

ui_curses.c
```diff
@@ -110,18 +110,19 @@ static char *play_queue_autosave_filename;
...
+static char error_buf[4096];
 static time_t error_time = 0;
 /* info messages are displayed in different color */
 static int msg_is_error;
 static int error_count = 0;
 
 static char *server_address = NULL;
 
-static char print_buffer[1024];
+static char title_buffer[4096];
```
nefthy, line 121: Im not too fond of replacing multiple local variables with a single file global variable.

nefthy (changes requested):

ui_curses.c
gavtroy: Actually, `title` will already have been padded by `format_str()`. Let me just remove the confusing padding specifier here.

nefthy (approved):

nefthy:
LGTM

flyingmutant:
I somehow missed that this was ready for merge, sorry! Finally merged.

PR #1078, opened 2021-04-18
Fix segfault when reading mp4 files

### LINKED ISSUE #1076: Segmentation fault when opening any m4a/m4b/mp4 file
I tried to refresh the music library on cmus today, but it seems that whenever I try to add an m4a file it encounters a `segmentation fault` error. I've tried this with files without metadata and it still happens. This also seems to happen with m4b and mp4 files as well.

So I got a stack trace with `gdb cmus`:

```
#0  0x00007fffe0002128 in ?? ()
#1  0x00007fffefd3ef86 in ?? () from /usr/lib/libmp4v2.so.2
#2  0x00007fffefd41e59 in ?? () from /usr/lib/libmp4v2.so.2
#3  0x00007fffefd42d29 in ?? () from /usr/lib/libmp4v2.so.2
#4  0x00007fffefd37c57 in MP4Read () from /usr/lib/libmp4v2.so.2
#5  0x00007ffff7072789 in ?? () from /usr/lib/cmus/ip/mp4.so
#6  0x0000555555580c13 in ip_open ()
#7  0x000055555556b275 in ?? ()
#8  0x000055555556c020 in cache_get_ti ()
#9  0x0000555555582269 in ?? ()
#10 0x0000555555582746 in ?? ()
#11 0x000055555558275b in ?? ()
#12 0x00005555555828a4 in ?? ()
#13 0x000055555559fa48 in ?? ()
#14 0x00007ffff7f85299 in start_thread () from /usr/lib/libpthread.so.0
#15 0x00007ffff7c2f053 in clone () from /usr/lib/libc.so.6
```
Which I then used to trace back [this issue](https://github.com/TechSmith/mp4v2/issues/47). It seems like a change in `libmp4v2` v4.1.4/4.1.5 is causing cmus to not compile [and maybe segfault]. Currently my version of `libmp4v2` is `4.1.4-1`, so I would be affected if that was the case.

**Edit:** On further inspection of the stacktrace, the function  it seems to reference is `MP4Read`, which is also the [same function](https://github.com/TechSmith/mp4v2/blob/master/include/mp4v2/file.h#L336) that contains the C++ instruction that causes the compilation issue in the first place. This change is probably the cause.

Is there anything that can be done here?

1) `uname -a`:
```
Linux primary 5.9.16-1-MANJARO #1 SMP PREEMPT Mon Dec 21 22:00:46 UTC 2020 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  aac:
    Priority: 50
    File Types: aac
  modplug:
    Priority: 50
    MIME Types:
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  ffmpeg:
    Priority: 30
    MIME Types:

  pulse
  alsa
  oss
  jack
  ao
```

OxygenCobalt:
Cmus will encounter a segmentation fault reading mp4/m4a/m4b if the shared `libmp4v2` library is at v4.1.4+. This is because that release changed the `MP4Read` function from an overloaded method to a single method with a default argument of `nullptr`, causing cmus to segfault when it only specified one argument. 

This PR fixes that by changing the function call to specify `NULL` as the second argument, which prevents cmus from segfaulting.

Fixes #1076.

Related: [mp4v2/#47](https://github.com/TechSmith/mp4v2/issues/47)

Feel free to raise any concerns with this fix.

### DIFF
--- a/ip/mp4.c
+++ b/ip/mp4.c
@@ -171,12 +171,7 @@ static int mp4_open(struct input_plugin_data *ip_data)
 	NeAACDecSetConfiguration(priv->decoder, neaac_cfg);
 
 	/* open mpeg-4 file */
-#ifdef __APPLE__
-	/* MP4Read doesnt have a second argument on macos */
-	priv->mp4.handle = MP4Read(ip_data->filename);
-#else
 	priv->mp4.handle = MP4Read(ip_data->filename, NULL);
-#endif
 
 	if (!priv->mp4.handle) {
 		d_print("MP4Read failed\n");
--- a/ip/mp4.c
+++ b/ip/mp4.c
@@ -171,7 +171,12 @@ static int mp4_open(struct input_plugin_data *ip_data)
 	NeAACDecSetConfiguration(priv->decoder, neaac_cfg);
 
 	/* open mpeg-4 file */
+#ifdef __APPLE__
+	/* MP4Read doesnt have a second argument on macos */
+	priv->mp4.handle = MP4Read(ip_data->filename);
+#else
 	priv->mp4.handle = MP4Read(ip_data->filename, NULL);
+#endif
 
 	if (!priv->mp4.handle) {
 		d_print("MP4Read failed\n");
--- a/ip/mp4.c
+++ b/ip/mp4.c
@@ -171,11 +171,8 @@ static int mp4_open(struct input_plugin_data *ip_data)
 	NeAACDecSetConfiguration(priv->decoder, neaac_cfg);
 
 	/* open mpeg-4 file */
-#ifdef MP4_DETAILS_ALL
-	priv->mp4.handle = MP4Read(ip_data->filename, 0);
-#else
-	priv->mp4.handle = MP4Read(ip_data->filename);
-#endif
+	priv->mp4.handle = MP4Read(ip_data->filename, NULL);
+	
 	if (!priv->mp4.handle) {
 		d_print("MP4Read failed\n");
 		goto out;
--- a/ip/mp4.c
+++ b/ip/mp4.c
@@ -171,9 +171,8 @@ static int mp4_open(struct input_plugin_data *ip_data)
 	NeAACDecSetConfiguration(priv->decoder, neaac_cfg);
 
 	/* open mpeg-4 file */
-
 #ifdef __APPLE__
-	/* libmp4v2 on Mac-OS doesnt have a second argument for some reason */
+	/* MP4Read doesnt have a second argument on macos */
 	priv->mp4.handle = MP4Read(ip_data->filename);
 #else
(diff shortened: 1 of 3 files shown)


OxygenCobalt:
Weird. `MP4Read` on mac only takes two arguments. Is there a way to do an `ifdef` for checking if we are compiling on a mac?

ip/mp4.c
```diff
@@ -171,11 +171,13 @@ static int mp4_open(struct input_plugin_data *ip_data)
 	NeAACDecSetConfiguration(priv->decoder, neaac_cfg);
 
 	/* open mpeg-4 file */
-#ifdef MP4_DETAILS_ALL
-	priv->mp4.handle = MP4Read(ip_data->filename, 0);
-#else
+#ifdef __APPLE__
```
nefthy, line 174: The API changed with version 2.1.0 so the conditional should be
```suggestion
#if MP4V2_PROJECT_version_major < 2 ||  (MP4V2_PROJECT_version_major == 2 && MP4V2_PROJECT_version_minor < 1)
```

OxygenCobalt: Good Idea.

OxygenCobalt:
Hm. Good idea. I'll try to take a crack at it later.

OxygenCobalt:
Okay, I got blind-sighted by other projects and forgot about this PR, but now I'm actually going to try to add a configuration for this, like you suggested. Give me some time and I'll try to make this PR actually become workable.

OxygenCobalt:
I give up. I can't get the configure script working at all. For some reason the mere act of adding my fix causes the CI to repeatedly fail, continues trying to import `mp4.h` despite that not existing and that shouldn't occurring in the first place. I have no idea why. I have no idea how. I can only assume that this bug is unfixable at this point. I'm just going to switch to MPD since that seems to be actually maintained. Good luck to anyone else trying to tackle this mess.

PR #1085, opened 2021-06-02
Implement pl_env_vars option for library/playlist/cache relocation

### LINKED ISSUE #245: relative paths for playlist (pls) files?
Hello,

I would like to use playlist (.pls) files with relative instead of absolute paths in cmus.  I would like for cmus to create relative paths to files when saving .pls files and I would like cmus to recognize those same paths.

For example, at the moment, if I have a playlist consisting of one song:

`08 Endless Endless.m4a`

And I save that playlist using the following command:

`:load ~/Music/playlist.pls`

Opening the playlist.pls file shows something like the following (on OSX):

`/Users/will/Music/audio_files/Kraftwerk/Trans Europe Express (Remastered)/08 Endless Endless.m4a`

But I would like for the playlist contents to look like this:

`~/Music/audio_files/Kraftwerk/Trans Europe Express (Remastered)/08 Endless Endless.m4a`

And I would like cmus to recognize this.  I am not sure if .pls files even permit relative paths though nor can I see options for changing any defaults in cmus.  I'm sure I'm missing something very simple but I'd appreciate advice.

Thanks,

Will

### LINKED ISSUE #834: Relative paths in lib.pl and playlist.pl
I mount a remote folder into my home folder, which contains my music collection. I would like to be able to share this lib.pl and playlist.pl file with other computers that I use, without needing to re-scan (because this takes many hours).

In other words, across two machines, these reflect the same directories remotely mounted:

/home/elje/music/J/Jesus, Zola/Versions/06- Sea Talk.mp3
/Users/elje/music/J/Jesus, Zola/Versions/06- Sea Talk.mp3

I'm requesting a feature to make cmus understand relative paths in both lib.pl and playlist.pl so that I can simply share these files across machines. I attempted to grep out the full path name and replace it with $HOME/... but it didn't work.

Thanks for your consideration!

pgaskin:
See `pl_env.h` for more details.

closes #834
closes #245

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1115,6 +1115,12 @@ pause_on_output_change (false)
 
     Supported output plugins: pulse.
 
+pl_env_vars
+	Comma separated list of environment variables to substitute when saving
+	library/playlist files. The paths must be absolute to take effect.
+
+	Note: This option will not take full effect until cmus has been restarted.
+
 pl_sort () [`Sort Keys`]
 	Sort keys for the playlist view (3). Empty value disables sorting and
 	enables manually moving tracks.
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1117,7 +1117,8 @@ pause_on_output_change (false)
 
 pl_env_vars
 	Comma separated list of environment variables to substitute when saving
-	library/playlist files. The paths must be absolute to take effect.
+	library/playlist files. The paths must be absolute to take effect. See
+	`Library Environment Variables`.
 
 	Note: This option will not take full effect until cmus has been restarted.
 
@@ -1685,6 +1686,46 @@ USER
 
 USERNAME
 	Fallback for *USER*.
+
+@h2 Library Environment Variables
+
+cmus supports using environment variables to allow the library, cache, and
+playlists to be relocated or synced. The environment variables to use are
+specified using the `pl_env_vars` configuration option, which takes effect when
+exiting cmus.
+
+A good starting point is `:set pl_env_vars=HOME`. This will allow the `.cmus`
+directory to be shared between devices or users. Additional environment
+variables can be appended (delimited with commas) to `pl_env_vars` to redirect
+specific paths. The environment variables must contain absolute paths to be
+used. The last matching directory prefix will be used.
+
+If a file does not exist, it will be treated the same way as missing files
+usually are. If the file is in the cache, it will be preserved, and an error
+will be displayed upon playback. If the file is not in the cache, it will be
+removed from playlists and the library. There is one special exception: if a
+file is missing due to the environment variable being invalid/empty/unset, but
+it is still in the cache, it will be preserved in the cache as-is, in addition
+to being preserved in the library like usual.
+
+When using `:save`/`:load` and in the on-disk library/cache/playlists, the paths
+will be stored with the environment variable delimited using the character
+`\\x1F`, which is the invisible ASCII unit separator.
+
+This feature can also be used to quickly move a library to a different location
+while preserving all metadata, playlists, and the play counts. To do this, set
+an environment variable for each base path you want to change, set `pl_env_vars`
+to the list of environment variables, restart cmus, change the environment
+variables, restart cmus, unset `pl_env_vars`, then restart cmus again.
+
+If a library using `pl_env_vars` is opened with a version of cmus not supporting
+it, tracks referenced with environment variables will be forgotten. To restore
+the on-disk files to be compatible with older cmus versions, just unset
+`pl_env_vars`, and restart cmus. This will restore the paths for all defined
+environment variables to their actual path.
+
+It is safe to change `pl_env_vars` from outside cmus since this variable is
+applied when cmus exits.
 
 @h1 D-Bus Interface (MPRIS)
 
--- a/cache.c
+++ b/cache.c
@@ -26,6 +26,7 @@
 #include "xstrjoin.h"
 #include "gbuf.h"
 #include "options.h"
+#include "pl_env.h"
 
 #include <stdlib.h>
 #include <stdio.h>
@@ -124,10 +125,18 @@ static int valid_cache_entry(const struct cache_entry *e, unsigned int avail)
 static struct track_info *cache_entry_to_ti(struct cache_entry *e)
 {
 	const char *strings = e->strings;
-	struct track_info *ti = track_info_new(strings);
+	struct track_info *ti;
 	struct keyval *kv;
 	int str_size = e->size - sizeof(*e);
 	int pos, i, count;
+	char *proc_filename;
+
+	if (pl_env_var(strings, NULL) && (proc_filename = pl_env_expand(strings))) {
+		ti = track_info_new(proc_filename);
+		free(proc_filename);
+	} else {
+		ti = track_info_new(strings);
+	}
 
 	ti->duration = e->duration;
 	ti->bitrate = e->bitrate;
@@ -318,6 +327,7 @@ static void flush_buffer(int fd, struct gbuf *buf)
 
 static void write_ti(int fd, struct gbuf *buf, struct track_info *ti, unsigned int *offsetp)
 {
+	char *proc_filename = pl_env_reduce(ti->filename);
 	const struct keyval *kv = ti->comments;
 	unsigned int offset = *offsetp;
 	unsigned int pad;
@@ -334,7 +344,7 @@ static void write_ti(int fd, struct gbuf *buf, struct track_info *ti, unsigned i
 	e.mtime = ti->mtime;
 	e.play_count = ti->play_count;
 	e.bpm = ti->bpm;
-	len[count] = strlen(ti->filename) + 1;
+	len[count] = strlen(proc_filename) + 1;
 	e.size += len[count++];
 	len[count] = (ti->codec ? strlen(ti->codec) : 0) + 1;
 	e.size += len[count++];
@@ -359,7 +369,7 @@ static void write_ti(int fd, struct gbuf *buf, struct track_info *ti, unsigned i
 	if (pad)
 		gbuf_set(buf, 0, pad);
 	gbuf_add_bytes(buf, &e, sizeof(e));
-	gbuf_add_bytes(buf, ti->filename, len[count++]);
+	gbuf_add_bytes(buf, proc_filename, len[count++]);
 	gbuf_add_bytes(buf, ti->codec ? ti->codec : "", len[count++]);
 	gbuf_add_bytes(buf, ti->codec_profile ? ti->codec_profile : "", len[count++]);
 	for (i = 0; kv[i].key; i++) {
@@ -369,6 +379,8 @@ static void write_ti(int fd, struct gbuf *buf, struct track_info *ti, unsigned i
 
 	free(len);
 	*offsetp = offset + pad + e.size;
+
+	free(proc_filename);
 }
 
 int cache_close(void)
@@ -436,6 +448,18 @@ struct track_info *cache_get_ti(const char *filename, int force)
 	unsigned int hash = hash_str(filename);
 	struct track_info *ti;
 	int reload = 0;
+
+	if (pl_env_var(filename, NULL)) {
+		struct growing_keyvals c = {NULL, 0, 0};
+		keyvals_terminate(&c);
+
+		ti = track_info_new(filename);
+		ti->duration = 0;
+		track_info_set_comments(ti, c.keyvals);
+
+		track_info_ref(ti);
+		return ti;
+	}
 
 	ti = lookup_cache_entry(filename, hash);
 	if (ti) {
--- a/job.c
+++ b/job.c
@@ -36,6 +36,7 @@
 #include "xstrjoin.h"
 #include "ui_curses.h"
 #include "cue_utils.h"
+#include "pl_env.h"
 
 #include <string.h>
 #include <unistd.h>
@@ -351,7 +352,9 @@ static int handle_line(void *data, const char *line)
 	if (is_http_url(line) || is_cue_url(line)) {
 		add_url(line);
 	} else {
-		char *absolute = path_absolute_cwd(line, data);
+		char *absolute = pl_env_var(line, NULL)
+			? pl_env_expand(line)
+			: path_absolute_cwd(line, data);
 		add_file(absolute, 0);
 		free(absolute);
 	}
--- a/options.c
+++ b/options.c
@@ -160,6 +160,7 @@ char *window_title_alt_format = NULL;
 char *id3_default_charset = NULL;
 char *icecast_default_charset = NULL;
 char *lib_add_filter = NULL;
+char **pl_env_vars;
 
 static void buf_int(char *buf, int val, size_t size)
 {
@@ -268,6 +269,7 @@ static const struct {
 	{ "pl_sort", "" },
 	{ "id3_default_charset", "ISO-8859-1" },
 	{ "icecast_default_charset", "ISO-8859-1" },
+	{ "pl_env_vars", "" },
 	{ NULL, NULL }
 };
 
@@ -505,6 +507,56 @@ static void set_tree_width_max(void *data, const char *buf)
 	if (parse_int(buf, 0, 9999, &cols))
 		tree_width_max = cols;
 	update_size();
+}
+
+/* get_pl_env_vars converts the pl_env_vars array into a comma-separated list */
+static void get_pl_env_vars(void *data, char *buf, size_t size)
+{
+	if (!pl_env_vars) {
+		strscpy(buf, "", size);
+		return;
+	}
+	char *p = buf;
+	size_t r = size - 1;
+	for (char **x = pl_env_vars; *x; x++) {
+		if (x != pl_env_vars) {
+			if (!--r)
+				return; /* overflow */
+			*p++ = ',';
+		}
+		size_t l = strlen(*x);
+		if (!(r -= l))
+			return; /* overflow */
+		strcpy(p, *x);
+		p += l;
+	}
+	*p = '\0';
+}
+
+/* set_pl_env_vars splits the pl_env_vars config into the pl_env_vars array */
+static void set_pl_env_vars(void *data, const char *buf)
+{
+	if (pl_env_vars) {
+		free(*pl_env_vars);
+		free(pl_env_vars);
+	}
+	if (!*buf) {
+		pl_env_vars = NULL;
+	}
+	size_t n = 1;
+	for (const char *x = buf; *x; x++) {
+		if (*x == ',') {
+			n++;
+		}
+	}
+	char **a = pl_env_vars = xnew(char*, n+1);
+	for (char *x = *a++ = xstrdup(buf); *x; x++) {
+		if (*x == ',') {
+			*a++ = x+1;
+			*x = '\0';
+		}
+	}
+	*a = NULL;
 }
 
 /* }}} */
@@ -1516,6 +1568,7 @@ static const struct {
 	DN(tree_width_percent)
 	DN(tree_width_max)
 	DT(pause_on_output_change)
+	DN(pl_env_vars)
 	{ NULL, NULL, NULL, NULL, 0 }
 };
 
--- a/pl_env.c
+++ b/pl_env.c
@@ -1,0 +1,255 @@
+/*
+ * Copyright 2021-2023 Patrick Gaskin
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "pl_env.h"
+#include "options.h"
+#include "xmalloc.h"
+
+#include <ctype.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+
+static bool pl_env_contains_delimiter(const char *str)
+{
+	return !!strchr(str, PL_ENV_DELIMITER);
+}
+
+/**
+ * pl_getenv_normalized gets an env var and puts it into the required format for
+ * pl_env_reduce and pl_env_expand, which does exact string matching/replacement
+ * against the file paths. It converts backslashes to slashes on Windows,
+ * removes consecutive slashes, removes './' path segments, simplifies '../'
+ * path segments (and returns NULL if it would result in going above the
+ * uppermost directory), and removes the trailing slash. In addition, it trims
+ * the variable name before looking it up.
+ */
+static char *pl_env_getenv_normalized(const char *var)
+{
+	/* duplicate the var */
+	char *var_d = xstrdup(var);
+
+	/* trim the var */
+	char *var_s = var_d;
+	char *var_e = var_d + strlen(var_d);
+	while (var_s < var_e && isspace(*var_s)) var_s++;
+	while (var_e > var_s && isspace(*(var_e-1))) var_e--;
+	*var_e = '\0';
+
+	/* if it's empty, return NULL */
+	if (var_e-var_s == 0) return NULL;
+
+	/* get the var */
+	char *val = getenv(var_s);
+	free(var_d);
+
+	/* if it's nonexistent or empty, return NULL */
+	if (!val || !*val) return NULL;
+
+	/* duplicate the value */
+	char *new = xstrdup(val);
+
+#ifdef _WIN32
+	/* convert backslashes to slashes */
+	/* note: cmus uses forward slashes internally, but Windows accepts both */
+	/* even though they will get normalized on windows, we need paths to match exactly */
+	for (char *p = path; *p; p++)
+		*p = *p == '\\' ? '/' : *p;
+#endif
+
+	/* canonicalize the path in-place */
+	size_t r = 1, w = 1;
+	while (new[r]) {
+
+		/* handle the start of a segment */
+		if (w == 0 || new[w-1] == '/') {
+
+			/* handle empty segments */
+			if (new[r] == '/') {
+
+				/* skip the duplicate slashes */
+				while (new[r] == '/') r++;
+
+				continue;
+			}
+
+			/* handle '.' segments */
+			if (new[r] == '.' && (new[r+1] == '/' || !new[r+1])) {
+
+				/* skip them */
+				if (new[r += 1]) r++;
+
+				continue;
+			}
+
+			/* handle '..' segments */
+			if (new[r] == '.' && new[r+1] == '.' && (new[r+2] == '/' || !new[r+2])) {
+
+				/* if there aren't any parent directories left to skip, return NULL */
+				if (!w) return NULL;
+
+				/* remove the previous segment up to the '/' */
+				for (w--; w && new[w-1] != '/'; ) w--;
+
+				/* skip the '..' */
+				if (new[r += 2]) r++;
+
+				continue;
+			}
+		}
+
+		/* write the next character */
+		new[w++] = new[r++];
+	}
+
+	/* remove the trailing slash if the path isn't / */
+	if (w >= 2 && new[w-1] == '/') {
+		w--;
+	}
+
+	/* terminate the path */
+	new[w] = '\0';
+
+	return new;
+}
+
+char *pl_env_reduce(const char *path)
+{
+	if (!pl_env_vars || !*pl_env_vars || pl_env_var(path, NULL))
+		return xstrdup(path);
+
+	for (char **var = pl_env_vars; *var && **var; var++) {
+		char *val = pl_env_getenv_normalized(*var);
+		if (!val)
+			continue;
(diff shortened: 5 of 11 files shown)


pl_env.c
```diff
@@ -0,0 +1,248 @@
...
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * pl_getenv_normalized gets an env var and puts it into the required format for
+ * pl_env_save and pl_env_restore, which does exact string matching/replacement
+ * against the file paths.
+ */
+static char *pl_env_getenv_normalized(const char *var)
```
pgaskin, line 34: Even though this is currently run at least twice for each track on startup and shutdown, it shouldn't cause performance issues. I did some benchmarks, and it's only around 1-3 microseconds per run even for complex paths on a few of my devices, compiled with and without optimizations. I can reduce it to around 0.2 if I stack allocate the variables instead, but since the amount is negligible, at least for my library size, I'm going to leave it this way for now. I might implement caching later, though.

Doc/cmus.txt
```diff
@@ -1101,6 +1101,12 @@ mpris (true)
 output_plugin [roar, pulse, alsa, arts, oss, sndio, sun, coreaudio]
 	Name of output plugin.
 
+pl_env_vars
```
pgaskin, line 1104: I'm planning to write a more detailed section in the documentation for this.

pl_env.h
```diff
@@ -0,0 +1,220 @@
...
+ */
+char *pl_env_var_str(const char *path, const char **remainder);
+
+/**
+ * pl_env_var_len returns the length of the substituted environment variable
+ * name, if present. Otherwise, it returns 0.
+ */
+int pl_env_var_len(const char *path);
+
+#endif
```
pgaskin, line 220: Should I switch to static thread-local buffers to store the results? It's safe for how these functions are currently used (only one track is ever processed at a time), and it would essentially eliminate the overhead of this option if it's enabled.

pl_env.c
pgaskin: Here's what it would look like (it still needs a bit of cleanup):

<details>

```diff
diff --git a/options.c b/options.c
index 627cc1f..9b1f4b3 100644
--- a/options.c
+++ b/options.c
@@ -41,6 +41,7 @@
 #include "debug.h"
 #include "discid.h"
 #include "mpris.h"
+#include "pl_env.h"
 
 #include <stdio.h>
 #include <errno.h>
diff --git a/pl_env.c b/pl_env.c
index 642401f..4eccc0e 100644
--- a/pl_env.c
+++ b/pl_env.c
@@ -16,7 +16,6 @@
  */
 
 #include "pl_env.h"
-#include "debug.h"
 #include "options.h"
 #include "xmalloc.h"
 
@@ -26,33 +25,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+extern char **environ;
+
 /**
- * pl_getenv_normalized gets an env var and puts it into the required format for
- * pl_env_save and pl_env_restore, which does exact string matching/replacement
- * against the file paths.
+ * pl_env_norm normalizes the provided path into the required format for
+ * pl_env_save and pl_env_restore, which does exact string matching/replacement.
  */
-static char *pl_env_getenv_normalized(const char *var)
+static char *pl_env_norm(char *path)
 {
-       /* duplicate the var */
-       char *var_d = xstrdup(var);
-
-       /* trim the var */
-       char *var_s = var_d;
-       char *var_e = var_d + strlen(var_d);
-       while (var_s < var_e && isspace(*var_s)) var_s++;
-       while (var_e > var_s && isspace(*(var_e-1))) var_e--;
-       *var_e = '\0';
-
-       /* get the var */
-       char *val = getenv(var_s);
-       free(var_d);
-
-       /* if it's nonexistent or empty, return NULL */
-       if (!val || !*val) return NULL;
-
-       /* duplicate the value */
-       char *new = xstrdup(val);
-
 #ifdef _WIN32
        /* convert backslashes to slashes */
        /* note: cmus uses forward slashes internally, but Windows accepts both */
@@ -62,58 +42,58 @@ static char *pl_env_getenv_normalized(const char *var)
 
        /* canonicalize the path in-place */
        size_t r = 1, w = 1;
-       while (new[r]) {
+       while (path[r]) {
 
                /* handle the start of a segment */
-               if (w == 0 || new[w-1] == '/') {
+               if (w == 0 || path[w-1] == '/') {
 
                        /* handle empty segments */
-                       if (new[r] == '/') {
+                       if (path[r] == '/') {
 
                                /* skip the duplicate slashes */
-                               while (new[r] == '/') r++;
+                               while (path[r] == '/') r++;
 
                                continue;
                        }
 
                        /* handle '.' segments */
-                       if (new[r] == '.' && (new[r+1] == '/' || !new[r+1])) {
+                       if (path[r] == '.' && (path[r+1] == '/' || !path[r+1])) {
 
                                /* skip them */
-                               if (new[r += 1]) r++;
+                               if (path[r += 1]) r++;
 
                                continue;
                        }
 
                        /* handle '..' segments */
-                       if (new[r] == '.' && new[r+1] == '.' && (new[r+2] == '/' || !new[r+2])) {
+                       if (path[r] == '.' && path[r+1] == '.' && (path[r+2] == '/' || !path[r+2])) {
 
                                /* if there aren't any parent directories left to skip, return NULL */
                                if (!w) return NULL;
 
                                /* remove the previous segment up to the '/' */
-                               for (w--; w && new[w-1] != '/'; ) w--;
+                               for (w--; w && path[w-1] != '/'; ) w--;
 
                                /* skip the '..' */
-                               if (new[r += 2]) r++;
+                               if (path[r += 2]) r++;
 
                                continue;
                        }
                }
 
                /* write the next character */
-               new[w++] = new[r++];
+               path[w++] = path[r++];
        }
 
        /* remove the trailing slash if the path isn't / */
-       if (w >= 2 && new[w-1] == '/') {
+       if (w >= 2 && path[w-1] == '/') {
                w--;
        }
 
        /* terminate the path */
-       new[w] = '\0';
+       path[w] = '\0';
 
-       return new;
+       return path;
 }
 
 static bool pl_env_contains_marker(const char *str)
@@ -121,32 +101,67 @@ static bool pl_env_contains_marker(const char *str)
        return !!strchr(str, PL_ENV_DELIMITER);
 }
 
+const char *pl_env_cache(const char *var, int var_len)
+{
+       static char **cache = NULL;
+
+       if (!var) {
+               for (char **x = cache; x && *x; x++)
+                       free(*x);
+               free(cache);
+
+               size_t n = 0;
+               for (char **x = environ; *x; x++)
+                       n++;
+
+               char **new = cache = xnew(char*, n+1);
+               for (char **x = environ; *x; x++)
+                       if (!pl_env_contains_marker(*x))
+                               if (!pl_env_norm(strchr((*new++ = xstrdup(*x)), '=') + 1))
+                                       free(*--new);
+               *new = NULL;
+
+               return NULL;
+       }
+
+       size_t vl = var_len == -1
+               ? strlen(var)
+               : var_len;
+
+       const char *vs = var;
+       const char *ve = var + vl;
+       while (vs < ve && isspace(*vs)) vs++;
+       while (ve > vs && isspace(*(ve-1))) ve--;
+       vl = ve-vs;
+
+       if (memchr(vs, (int) vl, PL_ENV_DELIMITER) || memchr(vs, (int) vl, '='))
+               return NULL;
+
+       for (char **x = cache; x && *x; x++)
+               if (strncmp(*x, vs, vl) == 0 && (*x)[vl] == '=')
+                       return *x + vl + 1;
+
+       return NULL;
+}
+
 char *pl_env_save(const char *path)
 {
        if (!pl_env_vars || !*pl_env_vars || pl_env_var(path, NULL))
                return xstrdup(path);
        
        for (char **var = pl_env_vars; *var && **var; var++) {
-               char *val = pl_env_getenv_normalized(*var);
+               const char *val = pl_env_cache(*var, -1);
                if (!val)
                        continue;
-               if (!*val || pl_env_contains_marker(val)) {
-                       free(val);
-                       continue;
-               }
-               
+
                size_t val_len = strlen(val);
 
 #ifdef _WIN32
-               if (strncasecmp(path, val, val_len) != 0) {
-                       free(val);
+               if (strncasecmp(path, val, val_len) != 0)
                        continue;
-               }
 #else
-               if (strncmp(path, val, val_len) != 0) {
-                       free(val);
+               if (strncmp(path, val, val_len) != 0)
                        continue;
-               }
 #endif
 
                const char *rem = path + val_len;
@@ -154,10 +169,8 @@ char *pl_env_save(const char *path)
                /* always keep the slash at the beginning of the path, and only
                   use the env var if it replaces an entire path component (i.e.
                   it is a directory) */
-               if (*rem != '/') {
-                       free(val);
+               if (*rem != '/')
                        continue;
-               }
 
                size_t var_len = strlen(*var);
                size_t rem_len = strlen(rem);
@@ -172,7 +185,6 @@ char *pl_env_save(const char *path)
                ptr += rem_len;
                *ptr = '\0';
 
-               free(val);
                return new;
        }
 
@@ -184,20 +196,16 @@ char *pl_env_restore(const char *path)
        if (!path)
                return NULL;
 
-       const char *rem;
-       char *var = pl_env_var_str(path, &rem);
-       if (!var)
+       int len;
+       const char *var;
+       if (!(var = pl_env_var(path, &len)))
                return xstrdup(path);
 
-       char *val = pl_env_getenv_normalized(var);
-       free(var);
+       const char *val = pl_env_cache(var, len);
        if (!val)
                return xstrdup(path);
-       if (!*val || pl_env_contains_marker(val)) {
-               free(val);
-               return xstrdup(path);
-       }
 
+       const char *rem = pl_env_var_remainder(path, len);
        size_t val_len = strlen(val);
        size_t rem_len = strlen(rem);
 
@@ -209,7 +217,6 @@ char *pl_env_restore(const char *path)
        ptr += rem_len;
        *ptr = '\0';
 
-       free(val);
        return new;
 }
 
@@ -230,17 +237,6 @@ const char *pl_env_var_remainder(const char *path, int length)
        return path+length+2;
 }
 
-char *pl_env_var_str(const char *path, const char **remainder)
-{
-       int len;
-       const char *var;
-       if (!(var = pl_env_var(path, &len)))
-               return NULL;
-       if (remainder)
-               *remainder = pl_env_var_remainder(path, len);
-       return xstrndup(var, len);
-}
-
 int pl_env_var_len(const char *path)
 {
        int len;
diff --git a/pl_env.h b/pl_env.h
index 6899ab7..74533c5 100644
--- a/pl_env.h
+++ b/pl_env.h
@@ -171,6 +171,16 @@
  */
 #define PL_ENV_DELIMITER '\x1F'
 
+/**
+ * pl_env_cache has cached and normalized environment variables. If called with
+ * NULL, the cache is regenerated from the environment. If called with a
+ * variable name, the name is trimmed and the matching environment variable is
+ * returned. The value of the environment variable is the normalized path. If
+ * var is not NULL and var_len is not -1, the first var_len characters of var
+ * are used.
+ */
+const char *pl_env_cache(const char *var, int var_len);
+
 /**
  * pl_env_save checks the base path against the configured environment
  * variables, replaces the first match with a substitution, and returns a
@@ -204,13 +214,6 @@ const char *pl_env_var(const char *path, int *out_length);
  */
 const char *pl_env_var_remainder(const char *path, int length);
 
-/**
- * pl_env_var_str is like pl_env_var, but it returns a malloc'd null-terminated
- * string instead. If remainder is not NULL, it is set to point to the remainder
- * of the path.
- */
-char *pl_env_var_str(const char *path, const char **remainder);
-
 /**
  * pl_env_var_len returns the length of the substituted environment variable
  * name, if present. Otherwise, it returns 0.
diff --git a/ui_curses.c b/ui_curses.c
index 670a00b..89db534 100644
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -51,6 +51,7 @@
 #include "mixer.h"
 #include "mpris.h"
 #include "locking.h"
+#include "pl_env.h"
 #ifdef HAVE_CONFIG
 #include "config/curses.h"
 #include "config/iconv.h"
@@ -2477,6 +2478,8 @@ int main(int argc, char *argv[])
                }
        }
 
+       pl_env_cache(NULL, 0);
+
        setlocale(LC_CTYPE, "");
        setlocale(LC_COLLATE, "");
        charset = getenv("CMUS_CHARSET");
```

</details>

tonyfischetti:
Tested this PR—works great for me! No problems!

Doc/cmus.txt
pgaskin: Also, we should consider setting a default value of `HOME` for the next major release.

pl_env.c
pgaskin: I've pushed this.

pl_env.h
pgaskin: I've decided to leave it as-is since the performance benefit of using buffers is negligible, and it's safer not to use them.

nefthy:
The code is high quality and nicely documented. I'm however unsure if the whole concept of substituting environment variables is something we should persu. Sorry for not explaining in the coresponding issue.

pgaskin (commented):
Tested:
- `:save`/`:load`.
- Launching and exiting cmus multiple times.
   - Ensuring the library remains consistent.
   - Ensuring the cache doesn't grow unexpectedly (due to filenames being mishandled or not handled everywhere).
   - Ensuring the play count is preserved.
   - Ensuring the cache is always in sync with the library.
   - Ensuring performance and memory usage is consistent.
- Moving files and setting the environment variable to the new path.
- Switching between `pl_env_vars` values, ensuring the paths are updated properly,
  - while cmus is running.
  - while cmus is stopped.
- `:add`.
- Normal cmus usage.

With:
- `pl_env_vars=`.
- `pl_env_vars=HOME`, `pl_env_vars=   HOME   `.
- `pl_env_vars=TEST,HOME` with some files in `$TEST`, and with `$TEST` unset/set/invalid/empty (ensuring tracks are preserved while `TEST` is unset, empty, or invalid), and with `$TEST` having unnecessary `./`, `/`, and `../`s.
- `pl_env_vars=TEST , TEST1 , HOME` with various combinations of files including some having a directory starting with `$TEST` but not within `$TEST` (e.g. `TEST=/test` with a filename starting with `/testtest).

Note that when an environment variable is not set, it'll display as `<1f>VARIABLE<1f>` due to cmus' utf8 handling, but I'm not yet comfortable enough with the text processing code to make it support the invisible ASCII delimiters yet. This isn't going to show up unless the user has a configuration problem, and if they do then it'll be pretty obvious what it is since they would have had to set `pl_env_vars` in the first place and the error when playing the track explains it, so I think it's OK to leave it this way for now.

pgaskin:
> I'm however unsure if the whole concept of substituting environment variables is something we should persue

I originally thought the same thing, but ended up changing my mind after going through the different options and resolving the remaining inconsistencies I found with the patch I wrote in https://github.com/cmus/cmus/issues/834#issuecomment-787817221. In particular:

- I wanted to ensure this wouldn't have any effect or risk of regressions on users who don't use this option. The only thing this implementation adds to the default code path is a fast-failing check for if paths contain `\x1F`. In addition, the second commit in this PR also preprocesses the environment variables for performance reasons, but that commit can be reverted if necessary to limit the risk of regressions. I did fuzz the path normalization implementation to ensure it wouldn't crash, though.
- I wanted to minimize the risk of losing the library/playlist contents if the user made a configuration mistake with `pl_env_vars` enabled. To do this, I added the special case to `cache_get_ti` to preserve the tracks if an environment variable is missing.
- I wanted to ensure this option does not affect the expected behaviour of cmus. As explained in `lib_pl.h`, unlike alternative implementations, this implementation is identical whether the `pl_env_vars` is enabled to both cmus itself (since all processing is done when saving/loading the state) and to users.
- I wanted this option to be self-contained to keep it maintainable. To do this, I moved all of the path processing into `pl_env.{c,h}` and documented how it affects other parts of cmus.

Did you have a specific concern?

tonyfischetti:
@nefthy Just one person's opinion: I think it's super worthwhile to pursue this. I think it's a feature that _a lot_ of people would like. For example, I have all my playlists version controlled. It's fine for my home gnu/linux machines but at work I have a mac and my home is `/Users/tony`. Before I could solve it with a symlink but now `/home` (and other things) are unwrite-able, even with root.

And it looks like the work is already done and works great/great documentation

nefthy:
@pgaskin I'd summirize my reservations as follows:

- It's a niche use case.
- Pulling stuff out of the environment und doing string stuff always bares risk of security issues (especialy in C).
- It does not feel like the way cmus does stuff.

That said, I'm ok with merging.

pgaskin:
> It's a niche use case.

On the other hand, most applications can follow the home directory around. I do admit it's somewhat niche though, but that's true for quite a few of cmus' features.

> Pulling stuff out of the environment und doing string stuff always bares risk of security issues

While I agree with that statement in general, especially if this was server software or multi-year, I don't think it's relevant for cmus. A user with access to cmus, it's socket, or its config dir can already access everything it has access to via load, save, and commands like that, or even just modifying the cache/config. A more relevant point related to this would be the risk for regressions due to stability issues in the string manipulation code, and I was concerned with that too, hence the extensive testing I did on the side with the env var path simplification, the comments, and how I implemented it such that barely any of the code would run for non-users of the option.

In addition, I used allocated strings rather than shared buffers. Related to your third point, *is* somewhat un-cmus-like, but is a lot safer. Remember how many issues with memory safety, undefined behavior, and segmentation faults have been found in cmus' existing UI and string manipulation code.

And about consistency for the rest of cmus, I specifically limited substitution to the first absolute match for path prefixes only to prevent this from breaking assumptions about files in the cache.

> It does not feel like the way cmus does stuff.

But, I'd say this implementation feels like the most cmus-like way to do things out of the different options I could think of (fake bind-mount implementation, arbitrary environment substitution, automatically attempting to search for files, hooking file loads themselves, etc). Plus, I'd argue this feature itself *is* cmus-like since it's a shell application, and most shell applications use environment variables rather than depending on fixed paths and support data synchronization using dotfile managers.

> That said, I'm ok with merging

If you don't mind, can you do a quick look over the documentation to ensure it makes sense to someone who didn't implement this feature?

flyingmutant:
> Also, @flyingmutant, do you have an opinion on this (feature or implementation of it)?

I think both feature and implementation approach are sensible. It is certainly a niche feature, but so is the whole concept of a console music player :-) I did not have time to review the code properly yet, but I see *a lot* of thought and polish has went into this PR.

pgaskin:
> a lot of thought and polish

pgaskin:
> a lot of thought and polish

Yes. I didn't want to risk introducing regressions with this, so I took a bit of time to properly understand and test this, and I took the most hands-off approach possible.

I did a few final optimizations with my last force-push. I've also tested this with Valgrind and ASAN. I think this PR is complete now unless you want any changes to the documentation or want any changes to the implementation.

I'm currently running and using this daily with 3 env vars on Termux (aarch64), Debian (in WSL), and Fedora (x86). The only thing I haven't tested is the Windows compatibility (I haven't got around to getting cmus to build in mingw-w64), but it should work fine and won't cause issues if it doesn't.

cache.c
```diff
@@ -124,10 +125,18 @@ static int valid_cache_entry(const struct cache_entry *e, unsigned int avail)
 static struct track_info *cache_entry_to_ti(struct cache_entry *e)
 {
 	const char *strings = e->strings;
-	struct track_info *ti = track_info_new(strings);
+	struct track_info *ti;
 	struct keyval *kv;
 	int str_size = e->size - sizeof(*e);
 	int pos, i, count;
+	char *proc_fn;
```
flyingmutant, line 132: what does `proc_fn` mean?

options.c
```diff
@@ -504,6 +506,54 @@ static void set_tree_width_max(void *data, const char *buf)
...
+	char *p = buf;
+	size_t r = size - 1;
+	for (char **x = pl_env_vars; *x; x++) {
+		if (x != pl_env_vars) {
+			if (!--r)
+				return;
+			*p++ = ',';
+		}
+		size_t l = strlen(*x);
+		if (!(r -= l))
```
flyingmutant, line 525: Hmm, am I wrong or is there no guarantee we will hit 0 here (or in the `--r` case above)?

options.c
```diff
@@ -504,6 +506,54 @@ static void set_tree_width_max(void *data, const char *buf)
...
+		free(*pl_env_vars);
+		free(pl_env_vars);
+	}
+	if (!*buf) {
+		pl_env_vars = NULL;
+	}
+	size_t n = 1;
+	for (const char *x = buf; *x; x++)
+		if (*x == ',')
+			n++;
```
flyingmutant, line 546: I'd prefer braces here as well.

job.c
```diff
@@ -346,7 +347,9 @@ static int handle_line(void *data, const char *line)
 	if (is_http_url(line) || is_cue_url(line)) {
 		add_url(line);
 	} else {
-		char *absolute = path_absolute_cwd(line, data);
+		char *absolute = pl_env_var(line, NULL)
+			? pl_env_restore(line)
+			: path_absolute_cwd(line, data);
```
flyingmutant, line 352: Is it possible to avoid substituting conditional expressions, and just substitute functions which handle both cases internally? The way it is now looks not very clean -- and makes adding new places that load/save filenames a bit harder.

cache.c
pgaskin: Processed filename.

options.c
pgaskin: That's correct. `r` is only there to prevent the buffer from overflowing (most of the other setting functions also just cut it off). The for loop condition is what will prevent us from going past the end of the string, since it'll end when the string hits the `NULL` terminator.

job.c
pgaskin: Yes, it's possible, but I did choose the current way for a few reasons:

1. This function shouldn't be used arbitrarily, since it is only needed for filenames cmus will read again later. Other file paths should be the resolved ones. If I was to move this into another function, what would it be named to make this clear? 
2. There are currently only two places this will be needed (cache and job), and it's unlikely there will be more. And, each of these requires different path handling (one needs the filename to be made absolute based on the working directory, the other uses it as-is). The way they handle freeing is also different too. Adding a function which combines both adds unnecessary abstraction for little benefit IMO.
3. This keeps `pl_env.{c,h}` purely concerned with doing path manipulation rather than also deciding when it needs to be done.

If you still think it'll be cleaner, I can change it.

options.c
pgaskin: Will do.

gavtroy (commented):
Overall it looks reasonable. The use case isn't relevant to me and I'm not planning to test it.

pl_env.c
```diff
@@ -0,0 +1,259 @@
...
+ * removes './' path segments, simplifies '../' path segments (and returns NULL
+ * if it would result in going above the uppermost directory), and removes the
+ * trailing slash.
+ */
+static char *pl_env_norm(char *path)
+{
+#ifdef _WIN32
+	/* convert backslashes to slashes */
+	/* note: cmus uses forward slashes internally, but Windows accepts both */
+	for (char *p = new; *p; p++)
```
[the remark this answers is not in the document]
pgaskin: I'll fix the variable. It got lost during refactoring since it was behind an ifdef.

> By the way does WSL / Cygwin / whatever not give us a unix-style view of the filesystem?

The issue is if the user passes a path. The compatibility layers generally do accept both, but for `pl_env` it needs to match exactly, so we normalize it.

pl_env.c
```diff
@@ -0,0 +1,259 @@
...
+				if (path[r += 1]) r++;
+
+				continue;
+			}
+
+			/* handle '..' segments */
+			if (path[r] == '.' && path[r+1] == '.' && (path[r+2] == '/' || !path[r+2])) {
+
+				/* if there aren't any parent directories left to skip, return NULL */
+				if (!w) return NULL;
```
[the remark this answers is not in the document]
pgaskin: > Does this account for symlink traversal though

That's been intentionally left out. Remember that the user might, for example, symlink their music folder to an external disk. They'd still expect an environment variable including the symlink in the path to work correctly (and the rest of cmus doesn't normalize symlinks in the playlists).

pl_env.c
```diff
@@ -0,0 +1,259 @@
...
+				free(*--new);
+	*new = NULL;
+}
+
+char *pl_env_save(const char *path)
+{
+	if (!pl_env_vars || !*pl_env_vars || pl_env_var(path, NULL))
+		return xstrdup(path);
+
+	for (char **var = pl_env_vars; *var && **var; var++) {
```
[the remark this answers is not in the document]
pgaskin: Yes.

pl_env.c
```diff
@@ -0,0 +1,259 @@
...
+	strcpy(ptr, val);
+	ptr += val_len;
+	strcpy(ptr, rem);
+	ptr += rem_len;
+	*ptr = '\0';
+
+	return new;
+}
+
+const char *pl_env_var(const char *path, int *out_length)
```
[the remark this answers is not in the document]
pgaskin: I do it this way for efficiency and to allow it to be used as part of if statements.

player.c
```diff
@@ -662,7 +663,12 @@ static void _producer_play(void)
 
 		rc = ip_open(ip);
 		if (rc) {
-			player_ip_error(rc, "opening file `%s'", ip_get_filename(ip));
+			int var_len;
+			const char *fn = ip_get_filename(ip), *var;
+			if ((var = pl_env_var(fn, &var_len)))
+				player_ip_error(rc, "opening file `%s': missing env var `%.*s`", pl_env_var_remainder(fn, var_len), var_len, var);
```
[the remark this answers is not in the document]
pgaskin: > For clarity, maybe put it as opening file '<unknown>/%s': missing env

That's a good idea.

> will handle if the env var contains multibyte characters

It'll treat each byte as an individual character, so it'll work fine in this case.

Doc/cmus.txt
```diff
@@ -1666,6 +1673,45 @@ USER
...
+variables can be appended (delimited with commas) to `pl_env_vars` to redirect
+specific paths. The environment variables must contain absolute paths to be
+used. The last matching directory prefix will be used.
+
+If a file does not exist, it will be treated the same way as missing files
+usually are. If the file is in the cache, it will be preserved, and an error
+will be displayed upon playback. If the file is not in the cache, it will be
+removed from playlists and the library. There is one special exception: if a
+file is missing due to the environment variable being invalid/empty/unset, but
+it is still in the cache, it will be preserved as-is.
```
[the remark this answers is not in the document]
pgaskin: Usually, cmus will remove files if they aren't in the filesystem. I've added an exception to the default behaviour to keep the file if the entry with the env var placeholder is still in the cache. I'm not sure if there's a less technical way to explain this.

pl_env.c
```diff
@@ -0,0 +1,259 @@
...
+	while (path[r]) {
+
+		/* handle the start of a segment */
+		if (w == 0 || path[w-1] == '/') {
+
+			/* handle empty segments */
+			if (path[r] == '/') {
+
+				/* skip the duplicate slashes */
+				while (path[r] == '/') r++;
```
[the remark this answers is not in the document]
pgaskin: It gets optimized out, and this keeps the indentation consistent for the different conditions. I originally had it the way you describe, but I found it a bit harder to understand when coming back to the code later.

Doc/cmus.txt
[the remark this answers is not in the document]
gavtroy: I'm still not sure, but I think it could be that the first half of your description is referring to cmus startup and the last part is referring to `:update`? At least the normal behaviour is different for both: on startup, missing files are kept if they're in the cache, while for `:update` missing files are always removed. It would make sense if you've added the exception to `:update`. But I suspect I'm still not getting it. ;)

cache.c
pgaskin: Done: `proc_filename`.

options.c
pgaskin: Comment added.

job.c
pgaskin: @flyingmutant?

pl_env.c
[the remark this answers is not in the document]
pgaskin: Fixed.

pl_env.c
```diff
@@ -0,0 +1,259 @@
...
+		*out_length = (int) (end-path);
+	return path;
+}
+
+const char *pl_env_var_remainder(const char *path, int length)
+{
+	return path+length+2;
+}
+
+int pl_env_var_len(const char *path)
```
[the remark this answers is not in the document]
pgaskin: Removed.

pl_env.h
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * pl_env_restore returns a malloc'd copy of path, with the environment variable
+ * substitution. The provided path must use forward slashes, and begin with a
+ * slash or a substitution followed by a slash (which will always be true for
+ * library paths within cmus). If the path does not have a substitution, the
+ * original path is returned. If the environment variable does not exist or is
+ * invalid, the original path is also returned.
+ */
+char *pl_env_restore(const char *path);
```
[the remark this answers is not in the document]
pgaskin: Renamed to expand/reduce.

pgaskin:
Rebased.

options.c
pgaskin: Done.

Doc/cmus.txt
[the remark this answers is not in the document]
pgaskin: Updated.

gavtroy (approved):
Reviewed again, it's still fine. My only reservation is what we've already covered: that the use case is niche, and that it's hard to discover even if you do need it. I do wonder if it would have been better only to support HOME through a more discoverable option like "cross_platform_home" or "lib_relative_to_home" or whatever. Anyway I'm still fine with merging the current solution.

I ended up also briefly testing it to understand it better, and it seemed to work well for me.

flyingmutant:
@pgaskin thanks for the code, @gavtroy thanks for the review! I do agree about the approach being hacky, but it is clever, self-contained and solves a real problem for people. I think the documentation does a good job explaining the feature, so it should be relatively discoverable.

PR #1104, opened 2021-07-11
Replaces improper fcntl with ioctl in op/oss

### LINKED ISSUE #1103: op/oss: 'fcntl()' vs. 'ioctl()'
While debugging a port of cmus for the Genode OS framework I noticed that the `oss_reset()` function calls `fcntl()` with the `SNDCTL_DSP_RESET` command, which failed. On FreeBSD this call returned -1 as well, whereas it worked when using `ioctl()` instead.

https://github.com/cmus/cmus/blob/75ef523d5657cb37cc5a8f9869764faf2ede7348/op/oss.c#L44-L50

nefthy:
fixes #1103

### DIFF
--- a/op/oss.c
+++ b/op/oss.c
@@ -31,7 +31,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
-#include <fcntl.h>
 
 static sample_format_t oss_sf;
 static int oss_fd = -1;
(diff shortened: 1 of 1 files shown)


pgaskin (approved):
LGTM, but untested.

flyingmutant:
Thanks!

PR #1107, opened 2021-07-21
Fix segfault

vlada-dudr:
Hi,

I got segfault, because this mark_clipped_text function got negative buffer length. This is simple fix, but it works for me. Maybe there is a worse bug, which originates the negative buffer_length. I don't have knowledge about cmus's codebase to go hunting for it, so here is what I got.

### DIFF
--- a/format_print.c
+++ b/format_print.c
@@ -44,6 +44,9 @@ static struct gbuf* str = &l_str;
 
 size_t mark_clipped_text(char *buffer, int buf_len)
 {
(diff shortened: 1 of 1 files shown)


infmagic2047:
Hi, please have a look at #1117, which is (I believe) the same issue as yours.

vlada-dudr:
Yes it is indeed

infmagic2047:
I created #1119, which addresses the more fundamental issue of using uninitialized values :P

PR #1094, opened 2021-06-03
Initialize active to zero in print_editable

pgaskin:
The uninitialized variable, active, would sometimes cause cmus to segfault while accessing the pairs array based on its value.

This commit also fixes implementation-defined behaviour for the implicit cast of track->marked.

fixes #1093

Issue #1093, opened 2021-06-03
Segfault when switching to sorted library view since commit 58958db


FBI-Guy:
1) `uname -a`:
```
Linux user 5.12.8-ck1 #1 SMP PREEMPT Wed Jun 2 20:25:29 EDT 2021 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.9.1-2-g58958db
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/local/lib/cmus/ip
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  aac:
    Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  ffmpeg:
    Priority: 30
    File Types: aa aac ac3 aif aifc aiff ape au fla flac m4a m4b mka mkv mp+ mp2 mp3 mp4 mpc mpp ogg opus shn tak tta wav webm wma wv
    MIME Types:

Output Plugins: /usr/local/lib/cmus/op
  alsa
  oss
  jack
  sndio
  ao

```

Since commit 58958db I am getting a segfault when switching to the sorted library view by pressing 2

pgaskin:
Do you have a backtrack with debug symbols, or a core dump with the binaries?

FBI-Guy:
[cmus_debug.zip](https://github.com/cmus/cmus/files/6593002/cmus_debug.zip)
Hopefully this has what you're looking for inside. Let me know what I should do if it doesn't

Thank you

pgaskin:
Does this patch fix the issue?

```diff
diff --git a/ui_curses.c b/ui_curses.c
index 670a00b..148b4ef 100644
--- a/ui_curses.c
+++ b/ui_curses.c
[...]

PR #1051, opened 2021-01-24
Indicate elided text with "…" & related formatting tweaks

gavtroy:
Closes #637.
Closes #1004.

PR #1120, opened 2021-10-04
Handle colors in active windows correctly

infmagic2047:
Originally, there was the editable_active variable in ui_curses.c to
indicate whether the window is active. However, this variable was
removed in #1051, presumably by mistake. It also leads to the active
variable being uninitialized in print_editable, causing #1093.

The uninitialized variable issues was fixed in #1094, but unfortunately
in a wrong way: it simply assumes all windows are inactive, leading to
incorrect colors in many views.

This commit restores the old logic of editable_active (renamed to
win_active for consistency). Now the colors should be correct.

### DIFF
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -141,6 +141,7 @@ static int track_win_w = 0;
 
 static int win_x = 0;
 static int win_w = 0;
+static int win_active = 1;
 
 static int show_cursor;
 static int cursor_x;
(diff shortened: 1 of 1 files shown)


infmagic2047:
By the way, if anyone is looking for a test example, use the default colorscheme and switch to playlist view. Without this patch, the selected line in the right part of window (list of tracks) will always be white. With this patch the line will be blue when the window is active, and white otherwise.

nefthy:
Looks resonable

PR #1151, opened 2021-12-31
Fix missing combining marks on last character in a field

gavtroy:
### ~~Better tolerate emoji and unrecognized characters~~

~~Range is selected based on https://en.m.wikipedia.org/wiki/Emoji#Unicode_blocks~~

~~This improves the experience somewhat but it's not really a fix. For that we probably need a library or drop in code.~~


gavtroy:
Reduced the scope of this PR; no longer try to recognize emoji as fullwidth or unrecognized characters as halfwidth. Instead I support the proposal in #1198 as a more comprehensive fix.

Issue #997, opened 2020-10-07
Rendering issues with emojis or "zalgo" text


bersLucas:
---

1) `uname -a`:
```
Linux pop-os 5.4.0-7634-generic #38~1596560323~20.04~7719dbd-Ubuntu SMP Tue Aug 4 19:12:34 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors

```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  aac:
    Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  ffmpeg:
    Priority: 30
    File Types: aa aac ac3 aif aifc aiff ape au fla flac m4a m4b mka mkv mp+ mp2 mp3 mp4 mpc mpp ogg shn tak tta wav webm wma wv
    MIME Types:

Output Plugins: /usr/lib/cmus/op
  pulse
  alsa
  jack
  ao
```

Given the following album: https://dismissyourself.bandcamp.com/album/000-deluxe-edition

The special charecters in the track and artist names causes some rendering issues when hovering over the artist name in the library view. I'm using Bash 5.0.17(1)-release on the Gnome terminal

[...]

PR #1152, opened 2021-12-31
Fix under-allocation when many zero-width characters are present

gavtroy:
Previously we were allocating a "worst case" of 4 per number of printed
characters, which ignores combining characters (for example).

Noticed through a segfault when restarting cmus and immediately playing
the first track of the album discussed in #997.

Other places in the code (ui_curses) also don't account for zero-wide
characters when setting bounds but I'm not sure how to go about fixing
that.


gavtroy:
This PR is probably oversold by the current headline. To quote the updated commit message:

```
... this issue is very rare, almost theoretical, for a few reasons:
- While zero-width characters are not accounted for at all, many other
  characters are significantly over-budgeted, e.g. every Latin character
  is over-accounted by 3 bytes.
- Zero-width characters are relatively scarce, excepting "zalgo text".
- gbufs only grow and are reused for the life of the program
```
Instead this change can be considered refactoring that happens to fix rare crashes with zalgo text. If I had realized how niche the issue is I might not have bothered fixing it, but at least the refactoring ends up improving the oveall readability of the code.

flyingmutant:
A bit scary to merge a large refactoring like this, but less code is definitely better -- lets do it now and hope that people running `master` will help catch any possible bugs.

PR #1165, opened 2022-01-25
Recognize artist/album sort keys from ffmpeg plugin

### LINKED ISSUE #1164: Cmus handles ID3 Tag and MP4/AAC Tag differently when sorting by Artist
I noticed that all my Artists get sorted in a different way. Some are sorted by their last name and some are sorted by their first name. After some searching I noticed that all songs with ID3 Tags (*.mp3 files) are sorted correctly by their last name (for example Bakken, Rebekka under "B") and all files with MP4/AAC Tag are sorted by their first name (Rebekka Bakken unter "R"). 
Somehow Cmus seems to handle tags differently. Is that supposed to be the case? It does not make sense to me.

gavtroy:
These values are experimental. I've no explanation for why "TSO2" is
used rather than something like "album-artist-sort".

This is particularly relevant for mp4/m4a as the mp4 input plugin is
currently broken/disabled on many platforms.

Fixes #1164.

### DIFF
--- a/comment.c
+++ b/comment.c
@@ -232,6 +232,14 @@ static struct {
 	{ "sourcemedia", "media" },
 	{ "MusicBrainz Track Id", "musicbrainz_trackid" },
 	{ "version", "subtitle" },
+	/* ffmpeg id3 */
+	{ "artist-sort", "artistsort" },
+	{ "TSO2", "albumartistsort" },
+	{ "album-sort", "albumsort" },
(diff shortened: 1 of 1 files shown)


gavtroy:
It should be good, ignore my terrible wording. I just meant that the field names were discovered by experiment (i.e. looking at what ffmpeg was returning on some real files) rather than finding it documented anywhere.

Issue #1064, opened 2021-03-15
Quitting while playback is paused takes several seconds


Markus00000:
cmus exits instantly when playback is on while quitting:

1. `$ CMUS_HOME=/tmp/cmus /usr/bin/cmus`
2. Play MP3 file
3. `:q` exits cmus and `cmus` terminates almost instantly

Exiting cmus while playback is paused takes several seconds:

1. `$ CMUS_HOME=/tmp/cmus /usr/bin/cmus`
2. Play MP3 file
3. Pause playback
4. `:q` closes the cmus interface, but `cmus` will keep running for several seconds before terminating

This problem only appeared recently.

---

1) `uname -a`:
```
Linux lati 5.11.6-arch1-1 #1 SMP PREEMPT Thu, 11 Mar 2021 13:48:23 +0000 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.8.0
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  aac:
    Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  mp4:
    Priority: 50
    File Types: mp4 m4a m4b
    MIME Types:
  mpc:
    Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  ffmpeg:
    Priority: 30
[...]

PR #1172, opened 2022-02-16
Workaround for #1064

er888kh:
I noticed that from probably PipeWire 0.3.34 up to now, cmus started taking a long time before exiting when the music is paused (#1064), otherwise it has no issue. It seems like disabling `_pa_stream_drain` on PipeWire when exiting cmus fixes the issue and has no apparent setbacks. This seems to be an issue with upstream (as it was once fixed there), but given the simplicity of the workaround, we can have this until upstream fixes the issue.

### DIFF
--- a/op/pulse.c
+++ b/op/pulse.c
@@ -17,7 +17,9 @@
  */
 
 #include <string.h>
+#include <stdbool.h>
 
+#include <pulse/introspect.h>
 #include <pulse/pulseaudio.h>
 
 #include "../op.h"
@@ -32,6 +34,8 @@ static pa_stream		*pa_s;
 static pa_channel_map		 pa_cmap;
 static pa_cvolume		 pa_vol;
 static pa_sample_spec		 pa_ss;
+
+static bool 		 is_pipewire = false;
 
 static int			 mixer_notify_in;
 static int			 mixer_notify_out;
@@ -175,6 +179,20 @@ static void _pa_sink_input_info_cb(pa_context *c,
 	}
 }
 
+static void _pa_server_info_cb(pa_context *c,
+					const pa_server_info *i,
+					void *data)
+{
+	is_pipewire = false;
+	if (i) {
+		if (strstr(i->server_name, "PipeWire") != NULL) {
+			// server is PipeWire
+			d_print("Pulseaudio server is pipewire. Disabling _pa_stream_drain()\n");
+			is_pipewire = true;
+		}
+	}
+}
+
 static void _pa_stream_success_cb(pa_stream *s, int success, void *data)
 {
 	pa_threaded_mainloop_signal(pa_ml, 0);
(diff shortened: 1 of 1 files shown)


op/pulse.c
```diff
@@ -246,6 +264,10 @@ static int _pa_stream_cork(int pause_)
 
 static int _pa_stream_drain(void)
 {
+	if (is_pipewire) {
+		return OP_ERROR_SUCCESS;
+	}
+
```
gavtroy, line 270: It looks like this could be removed as you also guard calling the function.

op/pulse.c
```diff
@@ -33,6 +35,8 @@ static pa_channel_map		 pa_cmap;
 static pa_cvolume		 pa_vol;
 static pa_sample_spec		 pa_ss;
 
+static bool 		 is_pipewire = false;
+
```
gavtroy, line 39: The indentation is off since our tabs are 8-wide. There's a similar mismatch below in the alignment of `_pa_server_info_cb()`'s params.

gavtroy (commented):
Sure, since it's not obvious what purpose `_pa_stream_drain()` serves in the first place I wouldn't object to adding this workaround.

er888kh:
Thanks for you review @gavtroy. I'm really busy until Thursday. I will most certainly fix all of the issues you mentioned by the end of the week though.

philippemilink:
This patch works fine and I pushed it in the Debian cmus package.

op/pulse.c
```diff
@@ -175,6 +179,20 @@ static void _pa_sink_input_info_cb(pa_context *c,
...
 
+static void _pa_server_info_cb(pa_context *c,
+			       const pa_server_info *i,
+			       void *data)
+{
+	is_pipewire = false;
+	if (i) {
+		if (strstr(i->server_name, "PipeWire") != NULL) {
+			// server is PipeWire
+			d_print("Pulseaudio server is pipewire. Disabling _pa_stream_drain()\n");
```
flyingmutant, line 190: Is there any upstream issue we can reference here, so that we know when we can drop the workaround?

er888kh: Hi. I really can't investigate this issue right now. I live in Iran and ... nothing good is going on right now. A few weeks ago the police was in our university (Sharif University of Technology. Go look at its wikipedia page) shooting students and beating faculty members (yeah our own professors) and arresting people and ... I just need some time to deal with this shit.

primalmotion: take care :(

gavtroy: Hope all is well.

I don't think there is an existing upstream report other than the original [pipewire/-/issues/946](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/946), but the change that broke it again ([pipewire/-/commit/b9c8f0f6b9](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/b9c8f0f6b9dd479a0372adae514653509ff1d957)) says it's expected behavior:

> A corked stream should not complete a pending drain operation

I experimented with uncorking the stream first and it also solves the issue, but just removing the call makes more sense to me. It doesn't seem like we ever had any reason to wait for the stream to drain in the first place.

flyingmutant: So, let's go with #1268?

PR #1198, opened 2022-05-21
Use wcwidth for u_char_width

### LINKED ISSUE #1196: cmus does not properly detect width of Japanese characters
On cmus version 2.9.1-25-g54d7213 (current commit at the time of this issue's creation), cmus is not able to determine the width of Japanese characters accurately, resulting in broken right alignment:

image
image

Replacing the `u_char_width` implementation with a call to POSIX `wcwidth` using glibc 2.32.2 appears to fix the problem:

```
int u_char_width(uchar u) {
        return wcwidth(u);
}
```

image
image

Unfortunately, I think this does not actually solve the problem completely, as different libcs provide different implementations, and glibc's `wcwidth` was not up to date for a number of years IIRC. 

I'd be more than happy to submit a PR if necessary, but I have two concerns:

* There are other issues open relating to character spacing, and it's unclear to me if this code will be replaced in the near future regardless
* The wcwidth implementations in musl and glibc are both table based, so it's unclear to me if it would be preferred to fix this by switching `u_char_width` to a table based approach
* There is also the possibility of vendoring the musl implementation, which is relatively concise, and seems like it would require less maintenance

wooosh:
Resolves #1196.

Changes:

* Old `u_char_width` was removed because it produced results inconsistent with wcwidth for many codepoints.
* The new implementation of `u_char_width` uses `wcwidth`
* A vendored implementation of `wcwidth` has been added from the musl libc, and will be used in situations where `sizeof(wchar_t) != sizeof(uchar)` or `wcwidth` does not exist.

Things that need to be addressed before this is ready to merge:
* ~~Test that the configure script and musl `wcwidth` work on windows.~~ (see below)
* The original `u_char_width` implementation would always return width of 1 for non-utf8 locales, even if it was a control character. I'm not sure if that was done for a specific reason, or if it was an oversight that never lead to issues in practice. The new implementation handle control characters even for non-utf8 locales, but this change might need to be reverted.
* I have not yet tested this with non-CJK languages or terminals besides xterm and urxvt.


wooosh:
Nevermind about the first concern w/windows, I somehow got confused into thinking that cmus supported windows, and AFAICT, it does not. This might mean that the vendored implementation can be removed entirely, as windows is the only modern platform I'm aware that doesn't define `wchar_t` to be 4 bytes.

uchar.c
```diff
@@ -178,93 +179,24 @@ size_t u_strlen_safe(const char *str)
...
-narrow:
-	return 1;
-wide:
-	return 2;
+
+	w = wcwidth_uchar(u);
+	if (w >= 0)
+		return w;
+	else
+		return 2;
```
gavtroy, line 198: Assuming it's for the likes of "￾" would 1 not make more sense? (Assuming we ever get here.)

gavtroy (commented):
> The original u_char_width implementation would always return width of 1 for non-utf8 locales, even if it was a control character. I'm not sure if that was done for a specific reason, or if it was an oversight that never lead to issues in practice.

An oversight in https://github.com/cmus/cmus/pull/1051/commits/7e4ec7aeb94dc731f17c90b76f13617c3a8ea6e0, and yes I don't imagine control characters are encountered often.

> windows

I'm not sure to what level it's _supported_, but cmus can be compiled with cygwin. I don't have cygwin to test how it handles wchar_t but google suggests it's probably 2 bytes / UTF-16.

uchar.c
wooosh: That would be my assumption, but my reading of the original source for this function lead me to believe it returned 2 for unknown characters, for reasons that aren't really clear to me:

```
	/* ? */
	if (u >= 0x30000U && u <= 0x3fffdU)
		goto wide;
```

Maybe the intent was that overallocating space would be preferred to underallocating? I don't have a strong opinion either way, as I believe this should only be reached for private use characters, which is probably better set to a default width of 1, since they are unlikely to show up in song metadata, and if the user wants to use them in their status bar, they can manually pad it if need be.

wooosh: It's worth noting that it's not incredibly uncommon for `wcwidth` to be overriden with `LD_PRELOAD` to handle icon widths as well, for example https://github.com/powerman/wcwidth-icons.

I will change the default width to 1 unless there was a specific reason for the old behavior.

wooosh:
> I'm not sure to what level it's supported, but cmus can be compiled with cygwin. I don't have cygwin to test how it handles wchar_t but google suggests it's probably 2 bytes / UTF-16.

The current code in this PR should work on windows because it will not use `wchar_t` if it isn't 4 bytes, but I cannot actually verify that myself without access to a windows machine. I also don't know if the terminals available for windows actually handle anything beyond 16 bit codepoints to begin with, though.

uchar.c
gavtroy: I would have taken `/* ? */` to mean "I don't know what to call this range but it seems to be filled with wide characters".

At least [U+FFFE](https://util.unicode.org/UnicodeJsps/character.jsp?a=fffe&B1=Show) does hit this case for me and displays as only 1 column wide, so I'd go with 1.

> It's worth noting that it's not incredibly uncommon for wcwidth to be overriden with LD_PRELOAD to handle icon widths as well

Not something I would have thought of, but I suppose it shouldn't matter to us? If anything this change might let it work properly.

wooosh: This should be resolved by 9a9d878.

gavtroy (approved):
@pgaskin perhaps you can test with Cygwin at some point, although seems to me this should cover all bases.

wooosh:
This appears to work reliably on the zalgo text test album in xterm and xfce4-terminal (which is VTE based), in terms of creating correct padding, though the first track appears blank:

image

My terminal is capable of displaying the song title, so I do not believe it is a font issue:
image

It's unclear to me exactly what is happening here, a test program I've written to test the `u_char_width` function appears to produce "reasonable" results: 

<details>
<summary>u_char_width test</summary>
```
$ ./wcwidth "ẹ̶͌l̴͓ ̵̜r̶̹e̴͕ẗ̵̹o̷̯r̴̲n̶̗o̵̝ ̸̣d̴̢è̵̙l̴͗ ̸̻g̵̲a̸̰m̶̫ḙ̵̬r̴̬"
codept - width
0x1eb9 - 1
0x0336 - 0
0x034c - 0
0x006c - 1
0x0334 - 0
0x0353 - 0
0x0020 - 1
0x0335 - 0
0x031c - 0
0x0072 - 1
0x0336 - 0
0x0339 - 0
0x0065 - 1
0x0334 - 0
0x0355 - 0
0x1e97 - 1
0x0335 - 0
0x0339 - 0
0x006f - 1
0x0337 - 0
0x032f - 0
0x0072 - 1
0x0334 - 0
0x0332 - 0
0x006e - 1
0x0336 - 0
0x0317 - 0
0x006f - 1
0x0335 - 0
0x031d - 0
0x0020 - 1
0x0338 - 0
0x0323 - 0
0x0064 - 1
0x0334 - 0
0x0322 - 0
0x00e8 - 1
0x0335 - 0
0x0319 - 0
0x006c - 1
0x0334 - 0
0x0357 - 0
0x0020 - 1
0x0338 - 0
0x033b - 0
0x0067 - 1
0x0335 - 0
0x0332 - 0
0x0061 - 1
0x0338 - 0
0x0330 - 0
0x006d - 1
0x0336 - 0
0x032b - 0
0x1e19 - 1
0x0335 - 0
0x032c - 0
0x0072 - 1
0x0334 - 0
0x032c - 0
```

</details>

My guess is that it is somehow related to the large number of combining characters, and zero width somehow meaning that the combining characters overwrite the letters they modify, but I'm not entirely sure. I'll investigate further to try to see what the issue is, especially since it worked with the old code.

wooosh:
Furthermore, attaching that song title to another mp3 with audio and attempting to play it causes cmus to segfault.

gavtroy:
> the first track appears blank

Although this issue doesn't exist on the current release, it's already broken on master since a strcpy somewhere was replaced by a more intelligent function. This is actually fixed by #1151 although it's not clear why from the description (and I don't remember either, but this is the exact bug that prompted the fix :slightly_smiling_face:).

> segfault

It could be a side effect of a rare allocation issue fixed by #1152... or more likely it could be a separate issue as you're thinking. But it's probably worth cherry-picking that refactoring/fix before you dig too deep.

wooosh:
I've been using this branch for close to two weeks (with CJK, cyrllic, emojis, and zalgotext) and have been unable to reproduce the crash after cherry picking that fix, I believe this should be ready for review for merging.

flyingmutant:
@gavtroy am I correct that this should go in together with #1151 and #1152?

gavtroy:
#1151 could be merged first, since this PR was rebased on it.

#1152 isn't strictly related, other than fixing a rare crash that can be encountered when manually testing tracks with very "interesting" titles. (Although feel free to merge it at any time :slightly_smiling_face: )

flyingmutant:
Thanks a lot, finally merged!

PR #1263, opened 2023-07-02
Search should only try filename when there are no tags

### LINKED ISSUE #1262: Artist/Album search with *//* includes title tags in view 1
---

1) `uname -a`:
```
Linux 5.10.0-23-amd64 #1 SMP Debian 5.10.179-1 (2023-05-12) x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.9.1
```

3) `cmus --plugins`:
```
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    MIME Types:
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  ffmpeg:
    Priority: 30
    File Types: aa aac ac3 aif aifc aiff ape au fla flac m4a m4b mka mkv mp+ mp2 mp3 mp4 mpc mpp ogg opus shn tak tta wav webm wma wv
    MIME Types:

  pulse
  alsa
  jack
  ao
```

---------------------------------------------------------------------------------------

See man cmus:
SEARCHING: [...] Type //WORDS or ??WORDS **to search only artists/albums in view 1** and only titles in views 2-4. [...]

When I search with *//* in view 1 the results include title tags which according to the manual it should not.

Thank you

gavtroy:
Although this might be up for debate, but to me it seems like this was probably always the intended behavior.

Fixes #1262.

### DIFF
--- a/track_info.c
+++ b/track_info.c
@@ -188,8 +188,8 @@ int track_info_matches_full(const struct track_info *ti, const char *text,
 		const char *word = words[i];
 
 		matched = 0;
-		if (flags_set(ti, flags) && match_word(ti, word, flags)) {
-			matched = 1;
+		if (flags_set(ti, flags)) {
+			matched = match_word(ti, word, flags);
 		} else {
 			/* compare with url or filename without path */
 			const char *filename = ti->filename;
@@ -201,7 +201,7 @@ int track_info_matches_full(const struct track_info *ti, const char *text,
 				matched = 1;
 		}
 
-		if (flags_set(ti, exclude_flags) && match_word(ti, word, exclude_flags))
+		if (match_word(ti, word, exclude_flags))
 			matched = 0;
 
 		if (match_all_words ? !matched : matched)


pgaskin (approved):
Tested. Works as expected.

https://copr.fedorainfracloud.org/coprs/pgaskin/cmus-testing/build/6153648/
https://github.com/pgaskin/cmus/commits/83309568a563a3e8f789c9bc3cef2b43b3a80101
https://nightly.link/pgaskin/rpms/actions/runs/5495006170

---

> seems like this was probably always the intended behavior

Personally, I think this makes sense. @flyingmutant, thoughts?

track_info.c
```diff
@@ -201,7 +201,7 @@ int track_info_matches_full(const struct track_info *ti, const char *text,
 				matched = 1;
 		}
 
-		if (flags_set(ti, exclude_flags) && match_word(ti, word, exclude_flags))
+		if (match_word(ti, word, exclude_flags))
```
flyingmutant, line 204: Why don't we call `flags_set` here?

gavtroy: It's just not needed because `match_word()` does the same checks anyway. This line isn't fixing anything, I just thought it was clearer to understand.

flyingmutant: I think the same goes for call to `flags_set` above: they should either both be present, or both be absent for consistency. Probably `flags_set` was used as an optimization, to avoid any string comparisons, but I am not sure.

gavtroy: The first one is needed though, to distinguish "no applicable tags -- fallback to filename search" from "no match -- don't fall back to filename".

PR #1269, opened 2023-07-15
 Fixed function detection. check_function failed to compile due to clang -Wint-conversion warning became error.

VladislavGrudinin:
In recent clang version the -Wint-conversion warning became an error. So the function detection doesn't work anymore.
Compiling this code with clang
`char resizeterm(); int main(int argc, char *argv[]) { return resizeterm; }`
results in error:
`main.c:1:62: error: incompatible pointer to integer conversion returning 'char ()' from a function with result type 'int' [-Wint-conversion]`

### DIFF
--- a/scripts/checks.sh
+++ b/scripts/checks.sh
@@ -495,7 +495,7 @@ check_function()
 	__func="$1"
 	shift
 	msg_checking "for function $__func"
-	if try_compile_link "char $__func(); int main(int argc, char *argv[]) { return $__func; }" "$@"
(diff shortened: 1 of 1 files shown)


pgaskin (approved):
LGTM.

> The -Wint-conversion warning diagnostic for implicit int <-> pointer conversions now defaults to an error in all C language modes. It may be downgraded to a warning with -Wno-error=int-conversion, or disabled entirely with -Wno-int-conversion.

https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html

pgaskin:
@flyingmutant

PR #1272, opened 2023-07-16
Add mouse bindings for right third of status bar

pgaskin:
Default to play/volume.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -195,35 +195,38 @@ commands in the COMMANDS section.
 
 @h2 Common: Playback
 @pre
-b                       player-next
-c                       player-pause
-x                       player-play
-z                       player-prev
-v                       player-stop
+b                           player-next
+c                           player-pause
+x                           player-play
+z                           player-prev
+v                           player-stop
 
-B                       player-next-album
-Z                       player-prev-album
+B                           player-next-album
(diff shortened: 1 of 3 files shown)


pgaskin (commented):
Tested. Works as expected.

https://copr.fedorainfracloud.org/coprs/pgaskin/cmus-testing/build/6176496/
https://github.com/pgaskin/cmus/commits/001f1a4806e48c9f298e41d047e43d669c9475c8
https://nightly.link/pgaskin/rpms/actions/runs/5568630004

pgaskin:
Thoughts, @flyingmutant?

flyingmutant:
Thanks, will review later this week.

gavtroy (approved):
Looks good with the minor caveat that documentation doesn't mention that "_right" means "the rightmost third of the bar", and I'm not sure why we pick 33% rather than 50%. But no big deal.

To make it more intuitive, it might be nice to change the default `format_statusline` to move the volume indication to the right of the bar (and maybe add the missing % symbol). For instance it could look like -
image
But the PR does do what it says on the tin.

pgaskin:
I chose 33% since with 50%, I changed the volume by accident too frequently (especially since my mouse usually starts over the track list).

Changing the default position of the volume indicator would probably be a good idea for another PR.

PR #1274, opened 2023-07-21
Add %{albumtracks} format string

### LINKED ISSUE #850: Format string key for total album tracks
It'd be great to have a special format string key for the total tracks tag, to allow for something like

```
 Black Honey - Black Honey -  1 (of 21) - I Only Hurt The Ones I Love
```

gavtroy:
Total (counted) number of tracks in album. Also works as %N.

Closes #850.

For the example in that ticket, it would be `:set format_current= %a - %l%! - %n%{?N? (of %N)} - %t%= %y `.

### DIFF
--- a/tree.c
+++ b/tree.c
@@ -555,6 +555,7 @@ static struct album *album_new(struct artist *artist, const char *name,
 	album->min_date = date;
 	rb_root_init(&album->track_root);
 	album->artist = artist;
+	album->num_tracks = 0;
 
 	return album;
 }
@@ -867,6 +868,9 @@ static void album_add_track(struct album *album, struct tree_track *track)
 
 	rb_link_node(&track->tree_node, parent, new);
 	rb_insert_color(&track->tree_node, &album->track_root);
+
+	album->num_tracks++;
+	tree_track_info(track)->lib_album = album;
 }
 
 const char *tree_artist_name(const struct track_info* ti)
@@ -1177,6 +1181,9 @@ static void remove_track(struct tree_track *track)
 		window_row_vanishes(lib_track_win, (struct iter *)&iter);
 	}
 	rb_erase(&track->tree_node, &track->album->track_root);
+
+	track->album->num_tracks--;
+	tree_track_info(track)->lib_album = NULL;
 }
 
 void tree_remove(struct tree_track *track,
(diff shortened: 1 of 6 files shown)


pgaskin (approved):

pgaskin:
@flyingmutant

tree.c
```diff
@@ -870,6 +879,9 @@ static void album_add_track(struct album *album, struct tree_track *track)
 
 	rb_link_node(&track->tree_node, parent, new);
 	rb_insert_color(&track->tree_node, &album->track_root);
+
+	album->num_tracks++;
+	album_for_each_track(album, ti_set_albumtracks, &album->num_tracks, 0);
```
flyingmutant, line 884: Isn't this an O(N^2) algorithm?

pgaskin: Yes, it would be. The alternative would be to make the field itself a pointer, but I'm not sure if it's worth it since I doubt we'll be having hundreds of tracks per album.

Another option would be to modify the RB tree to include a count of tree children, but I doubt I'm not sure it's worth doing that.

The last option would be to make formatting support getting fields from a pointer to another struct (the album's num_tracks field), but that should probably be done separately (in another commit or PR).

gavtroy: Since I understand the code a bit better after this draft, I can try it.

gavtroy:
Updated to hold a reference to album in track info.

I removed the sort key I had in the first version because it didn't really work on the fly (i.e. when the first track is added in the list, it's sorted in albumtracks == 1 position). On reflection, it's not useful anyway.

One quirk I've found is that filters will lower the album tracks count when they filter out other tracks from an album. Unfortunately there doesn't seem to be any easy solution that doesn't have performance concerns. The right fix would be to rework filters to only hide tracks rather than temporarily removing them, but that's beyond the scope of this change.

pgaskin:
> filters will lower the album tracks count

In some ways, that actually makes sense (you're counting the tracks you can see in the album), but yes, it would be good to refactor it so it can be chosen deliberately rather than being a side-effect of how cmus works.

pgaskin (approved):
Tested. Works as expected.

https://copr.fedorainfracloud.org/coprs/pgaskin/cmus-testing/build/6204539/
https://github.com/pgaskin/cmus/commits/def777cecad32711b30da1f0b3b43a9d6481d19e
https://nightly.link/pgaskin/rpms/actions/runs/5633236513

flyingmutant:
For some reason, the fact that this is the first field with (relatively) complex lifetime rules in `track_info` makes me uneasy about merging this. Tracks being oblivious of where they come from seems like a much simpler model.

gavtroy:
Closing to avoid cluttering the PR list.

PR #1275, opened 2023-07-21
Flag to prevent cycling with left-view / right-view

### LINKED ISSUE #1241: preventing cycling with left-view / right-view
Is it possible to disable cycling when using left-view and right-view command ?

i.e. preventing moving to first view when you are on last view and preventing moving to last view when on first view.

gavtroy:
Added to the default mouse scroll binds, which is quite intuitive and seems to be how web browsers like to handle it. (For reference they do allow cycling when activated by keyboard shortcut, although we don't have any default keyboard binds for left-view / right-view.)

Closes #1241.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -543,11 +543,18 @@ player-stop (*v*)
 prev-view
 	Goes to the previously used view.
 
-left-view
+left-view [-n]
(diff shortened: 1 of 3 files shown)


pgaskin (approved):

pgaskin:
@flyingmutant

PR #1276, opened 2023-07-21
Add option to sort albums by name in tree view

### LINKED ISSUE #909: View 1's album sort order doesn't suit
[I apologise in advance for possible duplication: I added the below as a comment on a different issue, before realising it had been closed].

I would really like the library view to not sort albums by date: it would be really helpful to be able to actually configure the precise ordering, instead of it being hard-coded. It's easy enough to alter the sort order for View 2 ("Sorted Library") to remove reference to date, for example. Something similar for view 1 would be highly desirable.

For example:

Screenshot from 2019-05-31 18-51-58

The fact that my recording of Berg's violin concerto is quite an early one shouldn't mandate that it appears at the top of the list! Violin concertos should be somewhere down the end of the list, and 'Adagio' should be at the top. For me, obviously, not everyone... but the basic point is that the album ordering on view 1 should (I feel) be amendable by each user as the mood takes him/her -and, in my specific case, date of recording shouldn't come into it at all.

Being rather new to cmus, it is entirely possible I'm using it wrong ...and if my album sorting order problem can be resolved by other means, I'd really appreciate knowing how!

Cheers.

gavtroy:
By default they are still sorted by date, which is the current behavior.

I considered to also add an equivalent of smart_artist_sort, to strip "The" from album names when sorting. It could be nice to add for consistency, but I find it's not that useful when the list of albums under a given artist tends to be quite small anyway.

Closes #909.

### DIFF
--- a/tree.c
+++ b/tree.c
@@ -828,14 +828,11 @@ static void add_album(struct album *album)
 	struct rb_node **new = &(album->artist->album_root.rb_node), *parent = NULL;
 	struct album *found;
 
-	/*
-	 * Sort regular albums by date, but sort compilations
-	 * alphabetically.
-	 */
-	found = do_find_album(album,
-			      album->artist->is_compilation ? special_album_cmp
-							    : special_album_cmp_date,
-			      &new, &parent);
+	if (sort_albums_by_name || album->artist->is_compilation)
+		found = do_find_album(album, special_album_cmp, &new, &parent);
+	else
+		found = do_find_album(album, special_album_cmp_date, &new, &parent);
+
 	if (!found) {
 		rb_link_node(&album->tree_node, parent, new);
 		rb_insert_color(&album->tree_node, &album->artist->album_root);
@@ -1240,6 +1237,8 @@ void tree_sort_artists(void (*add_album_cb)(struct album *),
 			struct rb_node *t_node, *t_tmp;
 			struct album *album = to_album(l_node);
 
+			remove_album(album);
+			add_album(album);
 			rb_for_each_safe(t_node, t_tmp, &album->track_root) {
 				struct tree_track *track = to_tree_track(t_node);
 
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -1089,6 +1089,9 @@ smart_artist_sort (true)
 	names, preventing artists starting with "The" from clumping together.
 	Real `artistsort` tags override this option, if present.
 
+sort_albums_by_name (false)
+	In tree view (1), albums will be sorted by name rather than date.
+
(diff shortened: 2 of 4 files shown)


pgaskin (approved):

pgaskin:
@flyingmutant

tree.c
```diff
@@ -668,9 +668,11 @@ static int special_album_cmp_date(const struct album *a, const struct album *b)
 	if (cmp)
 		return cmp;
 
-	cmp = a->date - b->date;
-	if (cmp)
-		return cmp;
+	if (!sort_albums_by_name) {
```
flyingmutant, line 671: I'd say this is questionable: function is called `special_album_cmp_date`. PR should aim to make the code better after adding the new feature, not for the minimal possible diff.

pgaskin: True; this could be done in add_album, which already decides to do alphabetical sort for compilations, I think. Missed this while skimming the diff.

gavtroy: I had actually done it properly in a previous draft. :facepalm:  I'm not sure what caused me to think this would be better, but originally I was calling the setting `sort-albums-by-date=0` so it stood out less then at least.

tree.c
```diff
@@ -1240,6 +1237,8 @@ void tree_sort_artists(void (*add_album_cb)(struct album *),
 			struct rb_node *t_node, *t_tmp;
 			struct album *album = to_album(l_node);
 
+			remove_album(album);
+			add_album(album);
```
pgaskin, line 1241: I'm not sure how much I like this, but it's okay, I think.

pgaskin (approved):
Tested. Works as expected.

https://copr.fedorainfracloud.org/coprs/pgaskin/cmus-testing/build/6204539/
https://github.com/pgaskin/cmus/commits/def777cecad32711b30da1f0b3b43a9d6481d19e
https://nightly.link/pgaskin/rpms/actions/runs/5633236513

tree.c
flyingmutant: Why is this required?

gavtroy: The artists / albums / tracks are originally inserted into their respective list in sorted order based on whatever the current sorting rules are. This function is only used to force a resort when the "smart_artist_sort" and "sort_albums_by_name" settings are changed.

The remove function just unlinks the album from the artist's album tree, and the add function just links it back in sorted position. It doesn't involve album_new()/album_free() like you might have expected from the name.

As for removing/inserting from inside the `rb_for_each_safe` loop -- although some elements may be visited twice if they are moved forwards in the rb_tree, all elements will be processed at least once and at most twice, thanks to the `_safe` part.

PR #1281, opened 2023-07-31
Check only title in track title replacement

mrHeavenli:
We only need to check if title is null to replace it with the filename in the track info.

### DIFF
--- a/mpris.c
+++ b/mpris.c
@@ -30,6 +30,7 @@
 #include "track_info.h"
 #include "utils.h"
 #include "uchar.h"
+#include "path.h"
 
 #define CK(v) \
 do { \
@@ -392,6 +393,11 @@ static int mpris_metadata(sd_bus *_bus, const char *_path,
 			u_to_utf8(corrected, ti->title);
 			CK(mpris_msg_append_ss_dict(reply,
 					"xesam:title", corrected));
+		} else {
+			char corrected[u_str_print_size(path_basename(ti->filename))];
+			u_to_utf8(corrected, path_basename(ti->filename));
+			CK(mpris_msg_append_ss_dict(reply,
+				"xesam:title", corrected));
 		}
 		if (ti->album) {
 			char corrected[u_str_print_size(ti->album)];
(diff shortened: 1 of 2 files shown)


mrHeavenli:
@flyingmutant 
I see Mr. Gaskin is very busy, can you review this?

flyingmutant:
Can you please explain why is this change required?

gavtroy:
Normally if a track doesn't have tags then `altformat_*` are used, which by default either show the full path (`%f`) or the basename (`%F`) depending on context. This breaks with your change, and I'm not sure what it attempts to fix in return. If you just want the track window formatting to align you just need to match `altformat_trackwin` to `format_trackwin` and match `altformat_playlist` to `format_playlist`, except replacing `%t` with `%F`.

mrHeavenli:
> Can you please explain why is this change required?

Sure. When I was doing something with playerctl, tracks that had no title tag just gave nothing. So I looked at the code and found that the title would be filled in with the filename if there was no title tag, but that didn't show in the metadata that playerctl gave me. So I deleted that bit in the PR, and it actually gave me the filename in the metadata. I didn't really understand what the condition was there for, but now @gavtroy is telling me that it breaks something when removed. I guess this could also be done in mpris.c without breaking something?
by adding something in this region:
```c
if (ti->title) {
			char corrected[u_str_print_size(ti->title)];
			u_to_utf8(corrected, ti->title);
			CK(mpris_msg_append_ss_dict(reply,
					"xesam:title", corrected));
		}

```

mrHeavenli:
@gavtroy What do you think of this?

mpris.c
```c
#include "path.h"
...
if (ti->title) {
	char corrected[u_str_print_size(ti->title)];
	u_to_utf8(corrected, ti->title);
	CK(mpris_msg_append_ss_dict(reply,
			"xesam:title", corrected));
}else{
	char corrected[u_str_print_size(path_basename(ti->filename))];
	u_to_utf8(corrected, path_basename(ti->filename));
	CK(mpris_msg_append_ss_dict(reply,
			"xesam:title", corrected));
}
```

gavtroy:
With spaces around the else tidied up, looks good to me.

mrHeavenli:
@flyingmutant now gavtroy approves, can this be put in?

PR #1288, opened 2023-08-13
Fix ignore_duplicates missing filtered tracks

gavtroy:
Really just checking for duplicates again when unfiltering. A bit messy but it does the job.

The "entryp" naming is consistent with `hash_remove()` etc.

Relates to #423.

### DIFF
--- a/lib.c
+++ b/lib.c
@@ -631,7 +631,7 @@ static void hash_add_to_views(void)
 		while (e) {
 			struct track_info *ti = e->ti;
 
-			if (!is_filtered(ti))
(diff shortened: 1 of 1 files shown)


flyingmutant:
This makes code and intent really hard to follow: we are modifying hash (in a non-obvious way) inside a function called `hash_add_to_views`. I'd say some further work on this is required.

gavtroy:
Then for simplicity we can leave the duplicates in the hash and just skip adding them to the views. The only slight downside is future `hash_add_to_views` invocations (i.e. clearing a filter) could be slightly slower with a lot of duplicates... not something that matters in practice.

PR #1289, opened 2023-08-21
Fix gbuf to account for print width of control characters

gavtroy:
Such characters are stored as one byte but print as "<07>" for example.

(Although I'd argue we should just print them as '?' and save all the extra trouble.)

### DIFF
--- a/gbuf.c
+++ b/gbuf.c
@@ -109,10 +109,11 @@ static int gbuf_mark_clipped_text(struct gbuf *buf)
 
 void gbuf_add_ustr(struct gbuf *buf, const char *src, int *width)
 {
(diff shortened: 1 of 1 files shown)


flyingmutant:
I've started to look into this and got lost a bit. `u_str_print_size` uses `u_print_size` which is `u_char_size` + control exception, however `u_copy_chars` uses `u_char_width` inside; this looks inconsistent.

gavtroy:
`u_char_width` is how many columns of screen space are needed, it's not relevant to buffer usage. The `u_print_size` is to match the amount of buffer needed by `u_set_char`.

PR #1290, opened 2023-08-21
Fix formatting of whole number doubles

gavtroy:
Print for example "5.0" instead of "5." with an awkward trailing decimal point.

This formatter is only used for printing replay gain info (rg_track_gain etc.) which by default we don't.


flyingmutant:
Isn't just `%f` enough to get this?

gavtroy:
If you meant `%g` then the short answer is yes.

PR #1294, opened 2023-09-16
Add %{totaldiscs} to available format string

mathieu-lemay:
Allows displaying the number of discs of an album. Can also be used as a condition. Shorthand is `%T`.

I added it because I wanted to display the disc number only if the total number of discs is greater than one.


flyingmutant:
Thanks!

Issue #1066, opened 2021-03-26
Managing playlists externally while cmus is running


sebw:
I want to rate my music with Ulauncher while cmus plays music, but I'm facing an issue.

1. I start cmus
2. I have 5 empty playlists (rated_1_stars, rated_2_stars, and so on)
3. I start playing music
4. In Ulauncher I type "rate 4" which triggers a script which adds the current song in the rated_4_stars playlist
5. I quit cmus
6. Playlists are emptied

Rating script is inspired from https://gist.github.com/Markus00000/ad8c0ff46290f9dc2887

Why are playlists reverted back to the start they were when cmust started. What am I doing wrong?

pgaskin:
cmus loads the playlist on startup and saves it on exit. The proper way to do this would be with `cmus-remote`, but I don't think there's currently a command to add the current track to an arbitrary playlist without selecting it first.

cubicvoid:
https://github.com/cmus/cmus/pull/1297 made progress on this recently but mentioned "design discussion" that needs to happen for the remaining work... would a PR adding a `pl-add-playing` command (to add the currently-playing song or album to a playlist) be welcome? If so, are there any constraints I should know before I try adding it?

rodenet-hier:
> #1297 made progress on this recently but mentioned "design discussion" that needs to happen for the remaining work... would a PR adding a `pl-add-playing` command (to add the currently-playing song or album to a playlist) be welcome? If so, are there any constraints I should know before I try adding it?

That should be another low-hanging fruit in terms of design and implementation. I'd suggest building a set of commands around marking. `mark-playing` and `mark-pl <name>`/`pl-mark <name>` should be enough to implement `pl-add-playing` if combined with `win-add-p` for a start.

cubicvoid:
> I'd suggest building a set of commands around marking. `mark-playing` and `mark-pl <name>/pl-mark <name>` should be enough to implement `pl-add-playing` if combined with `win-add-p` for a start.

`mark-playing`would interact confusingly with `win-add-p` since the music that's currently playing isn't necessarily markable in the current window. (Is there a clean way to resolve that? Even restricting to the sorted view isn't enough, since the currently playing song doesn't need to actually be in the library.)

[...]

PR #1297, opened 2023-09-30
Add commands for external playlist management

rodenet-hier:
This PR implements the commands "pl-delete" and "pl-delete-all" that remove playlists for use with cmus-remote.

Other playlist management functionality needs further design discussion but these commands should be sufficient if all playlists are managed externally (which is my use case).

Partially addresses #1066

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -516,6 +516,9 @@ pl-import [filename]
 pl-rename <name>
 	Renames the selected playlist.
 
+pl-delete <name>
+	Deletes the playlist with the given name.
+
 pl-delete-all
 	Deletes all playlists.
 
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -516,6 +516,9 @@ pl-import [filename]
 pl-rename <name>
 	Renames the selected playlist.
 
+pl-delete-all
+	Deletes all playlists.
+
 player-next (*b*)
 	Skips to the next track.
 
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -505,6 +505,12 @@ mute
 pl-create <name>
 	Creates a new playlist.
 
+pl-delete [-a] <name>
+	Deletes the playlist with the given name.
+
+	-a
+		Deletes all playlists
+
 pl-export <filename>
 	Exports the currently selected playlist. The file will be overwritten if
 	it exists.
@@ -515,12 +521,6 @@ pl-import [filename]
 
 pl-rename <name>
 	Renames the selected playlist.
-
-pl-delete [-a] <name>
-	Deletes the playlist with the given name.
-
-	-a
-		Deletes all playlists
 
 player-next (*b*)
 	Skips to the next track.
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -516,11 +516,11 @@ pl-import [filename]
 pl-rename <name>
 	Renames the selected playlist.
 
-pl-delete <name>
+pl-delete [-a] <name>
 	Deletes the playlist with the given name.
 
-pl-delete-all
-	Deletes all playlists.
+	-a
+		Deletes all playlists
 
 player-next (*b*)
 	Skips to the next track.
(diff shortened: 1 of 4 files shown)


pl.c
```diff
@@ -459,6 +452,35 @@ static void pl_delete_selected_pl(void)
...
+	struct playlist *temp;
+	struct playlist *last = NULL;
+	list_for_each_entry_safe(pl, temp, &pl_head, node) {
+		if (list_len(&pl_head) == 1) {
+			last = pl;
+			break;
+		}
+		pl_delete(pl);
+	}
+	pl_create_default();
```
gavtroy, line 480: It would be cleaner to do this inside `pl_delete` where it checks if it's the last playlist. And at that point we could also remove the "cannot delete the last playlist" part of `pl_delete_selected_pl`.

command_mode.c
```diff
@@ -2636,6 +2646,8 @@ struct command commands[] = {
 	{ "pl-export",             cmd_pl_export,        1, -1, NULL,                 0, 0          },
 	{ "pl-import",             cmd_pl_import,        0, -1, NULL,                 0, 0          },
 	{ "pl-rename",             cmd_pl_rename,        1, -1, NULL,                 0, 0          },
+	{ "pl-delete",             cmd_pl_delete,        1, 1,  NULL,                 0, 0          },
+	{ "pl-delete-all",         cmd_pl_delete_all,    0, 0,  NULL,                 0, 0          },
```
gavtroy, line 2650: I suggest "pl-delete -a" instead but I don't feel too strongly either way. Other commands already use flags so it wouldn't be a heavy lift.

rodenet-hier: 3a8b7b5

pl.c
rodenet-hier: 68cf608

rodenet-hier:
All done

Doc/cmus.txt
```diff
@@ -516,6 +516,12 @@ pl-import [filename]
 pl-rename <name>
 	Renames the selected playlist.
 
+pl-delete [-a] <name>
+	Deletes the playlist with the given name.
+
+	-a
+		Deletes all playlists
```
gavtroy, line 523: \+ just noticed pl-delete isn't keeping to alphabetical order

pl.c
```diff
@@ -459,6 +452,41 @@ static void pl_delete_selected_pl(void)
...
+void pl_delete_all(void)
+{
+	struct playlist *pl;
+	struct playlist *temp;
+	struct playlist *last = NULL;
+	list_for_each_entry_safe(pl, temp, &pl_head, node) {
+		if (list_len(&pl_head) == 1) {
+			last = pl;
+			break;
+		}
```
gavtroy, line 484: Can this special case for the last item be removed now?

Doc/cmus.txt
rodenet-hier: 3a4ca11

pl.c
rodenet-hier: I've tested the code without that block and it was working but I'm not entirely sure how the list functions with additions inside the loop so I didn't want to risk an infinite loop. I could remove it if it's fine. I removed the "last" variable with 68d9b0b for now.

pl.c
```diff
@@ -459,6 +452,38 @@ static void pl_delete_selected_pl(void)
...
+	error_msg("couldn't find a playlist named '%s' to delete", name);
+}
+
+void pl_delete_all(void)
+{
+	struct playlist *pl;
+	struct playlist *temp;
+	list_for_each_entry_safe(pl, temp, &pl_head, node) {
+		if (list_len(&pl_head) == 1)
+			break;
```
gavtroy, line 481: Still not getting this, won't the for_each loop just end naturally when all the items are gone?

gavtroy: Wait I get it, because we add back the default :grin:

gavtroy (approved):

pl.c
rodenet-hier: Yeah it could still be fine but I couldn't really wrap my head around the implementation too well to confirm it.

flyingmutant:
Thanks!

rodenet-hier:
Thank you both as well for your time

PR #1303, opened 2023-11-25
Fix library detection on Apple Silicon

cubicvoid:
On Apple Silicon, Homebrew doesn't install ARM packages to the global header/library paths. This means that the common package dependencies `faad2` and `mp4v2` can no longer be detected by checking `/usr/local`, so `configure` fails to find them and disables AAC and MP4.

This is fixed by using `pkg-config` to check for those dependencies, since Homebrew's `pkg-config` search path accounts for the CPU architecture, and gives the correct results on both Intel and ARM.

I left the previous behavior in place as a fallback, for systems that have the libs installed globally without a working `pkg-config`.

Minor housekeeping: I also changed the temp file cleanup at the end to use `-r` with `rm`, since on macOS the configure script generates `.dSYM` directories from its temp file builds, and the script was printing errors every time it failed to delete them.

### DIFF
--- a/configure
+++ b/configure
@@ -450,7 +450,7 @@ check_roar()
 
 check_mp4()
 {
-	pkg_config MP4 "mp4v2 faad2" "" "-lmp4v2 -lfaad -lm"
+	pkg_config MP4 "mp4v2 faad2" "" "-lmp4v2 -lfaad -lm" || return $?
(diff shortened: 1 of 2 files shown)


gavtroy (approved):
looks fine to me, and still works fine on my (Linux) system

PR #1309, opened 2024-02-09
add play_queue total_time

Lagrang3:
The format option "total" will display the total time in the library or the current playing list, but if the queue is active it should instead display the total time of the queue.

### DIFF
--- a/play_queue.c
+++ b/play_queue.c
@@ -83,3 +83,8 @@ int play_queue_for_each(int (*cb)(void *data, struct track_info *ti),
 	}
 	return rc;
 }
+
+unsigned int play_queue_total_time(void)
+{
+	return pq_editable.total_time;
+}
--- a/play_queue.c
+++ b/play_queue.c
@@ -93,3 +93,8 @@ int queue_needs_redraw(void)
 {
 	return pq_editable.shared->win->changed;
 }
+
+void queue_post_update(void)
+{
+	pq_editable.shared->win->changed = 0;
+}
--- a/play_queue.c
+++ b/play_queue.c
@@ -88,3 +88,8 @@ unsigned int play_queue_total_time(void)
 {
 	return pq_editable.total_time;
 }
+
+int queue_needs_redraw(void)
+{
+	return pq_editable.shared->win->changed;
+}
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -641,8 +641,12 @@ const struct format_option *get_global_fopts(void)
 	int buffer_fill, vol, vol_left, vol_right;
 	int duration = -1;
 
-	fopt_set_time(&track_fopts[TF_TOTAL], play_library ? lib_editable.total_time :
-			pl_playing_total_time(), 0);
+	unsigned int total_time = pl_playing_total_time();
+	if (cmus_queue_active())
+		total_time = play_queue_total_time();
+	else if (play_library)
+		total_time = lib_editable.total_time;
+	fopt_set_time(&track_fopts[TF_TOTAL], total_time, 0);
 
 	fopt_set_str(&track_fopts[TF_FOLLOW], follow_strs[follow]);
 	fopt_set_str(&track_fopts[TF_REPEAT], repeat_strs[repeat]);
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -2040,6 +2040,9 @@ static void update(void)
 			do_update_commandline();
 		post_update();
 	}
+
+	/* Reset changed flags */
+	queue_post_update();
 }
 
 static void handle_ch(uchar ch)
--- a/ui_curses.c
+++ b/ui_curses.c
@@ -2015,7 +2015,9 @@ static void update(void)
 	}
 
 	/* total time changed? */
-	if (play_library) {
+	if (cmus_queue_active()) {
+		needs_status_update += queue_needs_redraw();
+	} else if (play_library) {
 		needs_status_update += lib_editable.shared->win->changed;
 		lib_editable.shared->win->changed = 0;
 	} else {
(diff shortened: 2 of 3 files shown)


gavtroy:
It could benefit from further consideration in `update()`, otherwise the code looks fine. But I wonder if it would be better to add it as a separate format option instead, if what you're really interested in is how low the queue is running.

I'm playing around with how we present the information in #1317 but a `%{queue_total}` or similar could be a complimentary change (and perhaps a `%{queue_nr_tracks}`).

Lagrang3:
If I add an explicit `queue_total` option, would I need to format my status line with a condition, kind of 
*if playing queue then show queue_total else show total*?

gavtroy:
> If I am on the queue

Incidentally since the queue isn't "active" until the first track gets pulled from it, it might make more sense to set that condition as (!queue_empty || queue_active).

The queue feels different to library vs playlist because it's a temporary state and there isn't a good indication of whether it's "active". But I'm not opposing your proposal, just wondering what works best for the use case. What you're suggesting is probably fine. If it was a separate format option then depending on the implementation you could have something like
`... %{total}%{?queue_total? (%{queue_total} enqueued)} ...` or
`... %{?queue_total?%{queue_total}?%{total}} ...`.


Either way if you expect it to be used in the statusline, you need to extend the redraw condition in `update()` https://github.com/cmus/cmus/blob/23afab39902d3d97c47697196b07581305337529/ui_curses.c#L1966-L1972

Lagrang3:
I think this is meant to indicate when the queue is active
https://github.com/cmus/cmus/blob/23afab39902d3d97c47697196b07581305337529/cmus.c#L547

I am using this change already locally, and it feels fine to me.

The only issue that I have, now that you mention it, is that the queue is *active* whenever there is at least one element there no matter if I am currently a track from a playlist or a the library; the next track will always be the one on the top of the queue. But that is not up to me.

Lagrang3:
At the end of the day I just want to extend the meaning of `{total}`. If I play from the library `total` will show the total time on the library, if I play from a playlist `total` will show the total time on the playlist. Therefore, one might expect that if I play from the queue `total` would show the total time on the queue. 
That's all I want to do.

gavtroy:
The logical behavior is subjective (and I'm not sure how useful the existing `%{total}` is in the first place...) but I don't think anyone would be upset by this change.

> I am using this change already locally, and it feels fine to me.

With regards to what I'm saying about `update()`, try adding something to the queue while cmus is paused/stopped. You'll see the statusbar isn't redrawn until it has a reason to be. If cmus is playing then it will redraw every second, so you might not notice this quirk.

Lagrang3:
True, I haven't noticed the update condition for the status bar. Now added a quick fix 8ea3b84.
It was fun reviewing the `update()` function. Thanks for the suggestion.

ui_curses.c
```diff
@@ -1964,7 +1968,9 @@ static void update(void)
 	}
 
 	/* total time changed? */
-	if (play_library) {
+	if (cmus_queue_active()) {
+		needs_status_update += queue_needs_redraw();
```
gavtroy, line 1972: You need to clear `changed` to 0 when you read it similar to what's done for lib_editable, otherwise it'll still be set on every loop.

Lagrang3: There is the same problem with the playlist view then. `pl_needs_redraw` checks on `pl_editable_shared.win->changed`, but `changed` is only set to 0 after `pl_draw` is called. So that means the status bar redraws on every loop if playing from the playlist but `pl_draw` is not called, eg. if on another view.

Lagrang3: I was thinking to set all `changed` flags to 0 in `post_update`. Would that make sense? I will propose it for the play queue, and later it could be also extended to the playlists and library.

Lagrang3: I did a check, and it wouldn't work resetting the `changed` flag in `post_update()`. 
I would get a frozen queue while on the queue view, for example after typing `:clear`.
I am not sure why. 

Instead I clear the flag at the end of the `update()` function and that seems to work.

gavtroy: True. I hadn't noticed that about playlist updates, but I guess you can just ignore it if you want.

> I would get a frozen queue while on the queue view

The queue view shouldn't freeze since it doesn't use `changed` as far as I see. But you probably meant playlist view since pl_draw() does check `changed` (although this optimization is probably not worth the headache).

gavtroy (approved):

pgaskin:
How does this change relate to bad9b0c4291ad6a5e9f5de0e14a533ebd661ef41 from #1317 @gavtroy?

Unless I'm misunderstanding the intention of this PR, it's now obsolete, right?

play_queue.c
```diff
@@ -79,3 +79,18 @@ int play_queue_for_each(int (*cb)(void *data, struct track_info *ti),
...
+}
+
+int queue_needs_redraw(void)
+{
+	return pq_editable.shared->win->changed;
+}
+
+void queue_post_update(void)
+{
+	pq_editable.shared->win->changed = 0;
```
pgaskin, line 95: Nitpick: I don't think this function should exist; it's good for encapsulation, but it's inconsistent with the rest of cmus, where the changed flags on the views are directly manipulated by the update function.

pgaskin (approved):
Works, makes sense.

The queue_post_update thing seems out of scope for this PR at first, but the existence of the `lib_editable.shared->win->changed = 0;` line earlier in the function (which has a similar purpose, but for the library/playlist views) kind of justifies this change being part of the PR (since otherwise the logic wouldn't be the same for the queue). It's a bit hacky, but no worse than what's already done.

pgaskin:
@flyingmutant, probably merge this as a rebase to keep the changes separate.

pgaskin:
@flyingmutant, do you plan to merge this, or are you still reviewing it?

flyingmutant:
Merged, finally! Thanks a lot for the PR.

PR #1317, opened 2024-02-27
Make trackwin heading format configurable, etc.

gavtroy:


pgaskin (approved):
Tested, code looks good, changes are really well though-out.

I especially like the statusline change; it makes significantly more room for the progress bar in your other PR, makes the volume mouse binding more intuitive, and looks cleaner overall.

The other changes like showing the track count during updates (especially now that the library total is moved to the list header) and showing the album/artist in the normal header are also nice.

pgaskin:
@flyingmutant, and this should be rebase-merged rather than squashed.

pgaskin:
@flyingmutant

PR #1318, opened 2024-02-29
Draw progress bar in statusline

gavtroy:


pgaskin:
It might look better to have the progress bar shown using a slightly different background color instead of as a character.

gavtroy:
That was my thought too. I started by playing around with using the entire status bar and just changing background color, but I made the mistake of just using the titlebar color and finding the contrast was either much too distracting (e.g. default theme) or that there was no background color at all (e.g. amazon theme). I switched gears at that point and found I really liked the simplicity of a thin line, but the first idea could be workable if we added a new `color_statusline_bar_bg` and set it for each theme to just a "slightly different" color as you say.

I was able to ctrl-z back to that experiment and push a branch [gavtroy/color-progress-bar](https://github.com/gavtroy/cmus/tree/color-progress-bar) if you want to get a feel for it.

gavtroy:
Changed ────── to ━━━━━━━━ . Also added a non-color "shuttle" option for consistency.

pgaskin:
So I tested the different options. The shuttle looks good in all cases (especially now that the line is thicker), but isn't as accurate on narrow screens. The color one looks good in some cases.

I think this is good enough to be merged, but it might be an idea to add a second boolean option to draw the progress bar in the command area, hiding it when text is displayed there. This will reduce the clutter in the status line, while also making it more useful on small displays (and then another PR could be opened to optionally make the command line into a mouse-controlled seekbar).

pgaskin (approved):
Good enough as-is, works well in general, no bugs even when resizing the terminal to various sizes.

Some potential future improvements:
- Another boolean option to display the progress bar in the command line rather than the status line could be nice to have.
- Some color schemes could have better color selections, but that can be done later.
- Mouse support could also be added in another PR later.

pgaskin:
@flyingmutant

pgaskin:
@flyingmutant

options.c
```diff
@@ -1368,6 +1371,27 @@ static void toggle_block_key_paste(void *data)
...
+
+static void set_progress_bar(void *data, const char *buf)
+{
+	parse_enum(buf, 0, 4, progress_bar_names, &progress_bar);
+}
+
+static void toggle_progress_bar(void *data)
+{
+	progress_bar++;
+	progress_bar %= 5;
```
flyingmutant, line 1391: Something like `NR_PROGRESS_BAR_TYPES` would be good to avoid hard-coding 5. Also, `cycle` (or similar) naming might be more appropriate than `toggle`.

gavtroy: Agree but currently it's only exposed as ":toggle" (or win-toggle/spacebar in the settings screen) and all matching settings use the toggle_* naming. I'm open to adding a :cycle alias but I think it's different from what you were asking.

(And oops, missed the rebase in my first push.)

flyingmutant:
Thanks! Merged, let's see what the wider feedback will be.

PR #1320, opened 2024-03-06
Trim gaps from mp4 playback

### LINKED ISSUE #1280: Gapless playback with m4a(MPEG-4)
Is gapless playback supported with M4A files? As I understand, it's supported by MP3 files and FLAC files, however all of my music was ripped from iTunes and is in the M4A format.

**This is probably the wrong place for this so lmk if it should go somewhere else** :)

gavtroy:
At least for me there can sometimes still be soft clicks between tracks that aren't there when playing back the externally decoded wavs, but the buffer looks okay and gapless mp3 has the same clicks so I guess it's an unrelated hitch in the player or on the output side. In any case this is a very mild artifact compared to what we had before.

Closes #1280.

### DIFF
--- a/ip/aac.c
+++ b/ip/aac.c
@@ -205,6 +205,7 @@ static void aac_get_channel_map(struct input_plugin_data *ip_data)
 		return;
 
 	buf = NeAACDecDecode(priv->decoder, &frame_info, buffer_data(ip_data), buffer_length(ip_data));
+	NeAACDecPostSeekReset(priv->decoder, 0);
 	if (!buf || frame_info.error != 0 || frame_info.bytesconsumed <= 0
 			|| frame_info.channels > CHANNELS_MAX)
 		return;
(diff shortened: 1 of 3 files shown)


gavtroy:
If testing, be aware that seeking makes it impossible to calculate the amount we need to trim from the end. Gaplesss mp3 playback has the same limitation.

pgaskin (approved):
Don't really have a good way of testing this, but it compiles, and I can play a MP4. Haven't really reviewed the code.

pgaskin/cmus@2658824598453fe8f7123dd2c7f4d586c745f622 ([rpms](https://github.com/pgaskin/rpms/actions/runs/9038416415))

gavtroy:
Thanks. I don't have a messy collection good for finding potential problem tracks, but for reference the tracks I was using to see it _work_ are [gapless_m4a_tracks.zip](https://i.cantlogout.com/gapless_m4a_tracks.zip)

pgaskin:
Have been running this for a while, and I don't see any problems with MP4s. Also tested with the provided gapless tracks. I don't know enough about AAC or MP4 to properly review this, but it seems fine.

@flyingmutant

flyingmutant:
Thanks! Merged.

PR #1325, opened 2024-03-27
Remember playlist position when resume=true

### LINKED ISSUE #729: Remember the last-played track in the playlist view across invocations of cmus
I restart cmus multiple times per day because it [hangs](https://github.com/cmus/cmus/issues/391), [stops playing](https://github.com/cmus/cmus/issues/481) through my sound card, or because of normal reboots due to OSX system updates. I have a very long playlist of ambient background music I listen to at work and have to manually find where I left off and resume playing each time I restart cmus.

It would be nice if cmus remembered its cursor position in the playlist view.

```
cmus --version
cmus v2.7.1
```

gavtroy:
Slightly overengineered as it could be done without a loaded event, but I wanted to avoid surprising behavior in the unlikely case where the playlist is cleared offline and then repopulated some time after cmus was started.

Fixes #729.

### DIFF
--- a/job.c
+++ b/job.c
@@ -380,6 +380,7 @@ static void add_pl(const char *filename)
 		cmus_playlist_for_each(buf, size, reverse, handle_line, cwd);
 		free(cwd);
 		munmap(buf, size);
+		add_ti(NULL); // marks end of load
 	}
 }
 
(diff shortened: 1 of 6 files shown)


jdholtz:
This works great for me. Thanks gavtroy!

pgaskin (approved):
Tested, LGTM.

pgaskin:
@flyingmutant

flyingmutant:
Thanks! Merged.

PR #1326, opened 2024-04-22
Fix view segv in rc config

adeason:
I noticed if I have a `view` command in my rc config (e.g. `view playlist`), cmus segfaults at startup. I can provide a more detailed backtrace if needed, but I thought it's clear enough what's happening from the description in the commit message in here.

I also added a note in the manpage to say not to use `view` in the rc; I assume you're not supposed to do that, because the `start_view` option exists instead. I feel like `view` used to either work or do nothing, for old versions from years ago, because I had this in my rc for an old install. Using `view` in the config could also throw an error, I wasn't sure what's preferred.

### DIFF
--- a/Doc/cmus.txt
+++ b/Doc/cmus.txt
@@ -711,6 +711,9 @@ version
 view <name or 1-7>
 	Changes the active view.
 
+	Do not use this in `$XDG_CONFIG_HOME/cmus/rc` to change the starting
+	view. Set the *start_view* option instead.
+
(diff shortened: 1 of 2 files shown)


pgaskin (approved):
Makes sense, fixes the crash.

The `start_view` option was added in f1e96d88f1a277cdaca2d2df790a2ca4b99c2002 and released in 2.8.0.

pgaskin:
@flyingmutant

flyingmutant:
Thanks! Merged.

Issue #1328, opened 2024-05-11
Pull-based output plugins


pgaskin:
So, I've been looking at refactoring the output plugin API, and potentially directly adding a pipewire (for better mixer integration, and because other players are doing it) and an aaudio one (so I don't need to run pulseaudio in termux).

One thing I noticed is that some output plugins (e.g., coreaudio, jack) have complex buffering logic since they only have pull-based APIs, other ones (e.g., pulseaudio) could have more control and about the same amount of code by switching to their pull-based API, and the remaining wouldn't be much more complex if a thread was started which pulls audio as necessary.

Using a pull-based API for output plugins would:

- Be more compatible and easier to use with most sound APIs.
- Allow for lower-latency (or higher-latency for power-saving) audio with less complexity.
- Give sound APIs more control over buffer sizes.
- Reduce the total amount of code for output plugins.
- Prevent the playback position in cmus getting out of sync in certain edge cases.

Input plugins shouldn't need to change.

The player code in cmus itself should be about the same length in the end.

@gavtroy thoughts?

gavtroy:
I'm not sure what that change would look like. For the example of pulseaudio, would it still implement `write`, in turn using `pa_stream_write`, or are you looking to replace one or both of those? That level of refactoring around the player code sounds ambitious (and I don't know much about it) but if you can figure it out I'm happy to review it.

Also no objections to adding pipewire / aaudio backends.

nefthy:
I don't think the buffering in the jack plugin is avoidable. The op_jack_cb is called from the jack thread created by the API and has time constraints. The requested samples need to be deliver so, that the whole jack graph can be processed on each cycle. The callback provided to jack, must not do heavy processing and especially no disk io.

PR #1336, opened 2024-06-03
[FFmpeg] Add .dsf to extension list (DSD support)

### LINKED ISSUE #796: DSD support
Is there any way to get cmus to play DSD files?

gavtroy:
\+ resample to a lower sample rate if needed, as some variants of DSD exceed the 524kHz we support internally -- and the 384kHz limit of at least some outputs (pulse).

Fixes #796.

### DIFF
--- a/ip/ffmpeg.c
+++ b/ip/ffmpeg.c
@@ -141,6 +141,7 @@ static int ffmpeg_open(struct input_plugin_data *ip_data)
 	int err = 0;
 	int i;
 	int stream_index = -1;
+	int out_sample_rate;
 	AVCodec const *codec;
 	AVCodecContext *cc = NULL;
 	AVFormatContext *ic = NULL;
@@ -237,6 +238,7 @@ static int ffmpeg_open(struct input_plugin_data *ip_data)
 	priv->output = ffmpeg_output_create();
 
 	/* Prepare for resampling. */
+	out_sample_rate = min_u(cc->sample_rate, 384000);
 	swr = swr_alloc();
 #if LIBAVCODEC_VERSION_MAJOR >= 60
 	if (cc->ch_layout.order == AV_CHANNEL_ORDER_UNSPEC)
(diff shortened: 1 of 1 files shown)


ip/ffmpeg.c
```diff
@@ -248,15 +250,15 @@ static int ffmpeg_open(struct input_plugin_data *ip_data)
...
 	ip_data->private = priv;
 #if LIBAVCODEC_VERSION_MAJOR >= 58
-	ip_data->sf = sf_rate(cc->sample_rate) | sf_channels(cc->ch_layout.nb_channels);
+	ip_data->sf = sf_rate(out_sample_rate) | sf_channels(cc->ch_layout.nb_channels);
 #else
-	ip_data->sf = sf_rate(cc->sample_rate) | sf_channels(cc->channels);
+	ip_data->sf = sf_rate(out_sample_rate) | sf_channels(cc->channels);
 #endif
 	switch (cc->sample_fmt) {
 	case AV_SAMPLE_FMT_U8:
```
gavtroy, line 264: By the way it looks like `default`/`AV_SAMPLE_FMT_S16` is the only one that actually plays back normally. Removing the other cases works but I'd like to understand it more.

(This is nothing to do with DSD or this change though. DSD is `AV_SAMPLE_FMT_FLTP` so it takes the `default` path without issue.)

pgaskin:
@gavtroy, are you still working on this?

ip/ffmpeg.c
gavtroy: The issue with the `AV_SAMPLE_FMT_U8` and `AV_SAMPLE_FMT_S32` cases is the hardcoded sizeof(int16_t) in the buffer_used_len calculation.

gavtroy:
I looked into why the limit was 384000 and the resampling was actually working fine, the limit was from the output side (pulse).

Do note that although this seems to work fine I'm not absolutely confident if it's "right". Forcing a higher sample rate to a lower sample rate like this behaves completely normally, but doing the opposite is clearly broken - the audio cuts off early and seeking is a bit weird. I think it probably just breaks some assumption we make about the output not growing and that the high->low case is fine, but ideally I'd understand it properly and right now I just don't have the energy to do that. In any case we only downsample when it's >384000 so this isn't relevant for pretty much anything except DSD128.

pgaskin (approved):
Code makes sense, shouldn't cause regressions, and seems to work fine with DSD on AAudio and PulseAudio.

> I think it probably just breaks some assumption we make about the output not growing and that the high->low case is fine, but ideally I'd understand it properly and right now I just don't have the energy to do that

It's not apparent to me either, but this works, and why higher sample rates don't is something we can figure out later, probably when I implement #1328.

> In any case we only downsample when it's >384000 so this isn't relevant for pretty much anything except DSD128.

Agreed. This change doesn't affect lower sample rates, and ones higher than this are rarely seen in mainstream formats.

pgaskin:
@flyingmutant, this should be rebase-merged since the sample size fix commit is a separate (but related) bugfix.

Issue #233, opened 2015-01-09
Feature request: Auto-update library


OOPMan:
Hi

Great music player you have here. I would love it if it were able to automatically update the library. I regularly add new music and this is a pretty important feature for me...

helveticade:
+1 to this, watching folders would be really handy

OOPMan:
@flyingmutant Yes, pretty much.

elliottetzkorn:
I would also love this feature!

stevenleeg:
Yeah this would be great. Is there any way to rescan the library's directories at the moment, or do I have to add a new album's folder manually?

HarveyHunt:
I think you can achieve this using an inotify script (or even [when-changed](https://github.com/joh/when-changed)) to pipe a raw command into ```cmus-remote``` to update your library - although I can't remember which command you would need to send to cmus...

KabbAmine:
:+1:  for auto-update functionality

bombush:
+1 ! wached folders would be great

DavidOliver:
Without a watched folder/auto update feature, is the solution to simply re-add a directory? (Asking as a very new cmus user Thanks!)

kuzyn:
@DavidOliver as far as I know, yes: `:clear` and then `:add /your/folder`

nefthy:
While :clear / :add will do the trick. Its not realy elegant.

A :remove <file> command should be quite easy to implement. So an external script that watches a folder could do something like when a file is modified. 
```
:remove /path/to/watched/file.mp3
:add /path/to/watched/file.mp3
```

An implementation in cmus with inotify is not that hard.

@flyingmutant or @mahkoh: can you assign this issue to me. I'll implement that when I find time.

dodona2:
I support that automatically update or rescan the library.

XVilka:
Would be really awesome to have.

c0llision:
Here is how I handle reloading my library.

Create the following script somewhere:
`#!/bin/bash
cmus-remote -C clear
cmus-remote -C "add ~/Music"
cmus-remote -C "update-cache -f"`

Replace ~/Music with the path to your library.

Then do:
`chmod +x your-script.sh`

Then open cmus and run:
` :bind -f common shell u shell ~/path/to/your-script.sh`

Then every time you press 'u', your library will be reloaded. I used the key 'u', because it was already bind to reload-cache, which my script does anyway.

[...]

PR #1343, opened 2024-07-21
Add an autoupdate contrib script

MikeSS8:
There's been a lot of interest in this for the last decade.
https://github.com/cmus/cmus/issues/233

### DIFF
--- a/contrib/cmus-autoupdate.sh
+++ b/contrib/cmus-autoupdate.sh
@@ -14,6 +14,7 @@ which inotifywait >/dev/null || { printf "This script requires inotify-tools to 
 # NB. if your files are on removable storage, -e unmount may be useful here.
 inotifywait --csv --monitor --recursive -e create -e delete "$1" | \
 while IFS=, read -r path event file; do
+	file="$(printf "%s\n" "$file" | sed -e 's/^"//;s/"$//')"
 	case "$event" in
     	CREATE) cmus-remote -C "add ${path}${file}" ;;
     	*) cmus-remote -C "update-cache" ;;
--- a/contrib/cmus-autoupdate.sh
+++ b/contrib/cmus-autoupdate.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+usage() {
+    printf "%s <music directory>
+    Automatically add/remove files to/from the Library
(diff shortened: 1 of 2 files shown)


contrib/cmus-autoupdate.sh
```diff
@@ -0,0 +1,22 @@
...
+    when files in the given directory are changed." $0
+    exit
+}
+
+[ -z "$1" ] && usage
+
+which inotifywait >/dev/null || { printf "This script requires inotify-tools to be installed."; exit; }
+
+# NB. if your files are on removable storage, -e unmount may be useful here.
+inotifywait --csv --monitor --recursive -e create -e delete "$1" | \
```
pgaskin, line 15: This is racy; created files may not be finished writing. The `close_write` may be a better alternative. Also, `moved` is also required to catch all relevant changes.

contrib/cmus-autoupdate.sh
```diff
@@ -0,0 +1,22 @@
...
+while IFS=, read -r path event file; do
+	file="$(printf "%s\n" "$file" | sed -e 's/^"//;s/"$//')"
```
pgaskin, line 17: To avoid this splitting logic and to simplify the parsing, using the `--format` argument with `%e|%w%f` and using `while IFS='|' read -r event path` should be simpler and handle all special characters properly.

pgaskin:
Apart from my feedback, the other question is whether this actually belongs in contrib rather than the wiki. The contrib folder currently only contains code which integrates cmus with the shell, or scripts which cmus itself can call. The existing scripts are also the kind where there is really only one way to do it. The wiki currently contains pretty much every other kind of script.

Thoughts, @flyingmutant?

flyingmutant:
> Apart from my feedback, the other question is whether this actually belongs in contrib rather than the wiki. The contrib folder currently only contains code which integrates cmus with the shell, or scripts which cmus itself can call. The existing scripts are also the kind where there is really only one way to do it. The wiki currently contains pretty much every other kind of script.
> 
> Thoughts, @flyingmutant?

I agree. I consider `contrib` mostly an artifact of times-before-github-wiki. Keeping the code in the wiki will be more convenient and conventional.

PR #1347, opened 2024-08-06
Fix configure function checks for recent clang versions

pgaskin:
Simply try taking the address of the function instead of doing funny stuff with the pointer.

### DIFF
--- a/configure
+++ b/configure
@@ -503,9 +503,8 @@ check_string_function()
 	msg_checking "for function $1"
 	string_function_code="
 #include <string.h>
-int main() {
-	return $1;
-}
+int (*ptr)() = (int (*)()) &$1;
+int main() { return 0; }
 "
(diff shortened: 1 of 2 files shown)


pgaskin (commented):
Tested on multiple compiler versions with and without strict and/or pedantic C89+, with existent and non-existent functions.

pgaskin:
Also see termux/termux-packages#15852.

pgaskin:
@flyingmutant

gavtroy (approved):
Looks okay. It might also have been enough to `void *x = (void*) &$1;` rather than using a function pointer signature.

pgaskin:
That results in a warning in some configurations (and an error if the compiler is configured to be strict) @gavtroy.

flyingmutant:
Thanks! Merged.

PR #1348, opened 2024-08-06
Include CPPFLAGS in configure checks

pgaskin:
This brings the logic in line with the Makefile.

https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html

### DIFF
--- a/scripts/checks.sh
+++ b/scripts/checks.sh
@@ -59,13 +59,13 @@ check_program()
 
 cc_supports()
 {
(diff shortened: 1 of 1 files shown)


pgaskin:
Also see [termux/termux-packages@15d4f5d/packages/cmus/configure.patch](https://github.com/termux/termux-packages/blob/15d4f5da826ff2263832b56f5ee4c209cd4148f0/packages/cmus/configure.patch).

pgaskin:
@flyingmutant

flyingmutant:
Thanks! Merged.

PR #1349, opened 2024-08-06
Fix missing NCURSES_CFLAGS in configure try_compile_link

pgaskin:
This makes it match the other libraries, and it's clearly a mistake since the line below appends NCURSES_CFLAGS.

### DIFF
--- a/configure
+++ b/configure
@@ -193,7 +193,7 @@ check_ncurses()
 	msg_checking "for working ncurses setup"
 	for flag in "" "-I/usr/include/ncurses" "-I/usr/include/ncursesw"
 	do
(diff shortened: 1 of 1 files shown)


pgaskin (commented):
Cannot break anything since the actual compilation behaves this way. Fixes edge cases with compilation, e.g., in [termux](https://github.com/termux/termux-packages/blob/15d4f5da826ff2263832b56f5ee4c209cd4148f0/packages/cmus/configure.patch).

pgaskin:
@flyingmutant

flyingmutant:
Thanks! Merged.

PR #1350, opened 2024-08-06
Fix potential double-close in ip/mad

pgaskin:
In input.c, ip_close calls ip->ops->close, then closes ip->data.fd (aka ip_data->fd) if fd is not -1. In ip/mad.c, close_func closes ip_data->fd without setting it to -1.

May be the cause of the fdsan failure reported in termux/termux-packages#18954. I've been looking at all the close calls and tracing the callers, and this seems to be the most likely issue.

### DIFF
--- a/ip/mad.c
+++ b/ip/mad.c
@@ -54,8 +54,11 @@ static off_t lseek_func(void *datasource, off_t offset, int whence)
 static int close_func(void *datasource)
 {
 	struct input_plugin_data *ip_data = datasource;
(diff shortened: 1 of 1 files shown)


pgaskin:
This may or may not actually be the bug, but it's harmless if it isn't.

@flyingmutant

gavtroy (approved):

flyingmutant:
Thanks! Merged.

Issue #1305, opened 2024-01-13
crash under musl libc when playing singles


z-erica:
https://github.com/cmus/cmus/blob/23afab39902d3d97c47697196b07581305337529/player.c#L838
the previous line crashes under musl when either string is null, which can happen when a track has no album associated with it

---

1) `uname -a`:
```
Linux complooter 6.6.8-0-generic #1 SMP PREEMPT_DYNAMIC Mon Dec 25 23:28:08 UTC 2023 x86_64
```

2) `cmus --version`:
```
cmus v2.10.0
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  cdio:
    Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  cue:
    Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  opus:
    Priority: 50
    File Types: opus
    MIME Types:
  vorbis:
    Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  wav:
    Priority: 50
    File Types: wav
    MIME Types:
  wavpack:
    Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  ffmpeg:
    Priority: 30
    File Types: aa aac ac3 aif aifc aiff ape au fla flac m4a m4b mka mkv mp+ mp2 mp3 mp4 mpc mpp ogg opus shn tak tta wav webm wma wv
    MIME Types:

Output Plugins: /usr/lib/cmus/op
  pulse
  jack
```

PR #1351, opened 2024-08-06
Fix possible null dereference when continue_album=0

pgaskin:
If continue_album=0 and the current or next track doesn't have album metadata, a strcmp with a null pointer will be done at the end of the track. Ensure track_info.album is non-null like other parts of cmus.

The continue_album option was added in f3ed07ee3d46b2bc98f98247654729b02457409b.


pgaskin:
@flyingmutant

gavtroy (approved):
Fixes #1305.

flyingmutant:
Thanks! Merged.

Issue #1202, opened 2022-06-06
Can't play various internet streams despite having required plugins (I think)


holderbp:
I am having trouble playing many online radio streams (which I have been able to access with, e.g., VLC).  For example:

[https://s2.radio.co/sedf30688d/listen](https://s2.radio.co/sedf30688d/listen)   (KUOI, Moscow, ID)
[https://streamer.radio.co/scad0cc067/listen](https://streamer.radio.co/scad0cc067/listen)  (KPSU, Portland, OR)

For both of these, if I look at the Http-Header-Live output while playing in Firefox, they are both:
```
Content-Type: audio/mpeg
```
And, as you see below, I have the mad plugin installed for such filetypes.  

I'm also not able to access this one (among many others):

[https://ice24.securenetsystems.net/WYCE](https://ice24.securenetsystems.net/WYCE)  (WYCE, Grand Rapids, MI)

It has a different content-type
```
Content-Type: audio/aacp
```
but this should be handled by my aac plugin

I'm able to access other streams fine, e.g., 

[http://stream.kalx.berkeley.edu:8000/kalx-256.mp3](http://stream.kalx.berkeley.edu:8000/kalx-256.mp3)  (KALX, Berkeley)

which is also an `audio/mpeg` Content-Type.

If I try to play any of the inaccessible streams in cmus by navigating to an m3u file, I get these errors:
```
Error: Couldn't get file information for https://s2.radio.co/sedf30688d/listen
Error: Couldn't get file information for http://stream.kpsu.org:8080/listen
Error: Couldn't get file information for https://ice24.securenetsystems.net/WYCE
```

If I try to add the stream manually using (`:add <stream address>`), I get:
```
Error: adding 'https://s2.radio.co/sedf30688d/listen': No such file or directory
```

Thanks for any assistance/ideas!

---

1) `uname -a`:
```
Linux 2019pi225 5.10.63-v7+ #1488 SMP Thu Nov 18 16:14:44 GMT 2021 armv7l GNU/Linux
```

2) `cmus --version`:
```
cmus v2.7.0
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2013 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  flac:
    Priority: 50
    File Types: flac fla
    MIME Types:
[...]

Issue #207, opened 2014-11-01
HTTPS streaming support


flyingmutant:
Right now cmus does not support HTTPS at all.

stigtsp:
+1

goberghen:
+1

geckyan:
+1 #714
Its [stream.m3u](http://listen.moe/stream.m3u) contains single url: https://listen.moe/vorbis.
Cmus output is, as expected:
```
Error: adding 'https://listen.moe/vorbis': No such file or directory
```

ml-:
I'd really appeciate this feature.
If you don't mind I'd like to put a bounty of $100 on this issue. Once the feature is merged you can bug me about it.

jteppinette:
+1

libcurl integration? Is anyone looking into this?

aaether32323:
@ml- what player did you move on to?

ml-:
@aaether32323 ncmpcpp (mpd client)
https://rybczak.net/ncmpcpp/

aschleck:
I'm contemplating looking at this, is there a reason the existing http.c has its own HTTP implementation instead of using an HTTP library?

I feel like the options to support this are
* add support to http.c for detecting the https protocol and then wrapping the socket using OpenSSL (unless there's a preference for some other SSL library?)
* change http.c to use Curl as suggested above (or some other library?)

I'd prefer to do the second and just use Curl, @flyingmutant what do you think?

aschleck:
Turns out this is not very fun because input.c and http.c are super, duper intertwined (with a guest appearance from cdio.c), but it's especially not very fun because the file descriptor of the request gets passed all around the application.

Using cURL gives a few options which all suck for a variety of options like:
* now have to pass the cURL context around the application as a replacement for the fd
* cURL uses a ton of callbacks, which is the opposite of what cmus is setup for
* have to replace all the fd specific calls

Using OpenSSL or (all? most?) other SSL libraries has the downsides of passing the SSL context around and replacing all the fd specific calls (like read, and close) with the equivalent ones from the SSL library, while not having cURL's upside of improve the HTTP handling code. I think maybe the way to make this work would be to pass around function pointers to wrappers of read/SSL_read and then have the http.c code set up the pointer to the correct wrapper, but there's probably a bunch of work to make it work. I haven't spent much time trying anything here.

[...]

PR #1352, opened 2024-08-06
[HTTPS] Adding HTTPS support with OpenSSL

jprissi:
As many internet radio stations do not support HTTP anymore, there have been several requests to add HTTPS support to CMUS (see #207).

This pull request adds HTTPS support to CMUS with the help of OpenSSL. This feature allows one to add and play HTTPS streams such as the ones provided by #1202 :

`https://s2.radio.co/sedf30688d/listen` (KUOI, Moscow, ID)
`https://streamer.radio.co/scad0cc067/listen` (KPSU, Portland, OR)

The choice of OpenSSL allows us to keep most of the existing network code unchanged. There were discussions to use libcurl instead but I think it would require more effort to interface it with existing code (libcurl relies on callbacks and its API is high-level whereas CMUS makes heavy use of sockets).

**About the changes**

The goal was to replace the relevant calls to `read(fd, buf, size)` with their `SSL_read(ssl, buf, size)` counterpart. To do that I introduced a struct that contains both `int* fd_ref;` and `SSL *ssl;` named `struct connection`. 
After adding the read / write functions to the struct and setting up our `struct connection`, we can replace existing calls to `read` with a generic `conn->read(conn, buf, size)`. This way, we can still handle regular HTTP requests as before while adding support for HTTPS requests.

I tried to keep the number of changes as low as possible and to avoid editing irrelevant parts, however : 
- Half of the files have been edited for consistency reasons, by replacing `is_http_url()` with `is_http_or_https_url()`
- I would have wanted to avoid editing `input.c` but we can't simply pass sockfd around anymore and now need to use the `struct connection` abstraction instead.
- We have an ugly `int *fd_ref` in `struct connection`. The rationale is that if we do not keep the original `int fd;` then we have to edit all ip/*.c methods that rely on `ip_data->fd`. This leads to awkward situations where we have to keep the pointer updated when creating, resetting or copying `http_get` or `input_plugin_data` structs.

There is room for improvement but I don't see how to make this feature cleaner without further changes to existing code.

**What isn't working**

- Not all HTTPS streams are supported (for instance `https://stream.radiofrance.fr/fipjazz/fipjazz_hifi.m3u8`). It seems .m3u8 & AAC streams fail because of the content-type `application/vnd.apple.mpegurl`. Adding this content-type is not enough however. It appears to be caused by the way playlists are handled differently from other streams. It would perhaps be cleaner to add support for these in a later PR.
- Proxy support for HTTPS stream: we need to be able to perform a CONNECT request to the proxy. This is not difficult to add but requires making the `struct http_get` more generic to allow for other methods and refactor `http.c`.

**What is working**

- HTTPS streams that are not .m3u8 such as the ones mentioned above.

Let me know what you think! I'd be glad to discuss these changes and find a way to finally get HTTPS in CMUS.


jprissi:
Having thought about it, I'm not satisfied with some of the changes. I'm switching to draft review for now.

Some of the things I noticed :
1. Not using .mp3 for my local audio files, I did not realize there was a segfault when using mp3. This is straightforward to fix by having the default `conn->read` initialized to `&socket_read` to keep the current behavior. Still, I'd like to understand why this did not happen with other file formats (.m4a and .opus)
2. The way the SSL connection is closed needs to be improved (especially the  `wait_eof()` part). Maybe switching the connection to blocking mode before calling SSL_shutdown would allow us to make this simpler. Or at least, add a timeout to remove the possibility of an infinite loop.
3. There are some chains of pointers that hint at some functions dealing with too many levels of abstraction. Solving this could help simplify some of the code.

I will make the changes to solve 1. and 2. and start looking into 3.

pgaskin:
It may be worth attempting to refactor the input plugin API entirely to eliminate the dependency on file descriptors. I've been considering doing this myself.

jprissi:
In the context of HTTPS, it would be welcomed if there is a way to reduce the dependency on file descriptors. What kind of API changes did you consider that would help achieve this?

This draft is making some changes to the API but would indeed benefit from a more thought-out refactor if there is such a need elsewhere in the project.

pgaskin:
I was thinking of a file struct containing at least the following function pointers (most of which can be optionally unsupported):

- reopen
- read
- seek
- readat
- size
- filename (just the last component)

This would be passed to input plugins instead of most of the stuff in ip_data.

jprissi:
Thank you for the clarifications! I've taken a look at the changes to the v2 output plugin API and this does seem like a clean solution to declutter the ip_data struct. If you attempt the refactor, I can make the corresponding changes to this PR. 

In the meanwhile, I have attempted to clean up this initial HTTPS PR. I don't think I will be able to improve it much further. I've updated the PR message to mention the missing features :
1. Some https streams are not supported (notably those relying on .m3u8, the code path used appears to be different. I will probably start working on these, relying on this initial https support feature)
2. When using a proxy, the https streams won't play. We need to perform a CONNECT request. In order to do so, we will need to refactor the code that handles requests (which only handles GET requests for now).

gavtroy:
You might need to bump the input plugin api version but otherwise this change LGTM. Would the input refactoring be before this change or done on top?

jprissi:
I bumped the input plugin ABI and fixed the CI build issues. (I had quite some trouble with macos and openssl and had to add the OPENSSL_CFLAGS to the CFLAGS variable. This is not ideal but I did not find a cleaner solution to provide the OpenSSL headers to the input plugins)

pgaskin (changes requested):
Looks good overall and works fine with the input plugins which work with HTTP, but can we make the OpenSSL support optional?

pgaskin:
>  had to add the OPENSSL_CFLAGS to the CFLAGS variable. This is not ideal but I did not find a cleaner solution to provide the OpenSSL headers to the input plugins

We can deal with that after refactoring the file operations to not work this way in a future PR.

http.c
```diff
@@ -150,7 +153,7 @@ int http_open(struct http_get *hg, int timeout_ms)
 	char *proxy = getenv("http_proxy");
 	if (proxy) {
 		hg->proxy = xnew(struct http_uri, 1);
-		if (http_parse_uri(proxy, hg->proxy)) {
+		if (parse_uri(proxy, hg->proxy)) {
```
anvilvapre, line 156: Trivial. Below: `Failed to parse HTTP proxy URI` becomes `Failed to parse HTTP(s) proxy URI`.

http.c
```diff
@@ -221,10 +224,55 @@ int http_open(struct http_get *hg, int timeout_ms)
...
+{
+	if (socket_open(hg, timeout_ms))
+		return -IP_ERROR_ERRNO;
+	*conn->fd_ref = hg->fd;
+
+	if(hg->is_https == 0)
+		return IP_ERROR_SUCCESS;
+
+	if(hg->proxy != NULL) {
+		/*
```
anvilvapre, line 237: This suggests that when using http you already return early and the proxy setting is ignored and no warning message is given that the proxy is not supported. (?)

http.c
```diff
@@ -221,10 +224,55 @@ int http_open(struct http_get *hg, int timeout_ms)
...
+		 * In order to do that, we need to refactor the code that does GET requests
+		 * to also support CONNECT request. This is getting out of scope for HTTPS support.
+		 * We can return an error message for now.
+		 */
+		d_print("HTTPS stream with proxy not yet supported.\n");
+		return -IP_ERROR_FUNCTION_NOT_SUPPORTED;
+	}
+
+	if (ssl_open(conn))
+		return -IP_ERROR_OPENSSL;
```
anvilvapre, line 251: No need to close and reset `fd` on error?

anvilvapre: Also no need to clear `conn->ssl` on error?

http.c
```diff
@@ -221,10 +224,55 @@ int http_open(struct http_get *hg, int timeout_ms)
...
+	int rc = 0;
+
+	if (conn->ssl != NULL)
+		rc = ssl_close(conn->ssl);
+	if (rc)
+		d_print("Error while closing ssl connection\n");
+
+	int fd = get_sockfd(conn);
+	close(fd);
+
```
anvilvapre, line 267: Consider clearing `fd` and `ssl`.

Check if fd is set before calling close?

http.c
```diff
@@ -289,17 +337,18 @@ static int read_timeout(int fd, int timeout_ms)
 }
 
 /* reads response, ignores fscking carriage returns */
-static int http_read_response(int fd, struct gbuf *buf, int timeout_ms)
+static int http_read_response(struct connection *conn, struct gbuf *buf, int timeout_ms)
 {
 	char prev = 0;
+	int fd = get_sockfd(conn);
```
anvilvapre, line 343: Adding `conn->has_read_timeout()` seems cleaner.

Also because SSL may have it's own BIO timeout settings that may not be applied on `fd` level.

http.c
```diff
@@ -415,18 +464,17 @@ int http_get(struct http_get *hg, struct keyval *headers, int timeout_ms)
...
-char *http_read_body(int fd, size_t *size, int timeout_ms)
+char *http_read_body(struct connection *conn, size_t *size, int timeout_ms)
 {
 	GBUF(buf);
-
-	if (read_timeout(fd, timeout_ms))
+	if (read_timeout(get_sockfd(conn), timeout_ms))
 		return NULL;
 	while (1) {
 		int count = 1023;
```
anvilvapre, line 473: Odd number.

http.h
```diff
@@ -40,7 +40,9 @@ struct http_uri {
 	int port;
 };
 
+struct connection;
 struct http_get {
+	int is_https;
 	struct http_uri uri;
 	struct http_uri *proxy;
```
anvilvapre, line 47: Since your uri parser function does not support ports i.e. `[:port]` and most often proxy run on different ports than the default 80 or 443 how does proxy support even work?

http.h
```diff
@@ -49,22 +51,24 @@ struct http_get {
...
 };
 
-int http_parse_uri(const char *uri, struct http_uri *u);
+int parse_uri(const char *uri, struct http_uri *u);
 
 /* frees contents of @u, not @u itself */
 void http_free_uri(struct http_uri *u);
 
-int http_open(struct http_get *hg, int timeout_ms);
+int socket_open(struct http_get *hg, int timeout_ms);
```
anvilvapre, line 59: I would hide everything socket away to only use connection. Make it private.

Only let fd be part of connection.

input.c
```diff
@@ -211,11 +211,16 @@ static int do_http_get(struct http_get *hg, const char *uri, int redirections)
 	hg->proxy = NULL;
 	hg->code = -1;
 	hg->fd = -1;
-	if (http_parse_uri(uri, &hg->uri))
+	hg->is_https = is_https_url(uri);
+	conn->write = hg->is_https ? &https_write : &socket_write;
+	conn->read = hg->is_https ? &https_read : &socket_read;
+
+	if (parse_uri(uri, &hg->uri))
```
anvilvapre, line 218: Consider passing a uri string to connection to be able to hide the uri parsing inside/behind connection. Now in many places you first parse the url and after create a connection.

input.c
```diff
@@ -260,9 +265,9 @@ static int do_http_get(struct http_get *hg, const char *uri, int redirections)
 
 		redirloc = xstrdup(val);
 		http_get_free(hg);
-		close(hg->fd);
+		connection_close(conn);
```
anvilvapre, line 268: Won't http_get_free also not already close the `fd`? Causing you to close it twice. `close` won't mind, still.

input.c
```diff
@@ -336,31 +348,40 @@ static int handle_line(void *data, const char *uri)
...
+	rpd->rc = setup_remote(rpd->ip, hg.headers);
 	http_get_free(&hg);
 	return 1;
 }
 
-static int read_playlist(struct input_plugin *ip, int sock)
+static void set_fd(struct input_plugin_data *data, int fd)
+{
+	data->fd = fd;
+	data->conn.fd_ref = &data->fd;
```
anvilvapre, line 372: This suggests you assume that `input_plugin_data` will already always have a connection. Which makes keeping `fd` even more irrelevant?

input.c
```diff
@@ -431,10 +462,10 @@ static void ip_reset(struct input_plugin *ip, int close_fd)
 	ip_init(ip, ip->data.filename);
 	if (fd != -1) {
 		if (close_fd)
-			close(fd);
+			connection_close(get_connection(ip));
 		else {
 			lseek(fd, 0, SEEK_SET);
```
anvilvapre, line 467: How would this work? Your trying to do a lseek on a socket and/or ssl socket stream?

input.c
```diff
@@ -640,7 +671,9 @@ void ip_setup(struct input_plugin *ip)
 int ip_close(struct input_plugin *ip)
 {
 	int rc;
-
+	struct connection *conn = &ip->data.conn;
+	if (conn->ssl != NULL)
+		ssl_close(conn->ssl);
```
anvilvapre, line 676: `connection->close()` which would then internally check to see if `fd` and `ssl` are set.

read_wrapper.c
```diff
@@ -58,8 +60,32 @@ ssize_t read_wrapper(struct input_plugin_data *ip_data, void *buffer, size_t cou
...
+{
+	char *buffer = buf;
+	ssize_t pos = 0;
+
+	do {
+		ssize_t rc;
+
+		rc = conn->read(conn, buffer + pos, count - pos);
+		if (rc == -1) {
+			if (errno == EINTR || errno == EAGAIN)
```
anvilvapre, line 80: ssl bio can not also return other types of return codes that indicate `try again` and/or `interrupted` or `oef`.

read_wrapper.h
```diff
@@ -20,10 +20,12 @@
 #define CMUS_READ_WRAPPER_H
 
 #include "ip.h"
+#include "ssl.h" /* struct connection */
 
 #include <stddef.h> /* size_t */
 #include <sys/types.h> /* ssize_t */
 
 ssize_t read_wrapper(struct input_plugin_data *ip_data, void *buffer, size_t count);
+ssize_t read_all_from_conn(struct connection *conn, void *buf, size_t count);
```
anvilvapre, line 29: `connection_read_all`?

jprissi:
> Looks good overall and works fine with the input plugins which work with HTTP, but can we make the OpenSSL support optional?

I agree that it would be better overall. I'll look into it!

jprissi:
Thank for the code review @anvilvapre, I appreciate it. I'll look through them.

http.c
```diff
@@ -36,13 +39,13 @@
 #include <errno.h>
 
 /*
- * @uri is http://[user[:pass]@]host[:port][/path][?query]
+ * @uri is http[s]://[user[:pass]@]host[:port][/path][?query]
```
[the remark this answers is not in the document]
jprissi: AFAICT `[:port]` is handled correctly, although a bit later in the function. The `u->port = is_https(uri) ? 443 : 80` is only here for the initialization and gets overriden in the case where a port number is specified.

I checked the behavior was working as expected by running a local `mpd` stream on a non-default port.

jprissi:
> Looks good overall and works fine with the input plugins which work with HTTP, but can we make the OpenSSL support optional?

I have a first draft that can be built both with and without OpenSSL. I still need to tidy things up and I'll push it.

utils.h
```diff
@@ -198,6 +198,16 @@ static inline int is_http_url(const char *name)
 	return strncmp(name, "http://", 7) == 0;
 }
 
+static inline int is_https_url(const char *name)
+{
+	return strncmp(name, "https://", 8) == 0;
```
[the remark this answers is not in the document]
pgaskin: This should be fixed in another PR (strnicmp isn't portable, and the existing http check is also case-sensitive).

http.c
jprissi: Good catch, I hadn't considered the possibility that http_proxy could contain both HTTP/HTTPS URIs. I realize my assumption about `http_proxy` and `https_proxy` env variables were wrong and the current behavior is not right.

So before changing the error message we'll need to do a few things :
- Add a "not supported" warning message for `https_proxy` env variable (not supported until we implement CONNECT requests) and remove the current error message for HTTP proxy / HTTPS target URI which does not make sense.
- Test HTTPS proxy for HTTP target URIs (and if it works fine, add the previous error message).

I think this should be enough to remove any surprising behavior.

http.c
jprissi: This number is not from me and has been here for quite some time.
Isn't it a way to keep the buffer null-terminated and aligned?

I don't think there's much benefit to changing this value without further information.

jprissi:
> I have a first draft that can be built both with and without OpenSSL. I still need to tidy things up and I'll push it.

Done! This PR can now be built without OpenSSL (`./configure CONFIG_OPENSSL=n`). 
If the stream is https and OpenSSL is disabled, the user will get an error message when trying to play it.

I have also started integrating some of the most straightforward comments from anvilvapre but will need some more time to check the others.

.github/workflows/build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           brew install \
             libao ffmpeg libcddb libcdio libcue libdiscid faad2 flac jack mad \
             libmodplug mp4v2 musepack opusfile libsamplerate libvorbis wavpack \
-            pkg-config
+            pkg-config openssl@3
```
pgaskin, line 24: The README also needs to be updated with the new dependencies.

configure
```diff
@@ -555,6 +562,7 @@ HAVE_CDDB=n
 CONFIG_TREMOR=n
 CONFIG_MIKMOD=n
 CONFIG_BASS=n
+CONFIG_OPENSSL=y
```
pgaskin, line 565: Is there a reason why this isn't being left to automatic?

pgaskin (commented):
Build stuff for OpenSSL looks good at-a-glance, haven't got a chance to test it fully yet.

configure
jprissi: No actual reason. It probably makes more sense to have it set to automatic, indeed.

n-connect:
@jprissi 

Thanks for this PR, did a build with the https stream - most of the online https streams plays well!

Bumped into some errors:

First, I'm not sure if I made something wrong during build 
- ./configure CONFIG_OPENSSL=y # to make sure OpenSSL is used, therefore the https capability gets compiled in...
- make;make install DESTDIR=/tmp/cmus_ssl_2.12.0 )

Because cmus sometimes starts with this error:
```
cmus: Error: no such option search_resets_position
cmus: Press <enter> to continue.
```
Second, some of the online radio streams I've tried are not loading with error, for example :
- Error: Couldn't get file information for https://radio.radiocafe98.hu/radiocafe98_192k or https://stream.danubiusradio.hu/danubius_320k or https://stream.synthwaveradio.eu/listen/synthwaveradio.eu/radio.mp3 . These links works if opened in browser, although I'm not sure if its due to a different streaming method (other than clean Icecast/Shoutcast).

Third, others gave different errors, one even segfaulted cmus -> if I opened them in browser, the original links transformed to another (URL rewrite), and the playing new URL played well in cmus too - sas they definitely were not the direct stream URL, `cmus` has no chance to play them. Maybe some exception handling could be good here - which are generic streaming error handling not https.

http.c
gavtroy: ssl is cleared in ssl_close (-ish, anyway after https://github.com/gavtroy/cmus/commit/ea8b00e99513bcec7d5bf0a6673d243a9341647b)

The fd -1 thing sounds like a good idea, I can add it afterwards if it doesn't get done in this PR.

input.c
gavtroy: http_get_free only frees headers, the fd is kept in the ip data

input.c
gavtroy: By my reading it's just to keep the underlying fd consistent for when it is created. There's some rationale for this design in the PR description (keeping churn lower).

input.c
gavtroy: Yes it wouldn't. This code path is only hit by local files. The changes below were just for consistency I assume.

input.c
gavtroy: Agree connection_close seems not to be fully exploited everywhere, although I didn't read closely so maybe there's a reason for the distinction. But since the PR has stalled it's probably fine as is.

gavtroy (commented):
@jprissi any chance you could rebase, and possibly cherry pick from https://github.com/gavtroy/cmus/commits/https-support? Otherwise I can open a new PR.

Looking through the previous review comments there's nothing that should be a blocker to merging it IMO. It seems a shame to leave it sitting.

PR #1353, opened 2024-08-07
Fix pause_on_output_change not working when softvol is enabled

pgaskin:


pgaskin:
@flyingmutant

flyingmutant:
Thanks! Merged.

PR #1363, opened 2024-09-20
Improve compatibility with autoconf-style configure options

### LINKED ISSUE #1261: [Feature Request] Support for prefix? (Edit: Actually meaning --prefix= in addition to prefix=)
Hey there,

Would it be possible to add support for prefix e. g. --prefix or just a PREFIX or prefix variable in 
the Makefile?

I would like to easily control where to install cmus, e. g. /opt/cmus as one example. I can do so
on my own, of course, by putting the compiled files on the local system, but having support
for prefix directly would be very convenient. Thank you for reading this and possibly 
considering this request (I got a bit tired of ffmpeg and had some issues, so having small,
fast alternatives is actually useful).

Edit: Oh there is, via "./configure prefix=$HOME" just noticed it. Can the "--" also be
supported, so that we can use --prefix=? This may be easier for streamlining, I
actually tried --prefix first but got an error message on the commandline.

pgaskin:
Many packages need a workaround for setting the prefix without this.

fixes #1261

### DIFF
--- a/configure
+++ b/configure
@@ -562,7 +562,8 @@ Options:
   datadir         Read-only data         [\$prefix/share]
   libdir          Libraries              [\$prefix/lib]
   mandir          Man pages              [\$datadir/man]
-  exampledir      Examples               [\$datadir/doc/cmus/examples]
+  docdir          Other documentation    [\$datadir/doc/cmus]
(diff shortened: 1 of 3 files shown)


pgaskin:
@flyingmutant, this should be rebase-merged.

PR #1367, opened 2024-09-21
Fix ffmpeg configuration

### LINKED ISSUE #1366: Edge cases in configure/build processes causing incorrect linkage
---

1) `uname -a`:
```
Darwin Mac-Studio.i0ntempest.home 24.0.0 Darwin Kernel Version 24.0.0: Mon Aug 12 20:52:12 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T6020 arm64
```

2) `cmus --version`:
```
cmus v2.11.0
```

3) `cmus --plugins`:
N/A, ffmpeg enabled

I'll be honest that my system setup is kinda complicated, but there might be many user with similar configurations during to how MacPorts handles ffmpeg versions.
So, the main ffmpeg port is version 4 and installed at the usual location in the prefix. Version 6 and version 7 are in `${prefix}/libexec`. Therefore I need to set `PKG_CONFIG_PATH=${prefix}/libexec/ffmpeg7/lib/pkgconfig:${prefix}/lib/pkgconfig` for it to find the new ffmpeg. However due to some bugs in the configure script this isn't enough.

1. `libswresample` is not searched during configure but during building, therefore the same PKG_CONFIG_PATH variable need to be present in the build phase as well.
2. Because the ffmpeg v7 linker and cflags found by pkgconfig are put AFTER the main linker and cflags respectively,, if both ffmpeg v4 and v7 are installed, even when `PKG_CONFIG_PATH=${prefix}/libexec/ffmpeg7/lib/pkgconfig` is present, the resulting binary links ffmpeg v4.

pgaskin:
fixes #1366

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,6 @@ include scripts/lib.mk
 
 CFLAGS += -D_FILE_OFFSET_BITS=64
 
-FFMPEG_CFLAGS += $(shell pkg-config --cflags libswresample)
-FFMPEG_LIBS += $(shell pkg-config --libs libswresample)
-
 CMUS_LIBS = $(PTHREAD_LIBS) $(NCURSES_LIBS) $(ICONV_LIBS) $(DL_LIBS) $(DISCID_LIBS) \
 			-lm $(COMPAT_LIBS) $(LIBSYSTEMD_LIBS)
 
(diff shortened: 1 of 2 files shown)


pgaskin:
@flyingmutant, this should be rebase-merged.

PR #1370, opened 2024-09-22
Prevent search from wrapping infinitely

### LINKED ISSUE #1332: search, keyboard stops to respond after typing 4 slashes
Typing four times a slash locks up the keyboard/application. After three slashes no additional slashes appear on screen and the keyboard appears to be locked. Music keeps playing. 

Only a `kill -9` allows me to stop cmus after. 

Up to 3 the keyboard stays responsive, the slashes appear on screen and i can remove them using backspace. Each time on the fourth the keyboard end up in a state i do not how to get out of.

---

1) `uname -a`:
```
Ubuntu 22.04, x86_64 
```

2) `cmus --version`:
```
cmus --version
cmus v2.9.1
```

3) `cmus --plugins`:
```
default Ubuntu apt install.
```

pgaskin:
fixes #1332

### DIFF
--- a/search.c
+++ b/search.c
@@ -22,6 +22,7 @@
 #include "ui_curses.h"
 #include "convert.h"
 #include "options.h"
+#include "debug.h"
 
 struct searchable {
 	void *data;
@@ -39,7 +40,7 @@ static int advance(struct searchable *s, struct iter *iter,
 			*iter = s->head;
 			if (!s->ops.get_next(iter))
 				return 0;
-			*wrapped = 1;
+			*wrapped += 1;
 		}
 	} else {
 		if (!s->ops.get_prev(iter)) {
@@ -48,7 +49,7 @@ static int advance(struct searchable *s, struct iter *iter,
 			*iter = s->head;
 			if (!s->ops.get_prev(iter))
 				return 0;
-			*wrapped = 1;
+			*wrapped += 1;
 		}
 	}
 	return 1;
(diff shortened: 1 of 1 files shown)


pgaskin:
This workaround is hacky, but it's lightweight and non-intrusive, and it's probably good to have it even if we fix the bug properly since the search logic is complex and there's no reason why it should wrap more than once in any case.

pgaskin:
I've opened PR #1371 with a real fix for the bug... still probably a good idea to merge both.

gavtroy:
Looks fine. If you're confident that the other fix covers all cases you could BUG_ON instead.

IMO the comment could still be shortened considering how blunt the existing codebase tends to be.

pgaskin:
> BUG_ON instead

I did that in my first iteration, but I'm really not all that confident the search isn't subtly broken somewhere else (for an unrelated reason)... it took me long enough to understand the tree search enough to fix it.

I figure that if this gets hit, we shouldn't crash cmus since we don't have to, it takes a while to get bugfixes out, and if it's broken to the point of not finding something, someone will eventually open an issue and we'll see this in the logs.

> IMO the comment could still be shortened considering how blunt the existing codebase tends to be.

The length of comments I write tends to scale with how absurd the code it's commenting is, and how long it took me to figure out what led to me writing it. I do think cmus needs some more comments in general.

Fair point, though. @flyingmutant?

pgaskin:
I edited the comment.

PR #1372, opened 2024-09-22
Fix a couple of wrong *_attr defaults

gavtroy:
I missed updating the values list when adding `color_statusline_progress_attr`. This was causing the "bold" to match with `color_win_title_attr` instead of `color_trackwin_album_attr`.


pgaskin (approved):
LGTM: NR_ATTRS==13; 13 values in the array.

PR #1377, opened 2024-09-25
README updates

### LINKED ISSUE #616: Provide compile instructions for various distributions
On various "user-friendly" distributions it's often not clear how to get the needed libraries and header files to compile cmus. We should create wiki pages that document this for at least the following distributions:

* Ubuntu

* Fedora

pgaskin:
closes #616

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-cmus — C\* Music Player
-=======================
+# cmus — C\* Music Player
 
 https://cmus.github.io/
 
@@ -10,8 +9,7 @@ Copyright © 2004-2008 Timo Hirvonen <tihirvon@gmail.com>
 Copyright © 2008-2017 Various Authors
 
 
-Configuration
--------------
+## Configuration
 
 List available optional features
 
@@ -43,8 +41,7 @@ script:
 The Tremor library is supposed to be used on hardware that has no FPU.
 
 
-Building
---------
+## Building
 
     $ make
 
@@ -53,8 +50,7 @@ Or on some BSD systems you need to explicitly use GNU make:
     $ gmake
 
 
-Installation
-------------
+## Installation
 
     $ make install
 
@@ -67,8 +63,7 @@ This is useful when creating binary packages.
 Remember to replace `make` with `gmake` if needed.
 
 
-Manuals
--------
+## Manuals
 
     $ man cmus-tutorial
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,41 +9,30 @@ Copyright © 2004-2008 Timo Hirvonen <tihirvon@gmail.com>
...
+</td></tr>
+<tr><th align="left">+ RPMFusion</th><td>dnf install faad2-devel libmp4v2-devel</td></tr>
+<tr><th align="left">Arch Linux</th><td>pacman -S pkg-config ncurses libiconv faad2 alsa-lib libao libcddb libcdio-paranoia libcue libdiscid ffmpeg flac jack libmad libmodplug libmp4v2 libmpcdec systemd opusfile libpulse libsamplerate libvorbis wavpack</td></tr>
+<tr><th align="left">Alpine</th><td>apk add pkgconf ncurses-dev gnu-libiconv-dev alsa-lib-dev libao-dev libcddb-dev libcue-dev ffmpeg-dev flac-dev jack-dev libmad-dev libmodplug-dev elogind-dev opus-dev opusfile-dev pulseaudio-dev libsamplerate-dev libvorbis-dev wavpack-dev</td></tr>
+<tr><th align="left">Termux</th><td>apt install libandroid-support ncurses libiconv ffmpeg libmad libmodplug opusfile pulseaudio libflac libvorbis libwavpack</td></tr>
+<tr><th align="left">Homebrew</th><td>brew install pkg-config ncurses faad2 libao libcddb libcdio libcue libdiscid ffmpeg flac jack mad libmodplug mp4v2 musepack opusfile libsamplerate libvorbis wavpack</td></tr>
+</tbody></table>
```
gavtroy, line 32: It formats a little differently but you can / probably should use the table syntax
```
Distro | Dependencies
:---   | :---
x | apt install blahh
```
Comparing against the arch package, you're additionally listing:
- libcddb - but I guess we can
- libcue - at a glance we don't use it?
- libiconv - uses iconv from libc, not sure what the difference is
- pkg-config - could assume it's installed
- systemd - I guess for dbus? but I don't think we should be listing itp

For Fedora, is the pkgconfig syntax needed? Would be nice to make the list more compact.

pgaskin: I considered the table syntax, but this looks nice rendered (though it is uglier raw)... which do you like best @flyingmutant?

Regarding libcue, I think you're right... the last usage of libcue was removed in 85237ec35370f0837c8be3cd1503f9164dbf6703, albeit incompletely. I'll remove it and probably open another PR to remove it entirely.

Regarding libiconv, I'll look at it closer and see if it's actually required.

I listed pkg-config on Arch since I did for the others.

Regarding systemd, it is for dbus. I think we should list it given that we've had bug reports where people are wondering why MPRIS isn't working caused by this.

The pkgconfig syntax is the standard way to do build deps on modern RPM-based distros (the actual package providing it can be split without affecting anything else). Runtime dependencies are based on the compiled ELF sonames.

pgaskin: > libiconv

You're right about it; it's included in libc. The libiconv package seems to only exist for compatibility with programs which explicitly link or load libiconv. Not including it for cmus is the correct thing to do to prevent an unnecessary dependency from being added to the binary.

> libcue

I'll remove it from the README and open another PR for it.

README.md
```diff
@@ -10,41 +9,30 @@ Copyright © 2004-2008 Timo Hirvonen <tihirvon@gmail.com>
...
--------------
+## Configuration
 
-List available optional features
-
-    $ ./configure --help
-
-Auto-detect everything
-
-    $ ./configure
```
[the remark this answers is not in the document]
pgaskin: Done.

README.md
pgaskin: Done.

PR #1382, opened 2024-10-23
Apply new formats from #1317 on upgrade

pgaskin:
New formats for the statusline, headings, and trackwin were introduced to streamline the interface and make better use of new features.

Update the config to the new defaults if we're upgrading from an old version of cmus (identified by the lack of the newly added format options in the autosave) after loading the autosave, but before loading the user's rc (yes, this does mean that if the user set a custom one via the command line rather than the rc, the user will have to reconfigure it).

See c454d4e4683143ed8fae9b4ba73ef215e8026e36, bad9b0c4291ad6a5e9f5de0e14a533ebd661ef41.

### DIFF
--- a/options.c
+++ b/options.c
@@ -299,6 +299,8 @@ static const struct {
 	{ NULL, NULL }
 };
 
+static int seen_pr1317_fmts = 0;
+
 /* callbacks for normal options {{{ */
 
 static void get_device(void *data, char *buf, size_t size)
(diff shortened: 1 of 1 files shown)


pgaskin (commented):
Tested by:
- Building and running cmus v2.11.0.
- Building and running this PR, seeing the new layout.
- Running `:set format_statusline=test`.
- Restarting cmus and seeing the customized format persist.
- Doing all of this again with a fresh config, but putting the set command in the rc instead.

gavtroy (approved):
nit: `seen_v2.12_fmts` might be more meaningful to future readers

Otherwise looks fine, works fine, and it shouldn't be too big of a deal for advanced users to reapply customizations.

PR #1386, opened 2024-11-29
Set the minimum requested AAudio buffer capacity

pgaskin:
This fixes buffer underruns on the Pixel 9 due to the default selected buffer size being faster than cmus can produce audio.

There's probably a better way to deal with this, but it works.


pgaskin:
To test this easily, you can download the ndk build artifact and place `aaudio.so` in `/data/data/com.termux/files/usr/lib/cmus/op/`.

PR #1388, opened 2024-12-08
Check for libavutil pkg-config

pgaskin:
It's required for av_strerror.


pgaskin (commented):
Hopefully this doesn't break builds on older versions of ffmpeg, speaking of which, we should automatically test building ip/ffmpeg on all supported ffmpeg versions.

PR #1405, opened 2025-02-10
PulseAudio plugin: drain stream upon closing; change codestyle

ihy123:
I noticed that PulseAudio plugin stops playback before the stream has ended (at least 0.2 sec remains), so here's the fix.


gavtroy:
There's a history to why it's not drained, see https://github.com/cmus/cmus/commit/7adf497597e7ca83c866bce61c4ca7c2ac4a72d5. At a glance it seems like this would reintroduce the "hang at quit while paused" issue? The information that the stream stops early is interesting though, it's not something I've noticed.

ihy123:
You're right. I thing the following algorithm will solve it:
1. is stream corked or suspended or state != ready?
2. If it is then just quit
3. If it is not then drain the stream.

Also we can add a time event (pa_context_rttime_new) to stop draining after some time.
Determine the time using
pa_bytes_to_usec(pa_context_get_tile_size()).

I don't think adding a timeout is necessary though since we know that the stream is not paused.


I use pipewire with pulseaudio frontend, so it might handle closing differently.

ihy123:
I realized that I shouldn't have removed the `pa_` namespace from function names because it makes debug logging less informative.

Looks like at this point it's better to rewrite commit history. But will github accept `git push -f` or creating another pull request is required? This is my first pull request so I don't know.

Aside from this mess, all mentioned issues are fixed.

gavtroy:
push -f is okay. For the record it probably would have been better to keep the refactoring in a separate PR to make it easier to review and potentially object to (although it looks fine to me).

ihy123:
Ok, so I added `pulse_` namespace to most functions and cleaned up commit history. All done

pgaskin:
Upon a first glance, it seems sane enough, and the commits are split well. I'll do some testing later this week.

pgaskin:
Playback and pause/stop works fine for me in Termux (with PulseAudio 17.0) and Fedora 41 (with PipeWire 1.2.7). Still need to test exiting in various conditions.

pgaskin:
@flyingmutant rebase merge?

flyingmutant:
Thanks! In the future, it is probably best to avoid "style" reworks because they make history/blame harder to follow, and also make the review itself harder (although in this case the commits are well-split indeed).

ihy123:
Oh, I didn't think of the review process. I agree that style changes were unnecessary for this one and will avoid such changes in the future.

Issue #1397, opened 2025-01-09
No gapless playback and bad seeking for mka (matroska) file with cue sheet.


BoostCookie:
I have an album as Dolby Atmos in an mka container. To give cmus the information about the timestamps of the individual songs I have a cue file right next to it.

The seeking to the timestamps of the cue sheet is always a bit off. When selecting a song I often start around 1 second inside the song. This also leads to an album not playing gapless between songs which can sound annoying.

I know that my cue sheet has the correct timestamps because when I convert the Dolby Atmos stream to a flac and use the same exact cue sheet, the seeking to timestamps is perfect and gapless playback therefore works flawlessly.

1) `uname -a`:
```
Linux my-pc 6.6.69-1-lts #1 SMP PREEMPT_DYNAMIC Thu, 02 Jan 2025 22:00:49 +0000 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.12.0
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors
```

3) `cmus --plugins`:
```
Input Plugins: /usr/lib/cmus/ip
  mad:
    Default Priority: 55
    File Types: mp3 mp2
    MIME Types: audio/mpeg audio/x-mp3 audio/x-mpeg
  aac:
    Default Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  cdio:
    Default Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  cue:
    Default Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Default Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Default Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  mpc:
    Default Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  opus:
    Default Priority: 50
    File Types: opus
    MIME Types:
  vorbis:
    Default Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  wav:
    Default Priority: 50
    File Types: wav
    MIME Types:
  wavpack:
    Default Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  ffmpeg:
    Default Priority: 30
[...]

PR #1406, opened 2025-02-12
Some improvements to Cue handling

ihy123:


ihy123:
I suspect that #1397 will be fixed by this pull request. Can't say for sure because there's no Cue file attached to it.

gavtroy:
I only lightly reviewed but it looks sane. One part that that needs consideration is how .cue files are now added directly instead of magically substituted when another format with a matching filename is added. I do like adding the .cue directly but you probably need to keep and invert the "matching filename" logic to instead exclude the media file. Otherwise `:add ~/Music` will also add the un-cut-up .flac (or whatever).

ihy123:
Yeah, I think the only way to correctly add a directory with Cue files is to
1. add all Cue files and exclude audio files which were mentioned in these
2. add all remaining files

I checked Audacious source and they implement it in the same way.

Also, am I right that playlist files (.pl, .m3u) don't need this special treatment?

BoostCookie:
> I suspect that #1397 will be fixed by this pull request. Can't say for sure because there's no Cue file attached to it.

Hello. I opened this issue.
I just tried out your merge request and unfortunately the problem still persists.
In the attachments you can find my two example cue files and the ffprobe output of the corresponding mka file. One has TrueHD audio with Atmos, the other one has DTS. If you send me an email I can provide you with the mka files.
[examples.zip](https://github.com/user-attachments/files/18812271/examples.zip)

ihy123:
I tried this on a directory with 100+ Cue files in different subdirs and it worked, so it's ready for review

pgaskin (approved):
Seems fine to me. I only really looked at the scan logic, as I'm not qualified to test or review the cue stuff (and someone else has already tried it), and since the scanning works, the impact on other formats is minimal.

flyingmutant:
I am almost sure `associated_cue` mechanism that 791d190b532eef4d833d0325e8f56b069168b096 removes served some real purpose, but it has been a long time since I've used cue files in any form. Anyone actively using cue care to weight in?

flyingmutant:
@ihy123 can you please explain the logic behind 6d0cd8f49bef4d978a075a1a601bc8bfda7cce03?

ihy123:
> I am almost sure `associated_cue` mechanism that [791d190](https://github.com/cmus/cmus/commit/791d190b532eef4d833d0325e8f56b069168b096) removes served some real purpose, but it has been a long time since I've used cue files in any form. Anyone actively using cue care to weight in?

This comment explains its purpose well https://github.com/cmus/cmus/pull/1406#issuecomment-2661018642
But https://github.com/cmus/cmus/pull/1406/commits/5b98d2932fc5d7ad4ee3a49a9b37a9ab5045d125 + https://github.com/cmus/cmus/pull/1406/commits/6d0cd8f49bef4d978a075a1a601bc8bfda7cce03 remove the need for it

ihy123:
> @ihy123 can you please explain the logic behind [6d0cd8f](https://github.com/cmus/cmus/commit/6d0cd8f49bef4d978a075a1a601bc8bfda7cce03)?

Sure:
From `add_dir()` we call:
- `dir_list_recursive()` to do just that
- `ptr_array_sort()`, `ptr_array_unique()` to
  - remove duplicates. Don't remember why exactly (something with symlinks?), might be unnecessary
  - for binary search that `handle_cue_files()` does
- `handle_cue_files()` to remove files that are mentioned in cues from the whole list

This is not an elegant solution, but I don't know of a better one yet. My main goal was to remove the associated cue mechanism, because it's counter-intuitive (see commit message of https://github.com/cmus/cmus/pull/1406/commits/a87da7e51e43fbe50ff16b19eefebf72b28d5638)

ihy123:
mp3 files mentioned in cues will be added as cue items, standalone files will be added as standalone files.

BoostCookie:
My issue https://github.com/cmus/cmus/issues/1397 was fixed by https://github.com/cmus/cmus/pull/1444, btw.

Issue #1407, opened 2025-02-23
MPRIS: `Seeked` signals report old position instead of new


Endg4meZer0:
Hey! I was in the middle of refactoring a Go library for MPRIS, when I noticed, that in some cases cmus reports wrong position when firing Seeked signals. 

In my observations, it only affects the following situation: song is in the middle of playing (e.g. the position is 0:05); seeking in any direction results in a signal with body `int64 5000000` instead of expected `int64 0` or `int64 10000000`. But, for example, the "Seeked" signal after a song switch works correctly (even though it's not in the spec) and displays `int64 0` in the body. However I have not tested the behavior in any other cases.

1. `uname -a`
```
Linux endg4me-pc 6.13.3-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Mon, 17 Feb 2025 17:41:56 +0000 x86_64 GNU/Linux
```
2. `cmus --version`
```
cmus v2.12.0
Copyright 2004-2006 Timo Hirvonen
Copyright 2008-2016 Various Authors
```
3. `cmus --plugins`
```
Input Plugins: /usr/lib/cmus/ip
  aac:
    Default Priority: 50
    File Types: aac
    MIME Types: audio/aac audio/aacp
  cdio:
    Default Priority: 50
    File Types:
    MIME Types: x-content/audio-cdda
  cue:
    Default Priority: 50
    File Types:
    MIME Types: application/x-cue
  flac:
    Default Priority: 50
    File Types: flac fla
    MIME Types:
  modplug:
    Default Priority: 50
    File Types: mod s3m xm it 669 amf ams dbm dmf dsm far mdl med mtm okt ptm stm ult umx mt2 psm
    MIME Types:
  mpc:
    Default Priority: 50
    File Types: mpc mpp mp+
    MIME Types: audio/x-musepack
  opus:
    Default Priority: 50
    File Types: opus
    MIME Types:
  vorbis:
    Default Priority: 50
    File Types: ogg oga ogx
    MIME Types: application/ogg audio/x-ogg
  wav:
    Default Priority: 50
    File Types: wav
    MIME Types:
  wavpack:
    Default Priority: 50
    File Types: wv
    MIME Types: audio/x-wavpack
  ffmpeg:
    Default Priority: 30
[...]

PR #1408, opened 2025-02-25
fix: MPRIS seeked signals reporting old position instead of new

Endg4meZer0:
Fixing #1407, probably in a hacky way. Would've been great if someone knowledgeable took a look and said if it's OK as it is.

On that note I would also like to ask: why don't we make the seeking strict to integers, or the seeked signal to be sent with the most possible accuracy? Just curious, as I would love to see the consistency between the arrived info from DBus and the actual seeked position.

### DIFF
--- a/player.c
+++ b/player.c
@@ -74,6 +74,7 @@ static struct player_info player_info_priv = {
 	.metadata_changed = 0,
 	.status_changed = 0,
 	.position_changed = 0,
+	.position_seeked = 0,
 	.buffer_fill_changed = 0,
 };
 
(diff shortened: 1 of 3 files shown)


gavtroy (approved):
I'm not the someone knowledgeable you mentioned but it looks good to me.

pgaskin (commented):
Looks fine to me. I'll test this later.

On a side note, we might want to consider debouncing MPRIS signals.

Endg4meZer0:
Hey, is there any update on it?

pgaskin:
Not yet, I tend to do cmus maintenance in bursts since I don't currently have too much time to devote to it.

pgaskin (changes requested):
PR looks fine wrt concurrency and locking. It also happens to fix the consumer loop potentially getting blocked on a dbus emit of a seeked event.

The flag implementation looks correct overall (it can't get dropped mistakenly), but there's a few other minor issues:

- Position may still be old if changed between `player_unlock` and the next snapshot.
- Not safe to call `player_seek` from other threads.
- Duplicate seeked event on startup if `resume` is enabled.

Side note: this change also means that failed seeks don't emit seeked, but that makes sense.

PR #1416, opened 2025-03-19
Try to fallback to any plugin upon failure to init output plugin

WalrusDevil:
Bad values for output_plugin set in ~/.config/cmus/rc or ~/.config/cmus/autosave prevent output_plugin from being initialized.

This fixes that.

### DIFF
--- a/player.c
+++ b/player.c
@@ -1313,7 +1313,7 @@ void player_seek(double offset, int relative, int start_playing)
  */
 void player_set_op(const char *name)
 {
-	int rc;
+	int rc = 0;
 
 	player_lock();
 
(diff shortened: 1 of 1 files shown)


pgaskin:
I personally think we should still error for rc to keep it consistent with the other commands. @gavtroy, any opinion?

Probably add a global flag `is_loading_autosave` and set it before and after the autosave load in `options.c`, and check for it around the output plugin stuff here.

gavtroy:
You mean allow fallback to `op_select_any()` only on startup but keep strict behavior for an interactive "set output_plugin" command? Sure. In that case `ui_initialized` probably works fine to make the distinction.

WalrusDevil:
I think this already gives most of the desired behavior, my comment might be misleading.

If there is a bad value from autosave or rc it still errors, but calls `op_select_any()` since op is not init'd

Afterwards, `set output_plugin` would only ever call `op_select_any()` if the startup `op_select_any()` failed.

But I can add a check for `ui_initialized` so `op_select_any()` only gets called at startup.

PR #1427, opened 2025-05-10
Clear the marked playlist when focus is in the playlist list

### LINKED ISSUE #1423: Enable clearing a playlist when its title is selected (left view), not just when the songs it contains (right view) are selected
In order to clear a playlist (using the :clear command) in the third panel the focus needs to be on some song within the playlist (right view) and not on the playlist itself (left view). This is disorientating and unpractical.

How to fix : allow the :clear command to work primarily from the left view's selection of playlists by title.

jdholtz:
Fixes #1423. This PR allows users to clear the marked playlist when focused in the playlist list (left view). 

I opted to clear the marked playlist rather than the visible playlist as this has the same functionality as the `add` command. However, when selected the playlist that is visible will still be cleared. This may be confusing to users though, so I'm up for changing this to always clear the visible playlist (frankly, I think add working on marked playlists instead of visible playlists is confusing).

### Implementation notes
Although `pl_cursor_in_track_window` is currently always true when `pl_clear_marked_pl` is called, I still kept the check in there in case it will be used for other situations later and to keep congruency with `pl_clear_visible_pl`. I can remove this check if it's deemed unnecessary and just note that `pl_cursor_in_track_window` is assumed to be true in a comment.

### DIFF
--- a/pl.c
+++ b/pl.c
@@ -546,6 +546,16 @@ static void pl_clear_visible_pl(void)
 	pl_cancel_add_jobs(pl_visible);
 }
 
+static void pl_clear_marked_pl(void)
+{
+	if (pl_cursor_in_track_window)
+		pl_win_next();
+	if (pl_marked == pl_playing)
+		pl_playing_track = NULL;
+	editable_clear(&pl_marked->editable);
+	pl_cancel_add_jobs(pl_marked);
+}
+
 static int pl_name_exists(const char *name)
 {
 	struct playlist *pl;
@@ -795,10 +805,10 @@ void pl_rename_selected_pl(const char *name)
 
 void pl_clear(void)
 {
-	if (!pl_cursor_in_track_window)
-		return;
-
-	pl_clear_visible_pl();
+	if (pl_cursor_in_track_window)
+		pl_clear_visible_pl();
+	else
+		pl_clear_marked_pl();
 }
 
 void pl_mark_for_redraw(void)


pgaskin:
> frankly, I think add working on marked playlists instead of visible playlists is confusing

I generally agree with this too, except for when cmus is automated through cmus-remote.

It might be worth adding a configuration option to keep the marked playlist in sync with the selected one.

pgaskin (changes requested):
I think this should be conditional on a flag for the clear command to avoid breaking expectations for automation and to keep it consistent with the other commands. I'm not fully decided though, so I'm open to arguments in the other direction.

jdholtz:
> It might be worth adding a configuration option to keep the marked playlist in sync with the selected one.

Yeah, that makes sense to me. I can look into doing this if it’s desired.

> I think this should be conditional on a flag for the clear command to avoid breaking expectations for automation and to keep it consistent with the other commands

Is this in reference to the current changes or the configuration option proposal? If it’s the former, what are expectations for automation that these changes break? If it is the latter, wouldn’t the config option not affect cmus-remote? (I’m not too familiar with how cmus-remote works related to the GUI).

pl.c
```diff
@@ -546,6 +546,16 @@ static void pl_clear_visible_pl(void)
 	pl_cancel_add_jobs(pl_visible);
 }
 
+static void pl_clear_marked_pl(void)
+{
+	if (pl_cursor_in_track_window)
+		pl_win_next();
```
pgaskin: Drop this. It's unreachable, and it would be wrong even if it was since `pl_win_next` only makes sense when the visible playlist was the one emptied.

pgaskin (changes requested):
Let's:

- Make `:clear -p` clear the marked playlist unconditionally (so `-c` works properly and matches `add -p`). Since `flag_to_view` collapses explicit flags and current view into the same value, `view_clear` will need to know which case it is in.
- Make `:clear` in the playlist view clear the visible (selected) playlist regardless of pane (for #1423, and so it matches `win-remove`/`win-update`/`pl-rename`/`pl-export`).

pgaskin:
> expectations for automation that these changes break

Nothing breaks that wasn't already broken, to be precise, but `cmus-remote -c FILE` sends `clear -p` then `add -p FILE`. `add -p` always targets the marked playlist. `clear -p` currently targets the visible playlist, and only when the cursor is in the track window. With this PR it targets the visible playlist in the track window and the marked playlist in the left pane. Either way a remote `clear -p` depends on the currently focused view, resulting in `-c` behaving unexpectedly.

I'll add the config option I mentioned as a sepearate PR.

PR #1439, opened 2025-07-15
mpris: set xesam:url tag

LunaPresent:
This PR makes cmus set the xesam:url tag as described in the [mpris spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#xesam:url). This makes it easier for external scripts and applications to interact with the currently playing song, e.g. finding the matching .lrc file.

### DIFF
--- a/mpris.c
+++ b/mpris.c
@@ -423,6 +423,24 @@ static int mpris_metadata(sd_bus *_bus, const char *_path,
 			CK(mpris_msg_append_sas_dict(reply,
 					"xesam:comment", corrected));
 		}
+		if (ti->filename) {
+			if (is_url(ti->filename)) {
+				char corrected[u_str_print_size(ti->filename)];
+				u_to_utf8(corrected, ti->filename);
+				CK(mpris_msg_append_sas_dict(reply,
+						"xesam:url", corrected));
+			}
+			else {
+				// prepend schema for local files as described in the spec
+				// https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#uri
+				const char prefix[] = "file://";
+				char corrected[sizeof(prefix) - 1 + u_str_print_size(ti->filename)];
+				memcpy(corrected, prefix, sizeof(prefix) - 1);
+				u_to_utf8(corrected + sizeof(prefix) - 1, ti->filename);
+				CK(mpris_msg_append_sas_dict(reply,
+						"xesam:url", corrected));
+			}
+		}
 		if (ti->bpm != -1)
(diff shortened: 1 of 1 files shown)


mpris.c
```diff
@@ -423,6 +423,24 @@ static int mpris_metadata(sd_bus *_bus, const char *_path,
...
+			}
+			else {
+				// prepend schema for local files as described in the spec
+				// https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#uri
+				const char prefix[] = "file://";
+				char corrected[sizeof(prefix) - 1 + u_str_print_size(ti->filename)];
+				memcpy(corrected, prefix, sizeof(prefix) - 1);
+				u_to_utf8(corrected + sizeof(prefix) - 1, ti->filename);
+				CK(mpris_msg_append_sas_dict(reply,
+						"xesam:url", corrected));
```
pgaskin, line 441: I believe this needs URL escaping (except for slashes). It's not clear based on the MPRIS spec, but at least [GLib](https://gitlab.gnome.org/GNOME/glib/-/commit/39864716d6404766c325caf18c6bef722dc060cd) and [browsers](https://url.spec.whatwg.org/#file-state) will parse/strip query strings and fragments.

pgaskin (changes requested):

mpris.c
pgaskin: Yeah, it definitely needs URL escaping; playerctl and KDE depend on it, and most other players escape it too.

mpris.c
```diff
@@ -423,6 +423,24 @@ static int mpris_metadata(sd_bus *_bus, const char *_path,
...
+			}
+			else {
+				// prepend schema for local files as described in the spec
+				// https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#uri
+				const char prefix[] = "file://";
+				char corrected[sizeof(prefix) - 1 + u_str_print_size(ti->filename)];
+				memcpy(corrected, prefix, sizeof(prefix) - 1);
+				u_to_utf8(corrected + sizeof(prefix) - 1, ti->filename);
+				CK(mpris_msg_append_sas_dict(reply,
```
pgaskin: Should be ss, not sas.

pgaskin (changes requested):
I checked some other MPRIS implementations and added a few comments.

In addition, we should use the original string and not do additional normalization before encoding it.

Also, this doesn't handle cdda and cue URLs, but tha can be done separately.

pgaskin:
I am going to reimplement this from scratch myself.

PR #1454, opened 2025-09-28
ip/ffmpeg: fix seeking and compiler warning about %lld

ihy123:
### About %lld

[3157b78](https://github.com/cmus/cmus/commit/3157b788a7eeb8d1a02d6dad2c05fde6f900f4aa) introduced a warning about %lld on GCC. @thecloudexpanse try this PR with your compiler please

### Results of the seeking fix
Before:
```
ffmpeg_calc_skip_samples: seek_ts: 98.673288s, frame_ts: 97.985306s
ffmpeg_skip_frame_part: skipping frame: 1152 samples
... 24 more ...
ffmpeg_skip_frame_part: skipping frame: 1152 samples
ffmpeg_skip_frame_part: skipping 388 samples
```
After:
```
ffmpeg_calc_skip_samples: seek_ts: 98.673288s, frame_ts: 98.664490s
ffmpeg_skip_frame_part: skipping 388 samples
```

### DIFF
--- a/ip/ffmpeg.c
+++ b/ip/ffmpeg.c
@@ -437,7 +437,7 @@ static int ffmpeg_seek(struct input_plugin_data *ip_data, double offset)
 
 	priv->seek_ts = offset;
 	priv->skip_samples = 0;
-	int64_t ts = av_rescale(offset, st->time_base.den, st->time_base.num);
+	int64_t ts = offset / av_q2d(st->time_base);
 
 	int ret = avformat_seek_file(priv->format_ctx,
 			priv->stream_index, 0, ts, ts, 0);


thecloudexpanse (approved):
Explicitly casting to (long long) fixes the issue, so the value will match the format specifier. Apologies for missing that. I verified that this compiles cleanly with clang.

For what it's worth, changing the format specifier back to %ld and casting the value to (long) also eliminates the original warning on clang. However, since the value is a 64-bit int, I think (long long) is probably preferred.

pgaskin (approved):
Seeking works fine with fewer discards required. If I'm understanding/observing it correctly, it's more of an optimization than a fix (so I'd probably recommend updating the commit message to say that). Thoughts, @ihy123?

Format string thing is fine (and it's for debug logs only anyways). The correct solution would be to use `PRId64`, but cmus doesn't really do that.

ihy123:
> changing the format specifier back to %ld and casting the value to (long)

> The correct solution would be to use `PRId64`

IMO it's what works best for compatibility. `skip_samples` would easily fit into an `int` and %d could be used. I honestly don't care, so whatever you say, @pgaskin.

> so I'd probably recommend updating the commit message to say that

No problem

pgaskin (approved):

PR #1455, opened 2025-10-01
Natural sort playlist names in pl_list_compare

carterww:
Playlists with equal names up to a number are now displayed in numerical ascending order.

Previously:
- Volume 11
- Volume 25
- Volume 9

Now:
- Volume 9
- Volume 11
- Volume 25

Leading zeros are ignored, but items with more leading zeros (but equal otherwise) appear later in the list.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -19,6 +19,7 @@
 #include "misc.h"
 #include "prog.h"
 #include "xmalloc.h"
+#include "utils.h"
 #include "xstrjoin.h"
 #include "ui_curses.h"
 #include "config/libdir.h"
@@ -26,6 +27,7 @@
 
 #include <string.h>
 #include <stdlib.h>
+#include <stdarg.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


gavtroy:
When I almost implemented something for this problem I was thinking in the direction of https://github.com/gavtroy/cmus/commit/040ce1766403da9c49885a08287a0a3e9a446b3f. But I got upset about not respecting locale and just gave up. :) Even so I don't necessarily object to this change.

As for the algorithm, without looking too closely it's probably essentially "good enough"... but definitely not perfect, for example 11 > 100.

You can also undo the `&`s in `strptrnatcmp(&pl->name, &pr->name);` and just remove the extra dereference from the function.

carterww:
Thank you for the feedback!

After taking a look at this again, I agree that the algorithm is not adequate and has bugs. I will make a new commit that addresses the 11 > 100 issue and some of the other issues I recently noticed.

I can also try to respect locale, but I don't think the implementation will be clean. I was thinking of using strcoll to compare substrings in the name without numbers. strcoll expects null terminated strings (and doesn't take a length) so I'd need to temporarily overwrite a character in playlist->name to null terminate the substring. I'm not sure if this would mess with any other threads trying to read playlist->name. The numbers in the string can be compared by treating them as semantic versions.

If this sounds okay, I'll get to work on it!

gavtroy:
I didn't mean to sound like I was demanding locale support, although by all means feel free to try it. It's also fine to put it in a future PR.

PR #1458, opened 2025-10-07
Fix libmp4v2 check

gavtroy:
The modern gcc default is -std=gnu23 which the mp4v2 headers are not compatible with. This was only a problem for the check, since for the actual compile we add EXTRA_CFLAGS that set -std=gnu11 or -std=c11 anyway.


pgaskin (approved):
Not tested, but seems fine.

PR #1462, opened 2025-10-29
op/alsa: remove polling for events before pause (fixes #1461)

### LINKED ISSUE #1461: Fast pause/play causes deadlock
## Steps to reproduce
1. start playing a song
2. hold 'c' to rapidly pause/unpause. Pressing 'x' or 'v' and switching tracks might also help

## Result
UI is unresponsive. As we can see, all threads are stuck waiting for something.
Image

## Info
cmus version: master branch
ip: flac
op: alsa

ihy123:
Checking the call stack after the player hangs shows that main thread is blocked by `snd_pcm_wait()` which is waiting for events.
main_thread
I just removed it, there is no apparent reason for it to be there.

### DIFF
--- a/job.c
+++ b/job.c
@@ -328,6 +328,8 @@ static void handle_cue_files(struct ptr_array *files)
 
 		char **files_in_cue;
 		int n = cue_get_files(ents[i], &files_in_cue);
+		if (n == -1) 
+			continue;
 		char *cue_dir = path_dirname(ents[i]);
 
 		for (j = 0; j < n; j++) {


ihy123:
Another issue randomly popped up after this change... So there was some reason for the polling

ihy123:
The issue is that after pausing several times the playback just won't unpause even though `snd_pcm_pause()` returns successfully and changes alsa state to `RUNNING`. Same thing in Audacious. I am using a hardware ALSA device. No such issue on `dmix` audio device, or on pipewire's ALSA emulation. So this is either ALSA or driver problem.

ihy123:
> The issue is that after pausing several times the playback just won't unpause

This issue just disappeared

pgaskin (approved):
Seems correct since snd_pcm_ready is [apparently](https://www.mail-archive.com/alsa-devel@lists.sourceforge.net/msg12048.html) for waiting for room to write. Other projects don't [seem](https://github.com/search?q=snd_pcm_wait+snd_pcm_pause) to call snd_pcm_wait before snd_pcm_pause.

I didn't test this change.

PR #1463, opened 2025-11-01
Make sure to not try freeing an uninitialized pointer

xant:
While playing around with cmus and trying to load my Music archive I stumbled upon this bug ... reported as double free but actually trying to free an unitialized pointer.

### DIFF
--- a/job.c
+++ b/job.c
@@ -328,6 +328,8 @@ static void handle_cue_files(struct ptr_array *files)
 
 		char **files_in_cue;
 		int n = cue_get_files(ents[i], &files_in_cue);
+		if (n == -1) 
+			continue;
 		char *cue_dir = path_dirname(ents[i]);
 
 		for (j = 0; j < n; j++) {


job.c
```diff
@@ -326,7 +326,7 @@ static void handle_cue_files(struct ptr_array *files)
 		if (!ents[i] || !is_cue(ents[i]))
 			continue;
 
-		char **files_in_cue;
+		char **files_in_cue = NULL;
 		int n = cue_get_files(ents[i], &files_in_cue);
```
pgaskin: The uninitialized pointer happens when cue_get_files returns -1, which happens when cue_from_file returns NULL, which happens if mmap_file fails or the cue parse returns an error.

The fix should be to add an `if (n == -1) continue` after the cue_get_files call.

pgaskin (changes requested):

pgaskin (approved):
Looks good.

PR #1464, opened 2025-11-03
Reaching the end of the command list is not an error

xant:
the commands array is terminated with a null-filled structure, reaching the end shouldn't be reported as an error.

### DIFF
--- a/ip/flac.c
+++ b/ip/flac.c
@@ -117,7 +117,6 @@ static T(TellStatus) tell_cb(const Dec *dec, uint64_t *offset, void *data)
 	struct input_plugin_data *ip_data = data;
 	struct flac_private *priv = ip_data->private;
 
-	d_print("\n");
 	*offset = priv->pos;
 	return E(TELL_STATUS_OK);
 }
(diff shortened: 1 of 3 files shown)


xant:
> How are you making this happen? I wasn't able to reproduce it, and I tried a few things. Also, this change removes the only `unknown command` message, which is supposed to be there for an unknown command.

I was sending three chained commands using cmus-remote.
I need to get back to the system where this was happening, I can give you more details later today

pgaskin:
I tried `cmus-remote` with multiple `-C` options earlier. Can I see the actual `cmus-remote` command? Feel free to remove the file paths.

xant:
maybe I just don't need multiple "-C" ...

Process 101946 stopped
* thread #1, name = 'cmus', stop reason = breakpoint 2.1
    frame #0: 0x000055555ef51320 cmus`run_parsed_command(cmd="-C", arg=0x0000000000000000) at command_mode.c:2940:4
   2937                 const struct command *c = &commands[i];
   2938
   2939                 if (c->name == NULL) {
-> 2940                         error_msg("unknown command\n");
   2941                         break;
   2942                 }
   2943                 if (strncmp(cmd, c->name, cmd_len) == 0) {
(lldb) p cmd
(char *) 0x00005555961f8e40 "-C"
(lldb) 


but I'm pretty sure I tried this already and  if I don't pass multple -C it won' t run all the commands.

xant:
I think the problem is in server.c ... it shouldn't call run_parsed_command() on "-C" ...

xant:
according to the documentation only one '-C' should be used:

"If *-C* is given, all command line arguments are treated as raw commands.  "

and it seems indeed that running : `cmus-remote -C "view 3" "/Puglia"` works.

I'll close this PR and move forward fixing the way I was calling cmus-remote.

PR #1469, opened 2025-11-16
added the alias command to cmus

C4N-6:
with the command 
```
:alias name_of_alias=some command
```

you can create  a alias to a command

the next time you type that alias in it will run that command

### DIFF
--- a/alias.c
+++ b/alias.c
@@ -1,0 +1,66 @@
+#include "alias.h"
+#include "uchar.h"
+#include "xmalloc.h"
+#include <stdlib.h>
+
+struct alias_node *alias_list = NULL;
+
+struct alias *get_alias(char *name)
+{
+	int name_len = strlen(name);
+
+	for (struct alias_node *i = alias_list; i != NULL; i = i->next) {
+		if (strncmp(i->alias.name, name, name_len) == 0) {
+			return &i->alias;
+		}
+	}
+
+	return NULL;
+}
+
+void add_alias(char *name, char *command)
+{
+	struct alias *duplicate = get_alias(name);
+	if (duplicate == NULL) {
+		struct alias_node *new = xnew(struct alias_node, 1);
+		new->alias.name = xstrdup(name);
+		new->alias.command = xstrdup(command);
+		new->next = alias_list;
+		alias_list = new;
+	} else {
+		free(duplicate->command);
+
+		duplicate->command = xstrdup(command);
+	}
+}
+int delete_alias(char *name)
+{
+	if (alias_list == NULL) {
+		return 0;
+	}
+	int name_len = strlen(name);
+
+	if (strncmp(alias_list->alias.name, name, name_len) == 0) {
+		free(alias_list->alias.name);
+		free(alias_list->alias.command);
+		free(alias_list);
+
+		alias_list = NULL;
+		return 1;
+	}
+
+	for (struct alias_node *i = alias_list; i->next != NULL; i = i->next) {
+		if (strncmp(i->next->alias.name, name, name_len) == 0) {
+			struct alias_node *node_to_be_free = i->next;
+
+			i->next = node_to_be_free->next;
+
+			free(node_to_be_free->alias.name);
+			free(node_to_be_free->alias.command);
+			free(node_to_be_free);
+
+			return 1;
+		}
+	}
+	return 0;
+}
--- a/alias.c
+++ b/alias.c
@@ -7,10 +7,8 @@ struct alias_node *alias_list = NULL;
 
 struct alias *get_alias(char *name)
 {
-	int name_len = strlen(name);
-
 	for (struct alias_node *i = alias_list; i != NULL; i = i->next) {
-		if (strncmp(i->alias.name, name, name_len) == 0) {
+		if (strcmp(i->alias.name, name) == 0) {
 			return &i->alias;
 		}
 	}
--- a/command_mode.c
+++ b/command_mode.c
@@ -368,6 +368,11 @@ static void cmd_add(char *arg)
 	view_add(flag_to_view(flag), arg, flag == 'Q');
 }
 
+static void cmd_add_alias(char *arg)
+{
+
+}
+
 static void cmd_clear(char *arg)
 {
 	int flag = parse_flags((const char **)&arg, "lpq");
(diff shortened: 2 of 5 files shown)


gavtroy:
As the formatting is broken it can't really be reviewed.

But I might question how useful an alias feature is in the first place. What is the use case?

> \* added a nix build file

I'm not sure if you intended to add this change, but if you did it needs to be a separate pull request.

C4N-6:
I am sorry about the formatting, the reason why I added this in was because I had wrote a shell script that increased the volume gradually, but I found no way to make it act like a normal cmus command.

so I made this so that I could just use `:vol <volume>` instead of `:shell ~/scripts/cmus-volume.sh <volume>`

I am sorry about the nix build file, I kinda thought that this was never going to get merged so I just started using my own branch and had forgotten that I had created a pull request, if you are interested in merging this I can some how get rid of that.

C4N-6:
I have rebased my pull request so the formatting error and the nix file are removed

gavtroy:
Tabs are still being replaced with spaces. If you look through the "Files changed" tab here on github the indentation is quite broken.

> so I made this so that I could just use :vol <volume> instead of :shell ~/scripts/cmus-volume.sh <volume>

That sounds like a fair use case. For the record you could also set it as a keybind like `:bind common V push shell ~/scripts/cmus-volume.sh `.

> also I could not figure out how to make the aliases persistent on restart

Shouldn't be too hard, you really just have to print the command to recreate them. As a reference: https://github.com/cmus/cmus/blob/d335e90abd5d000eef128debcb979d852f0f6b7f/options.c#L1908-L1916

C4N-6:
I pretty sure I have replaced all the spaces with tabs

C4N-6:
there where more tabs that I did not see.

I also made it save the aliases

please tell me if I have any other mistakes

C4N-6:
I new about the bind command but that dose not accept parameters right?

C4N-6:
I am looking into adding an expansion function to the alias

C4N-6:
should I add CMD_UNSAFE to alias

C4N-6:
okay I have finished everything I want to add.

please tell me if you want to add CMD_UNSAFE to alias command.

I will be using my branch to see if I come across any bugs.

if you want me to change anything please leave a comment

alias.c
```diff
@@ -0,0 +1,64 @@
...
+		free(alias_list->alias.name);
+		free(alias_list->alias.command);
+		free(alias_list);
+
+		alias_list = NULL;
+		return 1;
+	}
+
+	for (struct alias_node *i = alias_list; i->next != NULL; i = i->next) {
+		if (strncmp(i->next->alias.name, name, name_len) == 0) {
```
gavtroy, line 51: Get rid of the special case and handle everything in the for loop.

```
for (...; i != NULL; ...)
  if (strcmp(i->alias.name ...
```

"strncmp" is probably wrong here because it would be a prefix check.

command_mode.c
```diff
@@ -368,6 +369,42 @@ static void cmd_add(char *arg)
...
+	for (i = 0; arg[i]; i++) {
+		if (arg[i] == '=') {
+			arg[i] = 0;
+			value = &arg[i + 1];
+			has_value = true;
+			break;
+		}
+	}
+
+	if (!has_value) {
```
gavtroy, line 387: it's fine but I don't think this is clearer than just checking if value is NULL

command_mode.c
```diff
@@ -368,6 +369,42 @@ static void cmd_add(char *arg)
...
+		struct alias *found_alias = get_alias(arg);
+		if (found_alias == NULL) {
+			error_msg("Error: no such alias %s\n", arg);
+		} else {
+			info_msg("alias: '%s=%s'\n", arg, found_alias->command);
+		}
+		return;
+	} else if (*value != '\0') {
```
gavtroy, line 395: Since you return above, it would be cleaner to make this a new "if".

command_mode.c
```diff
@@ -368,6 +369,42 @@ static void cmd_add(char *arg)
...
+		add_alias(arg, value);
+		return;
+	}
+	if (!delete_alias(arg)) {
+		error_msg("Error: no such alias %s\n", arg);
+		return;
```
gavtroy, line 401: remove this return and the one a few lines down

command_mode.c
```diff
@@ -2917,11 +3039,36 @@ int parse_command(const char *buf, char **cmdp, char **argp)
...
+	if (found_alias != NULL) {
+		int return_val;
+
+		if (arg_start == arg_end) {
+			return_val = parse_command(found_alias->command, cmdp, argp);
+		} else {
+			char *argument = xstrndup(buf + arg_start, arg_end - arg_start);
+			char *new_buf = xstrjoin(found_alias->command," ", argument);
+
+			return_val = parse_command(new_buf, cmdp, argp);
```
gavtroy, line 3055: No reason to strdup the argument first.

Should add a recursion depth counter and limit.

command_mode.c
```diff
@@ -2917,11 +3039,36 @@ int parse_command(const char *buf, char **cmdp, char **argp)
...
+
+			free(argument);
+			free(new_buf);
+		}
+
+		free(command);
+		command = NULL;
+
+		return return_val;
 	} else {
```
gavtroy, line 3065: There's a "return" right above, so it would be cleaner to drop the "else"

command_mode.c
```diff
@@ -2616,13 +2653,98 @@ static void expand_colorscheme(const char *str)
...
+}
+
 static void expand_commands(const char *str);
 
 /* tab exp }}} */
 
 /* sort by name */
 struct command commands[] = {
 	{ "add",                   cmd_add,              1, 1,  expand_add,           0, 0          },
+	{ "alias",                 cmd_add_alias,        1, 1,  expand_alias,         0, 0          },
```
gavtroy, line 2747: I think yes @ CMD_UNSAFE question, since it would follow similar reasoning to "bind".

command_mode.c
```diff
@@ -2616,13 +2653,98 @@ static void expand_colorscheme(const char *str)
...
+		char **array = xnew(char*, 1);
+		array[0] = xstrdup("");
+
+		tabexp.head = xstrjoin(a->name, "=", a->command);
+		tabexp.tails = array;
+		tabexp.count = 1;
+		return;
+	}
+
+	// info_msg("test");
```
gavtroy, line 2717: remove debug line

gavtroy (changes requested):
> I knew about the bind command but that dose not accept parameters right?

It's binding the "push" command to pre-fill the command line.

gavtroy:
And yes, an addition to [Doc/cmus.txt](https://github.com/cmus/cmus/blob/master/Doc/cmus.txt) is also needed. By the way, I'm just reviewing but I can't say if your feature will ultimately be accepted or not.

C4N-6:
I am going to close this pull request, now that I look at this code it seams very bad. (it was one of my first c codes)

I am going to redo this command and create another pull request, that hopefully has a better code in it.

PR #1472, opened 2025-12-04
Fix unterminated-string-initialization warnings

chggr:
When compiling cmus I get multiple warnings around unterminated string initialization. This patch fixes the warnings by replacing the string initializers with explicit char initializers to avoid truncation of null terminator.


pgaskin (approved):

flyingmutant:
TBH I don't like the change, as the code becomes much less readable. I think either disabling the warning or using `nonstring` attribute is a better fix (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117178).

PR #1477, opened 2025-12-12
Fix memory leak in bass.c

chggr:
Method encode_ascii_string() allocates memory which is then duplicated in comments_add_const() but is never freed.


pgaskin (approved):
encode_ascii_string calls xmalloc, comments_add_const calls xstrdup.

cybersnow02:
Beside this memory leak, the BASS implementation is incomplete. See this issue: https://github.com/cmus/cmus/issues/1496

Issue #1207, opened 2022-07-14
Current position in args for `status_display_program`


ghost:
---

1) `uname -a`:
```
Linux lenovo 5.18.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 25 Jun 2022 20:22:01 +0000 x86_64 GNU/Linux
```

2) `cmus --version`:
```
cmus v2.9.1
```

Hello! I have written a simple Go program that translates current Cmus status to Discord Rich Presence (https://github.com/snvmk/cmus-presence). I'm not using `cmus-remote -Q` in it, but rather parsing arguments given to it. And the problem is, that these arguments don't have the `position` key, and so I cannot actual timestamps to presence, because they reset after a pause. Is it intended that it isn't passed? Also the example parameters:
```
status playing file /home/snvmk/Music/Portishead - Biscuit.mp3 artist Portishead albumartist Portishead album Dummy tracknumber 10 title Biscuit date 1994 duration 305
```

gavtroy:
Seems both reasonable and easy to add, if you'd be interested in making the PR?

It would also be nice to include the genre tag. I can't possibly imagine how that's less important than the musicbrainz_trackid.

snvmk:
Yeah, i guess I can make a PR, though I never made one. will learn with this xD

snvmk:
Tbh i already found a line where it does the thing :) Will try to do it...

snvmk:
Okay, track_info structure doesn't have the `position` key. Now I need to where is this struct defined...

snvmk:
This struct rather have a `time_t mtime` field... Interesting, what does this one contain?

gavtroy:
You'll have to get it from the player_info. You can see where it's used in `cmd_status` or `mpris_seeked`. The mtime is probably the file modification time from when the track info was cached.

pgaskin:
OP has disappeared.

PR #1502, opened 2026-04-24
Pass playback position to status_display_program

leana8959:
Related to #1207 and #1501.

I passed a pointer to the `player_info` struct around and added another argument to the argument array of the `status_display_program`. I'm not really familiar with C, let me know if there's a better way :)

Are there automated tests for this part of the code? I ran the built binary locally against a `status_display_program` that echos the params it gets to verify it works.

### DIFF
--- a/cmus-status-display
+++ b/cmus-status-display
@@ -10,7 +10,7 @@
 #
 # All keys contain only chars a-z. Values are UTF-8 strings.
 #
-# Keys: status file url artist album discnumber tracknumber title date
+# Keys: status file url artist album discnumber tracknumber title date duration position_sec
 #   - status (stopped, playing, paused) is always given
 #   - file or url is given only if track is 'loaded' in cmus
 #   - other keys/values are given only if they are available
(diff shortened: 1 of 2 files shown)


ui_curses.c
```diff
@@ -1835,10 +1835,17 @@ static int fill_status_program_track_info_args(char **argv, int i, struct track_
 		argv[i++] = xstrdup(stream_title);
 	}
 
+	if (pi->pos) {
```
gavtroy: It probably makes sense to report the position even if it's still 0.

cmus-status-display
```diff
@@ -10,7 +10,7 @@
 #
 # All keys contain only chars a-z. Values are UTF-8 strings.
 #
-# Keys: status file url artist album discnumber tracknumber title date
+# Keys: status file url artist album discnumber tracknumber title date duration position_sec
```
gavtroy: I'd remove the _sec for consistency with `duration`

ui_curses.c
leana8959: My bad, I probably meant to check `if (pi)`. Fixed.

gavtroy (approved):

Issue #1426, opened 2025-05-06
[Feature] Split delimited album artists while preserving album association


jinithin:
When an album has multiple artists in the `albumartist` tag (e.g., `BADBADNOTGOOD; KAYTRANADA`), cmus treats this as a single entity. Functionality to split these delimited artists while maintaining their album association would be greatly appreciated.

If this is already possible via configuration, please enlighten me. I've tried to read the documentation as best as I can.

# Use Case
I have albums tagged as "Artist1; Artist2" in the albumartist field and need:

- Both artists listed individually in artist views
- Both artists associated with the same album
- The album appearing in both artists' discographies
- Split should be available for both `;` and `/`

Issue #806, opened 2018-06-15
Multiple artists


Clyme:
I have a rather large music collection and I also have pretty much all releases of every artist in there. Some artists published hundreds of tracks, with the occasional collaboration this results in a LOT of duplicates (as seen in the screenshot). I wonder if it is possible, or if not if it could be added, that instead of having every track by artist A, B and C listed separately as `A/B/C - track`, the track is shown under A, B and C. This worked fine in Foobar2k, which i used on my laptop when I started my collection, but since i moved it to my Arch PC i haven't been able to find an application that supports this
![screenshot_2018-06-15_04-18-45](https://user-images.githubusercontent.com/40283399/41447125-2df061ce-7043-11e8-9f06-20b5ec2ce459.png)

jtl999:
I have the same issue. Something needs to be done, both mp3 and FLAC files.

Clyme:
Sure thing, is the attached file ok?
[tags.txt](https://github.com/cmus/cmus/files/2833286/tags.txt)

nefthy:
Ok now I understand. This is not possible in cmus. The artist in your id3tags is "A / B / C". For cmus this is one artist with that name, not 3 artists. Assigning the song to all artists, requires guesswork and interpreting tags in ways, that go beyond what the standard says. I do not think that such a feature would be accepted.

RodrigoLeiteF:
Since Picard 1.3, it's used a multi-valued tag named "artists" to deal with this problem. Would it be possible to implement this functionality in cmus? According to [the documentation](https://picard.musicbrainz.org/docs/tags/) they're separated by a semicolon (`;`).

PR #1505, opened 2026-05-30
Fix feat. tracks handling in tree view

feelamee:
Current problem - If track have few artists, then cmus will handle them as one. This is not only bad due to artists with names like "A, B", "A/B", "A featuring B", but also that track will not be shown in A or B album.

There are already few similar issues #1426 #806.

I suggest simple solution for this - assume that there is separator and that first artists is own track and others are featuring artists. So, we can create only artist A and correctly fill it's album.

This is a draft version to start discussion. I hope we will come to something more acceptable. Current addition change old behavior, so, probably, we should hide it under config option with default to old behavior.

Also, I saw @nefthy comment in #806
> Ok now I understand. This is not possible in cmus. The artist in your id3tags is "A / B / C". For cmus this is one artist with that name, not 3 artists. Assigning the song to all artists, requires guesswork ... . I do not think that such a feature would be accepted.

I understand that this is guesswork, etc, but cmus is an application for users, so it should satisfy users needs. Yes, such solution can add code smell, decrease it quality, but I think the most important for app is to *satisfy users need*. If it isn't, code quality doesn't matter. 

> and interpreting tags in ways, that go beyond what the standard says

What standard do you mentioning? If this is [id3v2](https://id3.org/id3v2.3.0), then it allow multiple artists:
> TPE1
>    The 'Lead artist(s)/Lead performer(s)/Soloist(s)/Performing group' is used for the main artist(s). They are seperated with the "/" character. 

But I'm not sure that this is useful here because cmus work not only with id3 tags (I think, I can be wrong).

### DIFF
--- a/lib.c
+++ b/lib.c
@@ -177,16 +177,19 @@ static bool track_exists(struct track_info *ti)
 	struct artist *artist;
 	struct album *album;
 	struct tree_track *track;
+	char * artist_name;
 
 	if (!ti->collkey_title)
 		return false;
 
-	char *artist_collkey_name = u_strcasecoll_key(tree_artist_name(ti));
+	artist_name = tree_artist_name(ti);
+	char *artist_collkey_name = u_strcasecoll_key(artist_name);
 	rb_for_each_entry(artist, node, &lib_artist_root, tree_node) {
 		if (strcmp(artist->collkey_name, artist_collkey_name) == 0)
 			break;
 	}
 	free(artist_collkey_name);
+	free(artist_name);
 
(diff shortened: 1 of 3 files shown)


lib.c
```diff
@@ -177,16 +177,19 @@ static bool track_exists(struct track_info *ti)
 	struct artist *artist;
 	struct album *album;
 	struct tree_track *track;
+	char * artist_name;
```
feelamee, line 180:
```suggestion
	char *artist_name;
```

PR #1513, opened 2026-08-27
Search in albumsort and artistsort in tree view

leana8959:
It is more intuitive to allow searching by both the displayed name and the sorting name.

I have some music where the artist name is in CJK characters, and it is sorted by the romanized name. Since it's already sorted that way, I think it would be nice to allow the search matching function to also match on the sorted name.

### DIFF
--- a/track_info.c
+++ b/track_info.c
@@ -171,8 +171,12 @@ int track_info_has_tag(const struct track_info *ti)
 
 static inline int match_word(const struct track_info *ti, const char *word, unsigned int flags)
 {
-	return ((flags & TI_MATCH_ARTIST) && ti->artist && u_strcasestr_base(ti->artist, word)) ||
-	       ((flags & TI_MATCH_ALBUM) && ti->album && u_strcasestr_base(ti->album, word)) ||
+	return ((flags & TI_MATCH_ARTIST) &&
+				   ((ti->artist && u_strcasestr_base(ti->artist, word))
+				|| ((ti->artistsort && u_strcasestr_base(ti->artistsort, word))))) ||
+	       ((flags & TI_MATCH_ALBUM) &&
+				   ((ti->album && u_strcasestr_base(ti->album, word))
+				|| ((ti->albumsort && u_strcasestr_base(ti->albumsort, word))))) ||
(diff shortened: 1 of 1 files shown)


pgaskin:
Sounds reasonable to me, thoughts @gavtroy?

gavtroy:
Sure, I can see it being useful for 5tyliz3d spellings too... although chances are you'd have to add the tag yourself.

Since the PR only mentions tree view I'll mention it does also work in the other views and with :live-filter.

leana8959:
> although chances are you'd have to add the tag yourself.

I use picard and apparently they have albumsort and artistsort for a lot of music I have.

> Since the PR only mentions tree view I'll mention it does also work in the other views and with :live-filter.

Is it because the search predicate is shared? I didn't know this. I can document it if needed by the way, but I'm not sure if it's considered a breaking change.

track_info.c
```diff
@@ -171,8 +171,12 @@ int track_info_has_tag(const struct track_info *ti)
 
 static inline int match_word(const struct track_info *ti, const char *word, unsigned int flags)
```
pgaskin: This isn't just for the tree view. It would probably best to add explicit flags for the sort fields to avoid invisible matches.

track_info.c
```diff
@@ -171,8 +171,12 @@ int track_info_has_tag(const struct track_info *ti)
...
 {
-	return ((flags & TI_MATCH_ARTIST) && ti->artist && u_strcasestr_base(ti->artist, word)) ||
-	       ((flags & TI_MATCH_ALBUM) && ti->album && u_strcasestr_base(ti->album, word)) ||
+	return ((flags & TI_MATCH_ARTIST) &&
+				   ((ti->artist && u_strcasestr_base(ti->artist, word))
+				|| ((ti->artistsort && u_strcasestr_base(ti->artistsort, word))))) ||
```
pgaskin: Note that `artistsort` actually contains `albumartistsort` if present.

track_info.c
pgaskin: This needs to be updated too.

pgaskin (changes requested):
The manpage also needs to be updated.

PR #1515, opened 2026-09-20
feat: add support for PCM float format in ALSA; add float pcm support in wavpack plugin

jsabev:
This PR adds support for PCM Float audio (SND_PCM_FORMAT_FLOAT) in alsa output plugin and in wavepack input plugin.

Issue: The alsa output plugin initializes the hardware with fixed sample sizes. This causes extreme distortion for PCM float files. 

Solution: Adjust the sample format (sample_format_t) to include a bit to identify float files and use this info to initialize the hw parameters in alsa accordingly.

Note: there were no free bits in the uint32 used to keep the sample info. I reduced the channel support from 255 to 127 to free the last bit for the float toggle.


op/alsa.c
```diff
@@ -152,9 +152,13 @@ static int alsa_set_hw_params(void)
 	if (rc < 0)
 		goto error;
 
-	alsa_fmt = snd_pcm_build_linear_format(sf_get_bits(alsa_sf), sf_get_bits(alsa_sf),
+	if(sf_get_float(alsa_sf)){
```
pgaskin: Formatting doesn't match.

sf.h
```diff
@@ -24,32 +24,37 @@
...
 #define sf_get_rate(sf)		(((sf) & SF_RATE_MASK     ) >> SF_RATE_SHIFT)
 #define sf_get_bits(sf)		(((sf) & SF_BITS_MASK     ) >> SF_BITS_SHIFT)
 #define sf_get_channels(sf)	(((sf) & SF_CHANNELS_MASK ) >> SF_CHANNELS_SHIFT)
+#define sf_get_float(sf)	(((sf) & SF_FLOAT_MASK    ) >> SF_FLOAT_SHIFT)
 
 #define sf_signed(val)		(((val) << SF_SIGNED_SHIFT   ) & SF_SIGNED_MASK)
 #define sf_rate(val)		(((val) << SF_RATE_SHIFT     ) & SF_RATE_MASK)
 #define sf_bits(val)		(((val) << SF_BITS_SHIFT     ) & SF_BITS_MASK)
 #define sf_channels(val)	(((val) << SF_CHANNELS_SHIFT ) & SF_CHANNELS_MASK)
+#define sf_float(val)		(((val) << SF_FLOAT_SHIFT    ) & SF_FLOAT_MASK)
```
pgaskin: Cast to unsigned before shifting.

pgaskin (changes requested):
This implementation is very incomplete:

- Doesn't work with replaygain, which currently only does integer scaling. Either handle it or ignore it.
- All output plugins need to either reject floats or handle them.
- Endianness isn't handled properly.
- Handle float stuff in `ip_setup` explicitly.

Alternatively, you could convert to S32 within the wavpack input plugin.

